PR #2, opened 2017-12-14
Uncommented _text_shadow_color

clobrano:
_text_shadow_color is used in %needs_attention and button.color, so
it should be available


galgalesh:
LGTM, thanks! :+1:

PR #3, opened 2017-12-17
Fixed stackswitcher button hover effect

clobrano:
currently, stackswitcher buttons do not have the same hover effect than other buttons on a headerbar
This commit adds a hover target for this widget with:
- dark headerbar bg and fg color
- gtk-icon-effect highlight

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1526,6 +1526,12 @@ headerbar {
         // specificity bump
         @include reload_button_styles($retain_borders: true);
       }
+
+      &:hover {
+        @include button(hover, $headerbar_bg_color);
+        color: $headerbar_fg_color;
+        -gtk-icon-effect: highlight;
+      }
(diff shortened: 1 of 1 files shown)


galgalesh:
Can you add a screenshot showing what you are talking about exactly? So I know what to look for while reviewing this.

FYI: (<kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>print-screen</kbd> and you can select an area to screenshot and paste it right into a comment box)

clobrano:
Of course, I should have done it immediatly.

This is the effect I can reproduce on master branch (the program is Tilix)
stackswitch_before

This is after the fix in this PR
stackswitch_after

galgalesh:
Ooooh, niiiice! How did you make those gifs?
 LGTM, merging this now. :balloon:

clobrano:
:D I used Peek, that unfortunately works on Xorg only, but I swear it's the same effect on Wayland :)

Issue #5, opened 2017-12-18
one-pixel white line


galgalesh:
I get a one-pixel white line at the top of the screen on my second monitor. Anyone able to reproduce this?

![image](https://user-images.githubusercontent.com/1492981/34121618-f88ed51a-e429-11e7-963b-ef5c8fc56db4.png)

I don't get it with Ambiance and Adwaita. Strange that it's a gtk bug, I thought it was a gnome-shell bug at first, but changing the gtk theme fixes it.

godlyranchdressing:
That's a problem if desktop icons are enabled. It's a Nautilus thing, so that's why changing GTK3 themes might get rid of it.

PR #6, opened 2017-12-18
fix white line issue

godlyranchdressing:
close issue #5


galgalesh:
Hm, this doesn't seem to fix issue #5. I'm still seeing this one-pixel line after I build your branch, even after rebooting.

galgalesh:
Yes, this does it. Thanks! :rabbit:

PR #8, opened 2017-12-29
Fixed again stackswitcher button hover effect

clobrano:
Until stackswitcher buttons are not included in "normal" ones, we should keep their style aligned


galgalesh:
Thanks! Maybe it's useful to include the reasoning behind it in a comment? So people don't remove it accidentally.

clobrano:
That's actually a good idea

Issue #19, opened 2018-01-05
borderless hover highlight on headerbar buttons


clobrano:
![borderless-hover-hb-buttons](https://user-images.githubusercontent.com/2883614/34611851-7f18d670-f227-11e7-8803-27e12083c688.gif)

see [here](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/119)

godlyranchdressing:
I was thinking to update the button mixin to take a $bc value. If if you use `$bc: none` then "fake" borders using the background color would be used. That'll get around any size issues since without borders the target for headerbar buttons would be 2px smaller.

PR #23, opened 2018-01-05
Headerbar buttons now have no border on hover/active

godlyranchdressing:
Closes issue #19 and also fixes the problem with GNOME Software's stackswitcher-like buttons


galgalesh:
LGTM :birthday:

PR #24, opened 2018-01-06
Standardize border-radius and use Suru blue for focused elements

godlyranchdressing:
Using blue for focused elements is closer to Unity 8's design and also fixes a couple visiblity issues for the focused elements (buttons especially, as they already use a grey color)


galgalesh:
Focused textboxes and lists still have the orange color. Is this intentional?

peek 2018-01-07 23-24

godlyranchdressing:
Yeah, at least for now. I'm hoping we can have the entry field look different when in focus and not rely on the orange.

PR #32, opened 2018-01-11
Upstream dir cleanup

didrocks:
* Remove unused index and gtk+3.0 files from gnome-theme-standard
* Rename now the gtk-2.0 directory and move it directly under the usptream directory to match destination layout, keeping the source package name.


galgalesh:
confirmed that this only touches upstream

PR #37, opened 2018-01-11
Add CI to build every PR

didrocks:
Ensure any PR is checked for building in the current ubuntu archive.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+sudo: required
+language: generic
+
+services:
+  - docker
+
+env:
+  - TRAVIS_DEBIAN_DISTRIBUTION=artful TRAVIS_DEBIAN_DERIVATIVE=ubuntu TRAVIS_DEBIAN_MIRROR=http://archive.ubuntu.com/ubuntu TRAVIS_DEBIAN_POCKETS="main restricted universe multiverse" TRAVIS_DEBIAN_INCREMENT_VERSION_NUMBER=true TRAVIS_DEBIAN_SECURITY_UPDATES=false
+
+script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,14 @@
...
+language: generic
+
+services:
+  - docker
+
+env:
+  - TRAVIS_DEBIAN_DISTRIBUTION=artful TRAVIS_DEBIAN_DERIVATIVE=ubuntu TRAVIS_DEBIAN_MIRROR=http://archive.ubuntu.com/ubuntu TRAVIS_DEBIAN_POCKETS="main restricted universe multiverse" TRAVIS_DEBIAN_INCREMENT_VERSION_NUMBER=true TRAVIS_DEBIAN_SECURITY_UPDATES=false
+
+script:
+  - wget -O- https://raw.githubusercontent.com/didrocks/travis.debian.net/gh-pages/script.sh | sh -
```
galgalesh: Ok, lots to process in 14 Loc :smile:. If I understand it correctly, this wil setup a docker container with an Ubuntu inside, build the binary package and run a linter against it? :thinking: 

So what's the difference between building it inside the Travis VM directly? Is it to get a different Ubuntu userspace?
And why does this download your local mirror?

didrocks: Indeed, in addition to adding all build-deps and such.

The advantage over building directly in the Travis VM is to be able to control easily which versions we build on, even building on multiple version (like, just adding another env line) and support matrix building then!
Also, the script is just a small modification over travis.debian.net, and I'll try to get those changes upstream. (less maintenance)

PR #71, opened 2018-01-16
Update GTK2 buttons, update stackswitcher + pathbar, various fixes

### LINKED ISSUE #67: Dimensions for context menu items are too large
For programs with very long menus such as LibreOffice, it becomes an inconvenience on large monitors.

godlyranchdressing:
- Update GTK2 buttons
- Fix disabled headerbar entry fields not using headerbar colors
- Update stackswitcher & path-bar to match mockups (gray underline unless selected)
- Reduce menu size (#67)
- Disabled headerbar buttons use a slightly transparent text color so they're less prominent
- Cleanup drawing file
- Tidy up checks and radios
- Use pan-end icons instead of suru icons for menu arrows

Closes #67


galgalesh:
I actually really liked the Suru arrows for menu items. I'd also like to see it used more in for example the arrows in the shell, like here:

image

Otherwise, LGTM :+1:

PR #95, opened 2018-01-23
Mockup/new breadcrumbs style

### LINKED ISSUE #94: Apply new breadcrumb style from mockup
image

design from [here](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/174)

clobrano:
closes #94 

actual result
new-pathbar01

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1612,6 +1612,11 @@ headerbar {
         }
         &:not(:backdrop):hover { color: $selected_fg_color; }
       }
+      &:not(:disabled):first-child,
+      &:not(:disabled):last-child {
+          color: $silk;
+          &:hover { color: $white; }
+      }
 
       &:first-child, &:last-child,
       &:backdrop:first-child, &:backdrop:last-child {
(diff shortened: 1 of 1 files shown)


galgalesh:
The enabled arrows next to the path bar still look disabled. In this screenshot, from left to right the arrows are

enabled - disabled - enabled - <path> - disabled

image

I think the pathbar arrows and font should follow the same color scheme as the arrow buttons on the left; that would make it more clear which ones are enabled and which ones aren't.

didrocks:
> I think the pathbar arrows and font should follow the same color scheme as the arrow buttons on the left; that would make it more clear which ones are enabled and which ones aren't.

+1

clobrano:
> I think the pathbar arrows and font should follow the same color scheme as the arrow buttons on the left; that would make it more clear which ones are enabled and which ones aren't.

Consider that the arrow on the leftmost has a different behavior: disable/enable -> dark/light, hover -> background gets gray.
leftmostarrow


The arrows at the borders of pathbar, instead, do not have any background highlight for consistency with the rest of the pathbar. This means we need three level of highlight: disable:dark, enabled:light, enabled+hover:lighter
pathbararrow



That said, I am not too happy with current solution, too, but I'm already using $silk (#CCC) for enabled color and white for enabled+hover and it looks too dark as well.

galgalesh:
Hmm, this is a tough one :confused: 

The flat headerbars are really difficult to get right, it seems..

I'll merge this so the others can give feedback on it.

PR #105, opened 2018-01-25
Fix entry caret color, add headerbar separator, fix small issues

### LINKED ISSUE #104: Text cursor not visible in Nautilus location bar
In Nautilus when I press CTRL + L to enter the path manually in the location bar the text cursor is almost invisible against the dark background. This is on Ubuntu 17.10.1

Version: 18.04 wip1+201801240731+097b554~ubuntu17.10.1

reported first [here](https://github.com/Ubuntu/gnome-shell-communitheme/issues/33)

godlyranchdressing:
- Spinner uses _backdrop_color($blue) in backdrop
- Some linked entries would not have a colored border in focus
- Use $graphite for headerbar separators
- Change $headerbar_backdrop_* to $backdrop_headerbar_* for consistency
- Use $text_color in menus
- Notebook tabs have a separator
- Shorten switch shadow to 1px
- Remove inapplicable states for titlebuttons (checked, disabled, etc.)
- Restore hover + active effects for close button

Closes #104

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1614,7 +1614,7 @@ headerbar {
         color: $headerbar_insensitive_color;
         &:backdrop {
           box-shadow: inset 0 -2px transparentize(black, 0.7);
-          color: $headerbar_backdrop_insensitive_color;
+          color: $backdrop_headerbar_insensitive_color;
         }
         &:not(:backdrop):hover { color: $selected_fg_color; }
       }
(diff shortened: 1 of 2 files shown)


clobrano:
I'm not sure about the separator between top bar and headerbar when window's maximized. It looks better without any separation to me. It looks good to me otherwise 👍

godlyranchdressing:
@clobrano It can stay or go for me. I just think it eases up the look of the dead space when you have something like GIMP open. I remember there were comments on that being an issue early on.  

The thing that ruins it for me is that the dock has a shadow instead of a separator at the top of it. Maybe we can experiment with a similar border instead of the shadow.

clobrano:
> The thing that ruins it for me is that the dock has a shadow instead of a separator at the top of it. Maybe we can experiment with a similar border instead of the shadow.

Yes, I agree with that. The shadow is even more disturbing when top panel is more transparent

PR #114, opened 2018-01-27
Fixed missing black underline on first/last child in stackswitcher

### LINKED ISSUE #106: Dark line under stackswitcher does not work on first/last child
image

This is a problem I introduced in my implementation of new the breadcrumb. I didn't considered that first/last child are not only the arrows in breadcrumb

clobrano:
closes #106


galgalesh:
Good! :bee:

PR #117, opened 2018-01-29
outline-radius property missing the “-gtk-” prefix in a couple of instances

fitojb:
Gtk+ nagged me about it on launching a graphical program from the terminal.


galgalesh:
Merged, thanks! :sun_with_face:

PR #118, opened 2018-01-29
Pathbar and stackswitcher updates

godlyranchdressing:
- Path-bar buttons all have a shared background color


clobrano:
Arrows in path-bar do not have any hover effect, is that on purpose?

godlyranchdressing:
Yes, they're aren't checkable buttons and the underline effect is only reserved for checkable buttons. If we want them to have some type of color feedback we could make the arrows themselves turn orange.

clobrano:
The only problem to me is the missing feedback when hover on the arrows, so maybe it's better with an orange color

godlyranchdressing:
This is a silly question, but how do you get the arrows to function? They're always disabled for me.

galgalesh:
open a long path and make the window of nautilus small so the path can't fit in the header completely.

clobrano:
LGTM 👍

PR #122, opened 2018-01-30
Add -security and -updates pockets

didrocks:
Fix the CI by adding some missing pockets as older version on the archive aren't apt gettable.


didrocks:
And CI is fixed! Merging

PR #131, opened 2018-01-31
Fix *@2.png assets sizes

nana-4:
Since 0.92, Inkscape has changed the default resolution from 90dpi to 96dpi:
http://wiki.inkscape.org/wiki/index.php/Release_notes/0.92#Important_changes

This PR changes the export dpi to `192` for `*@2.png` assets and fixes the sizes.


galgalesh:
Thanks a lot for these fixes! :crown:

PR #134, opened 2018-01-31
Add 20% transparent to disabled slider

### LINKED ISSUE #119: Deactivated switches and switches in off state looks the same
At a glance, deactivated switches and switches in off state looks almost the same. Adding a bit of gray to the white knob itself, would emphasize that this is deactivated. If possible I would use the same gray as the background (of the switch) at 50% opacity.

clobrano:
closes #119

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2797,7 +2797,7 @@ switch {
   &:hover slider { /* ?? */ }
 
   &:disabled slider {
-    @include button(insensitive, if($variant == "light", white, darken(white, 20%)), $flat:true);
+    @include button(insensitive, if($variant == "light", transparentize(white, 0.2), darken(white, 20%)), $flat:true);
   }
 
   &:backdrop {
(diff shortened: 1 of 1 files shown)


godlyranchdressing:
How about 

```
  &:disabled slider { background-color: mix($dark_fill, white, 10%); }

  &:backdrop {
    slider {
      transition: $backdrop_transition;

      @include button(backdrop, white, $flat:true);
    }

    &:disabled slider { background-color: mix($backdrop_dark_fill, white, 20%); }
  }
```
It's based on how $insensitive_bg_color is calculated and it'll give us:

Normal:
switch

Backdrop:
switch 2

I think I'll do the same to fix the issue with scale sliders being transparent in the background.

clobrano:
I like the simplification made removing the include button, but (unless I did something wrong) it doesn't seem to fix the initial problem (disabled too similar to enabled-off one)

image

I kept the simplification and added the transparency obtaining this one

image

What I don't get now is why the slider's color changes between checked and unchecked switches in backdrop windows.
image

It seems that this switcher's line causes this, but shouldn't it be referred to switcher only?

```css
    &:checked {
      box-shadow: none;
      background-color: _backdrop_color($c); <--- removing this, sliders keep same color in backdrop
    }
```

clobrano:
Closing this since design went to another direction

Issue #127, opened 2018-01-30
Titles in header bars look like buttons (or vice versa)


ya-d:
Titles in header bars should probably be bold (or anything else) to seperate them from text buttons (see e.g. gEdit; Settings > Details > Users). Titles are already bold if they are in breadcrumbs, tab interfaces or title bars.

<img width="937" alt="header bar - no bold titles" src="https://user-images.githubusercontent.com/30781876/35591141-46953c44-0609-11e8-8bf3-9f3226670d97.png">

Issue #85, opened 2018-01-20
Headerbar buttons are too big


godlyranchdressing:
With a smaller headerbar, buttons are a little too close to the edges of the headerbar. For uniformity they should be smaller overall, maybe just shave off 1/2px. Whatever dimensions we choose, image buttons need to be 1:1 in size.

PR #138, opened 2018-02-01
Bold window titles. Decrease button, entry, tab, check + radio sizes

godlyranchdressing:
- Fixes the issue of headerbar buttons being so large that they appear cramped
- Make window titles not blend in with flat headerbar text buttons
- Also fixes a whoopsie with checked widgets (using the underline effect) not using $selected_fg_color

Closes issue #85 
Closes issue #127

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -257,7 +257,7 @@ spinner {
 %entry,
 entry {
   %entry_basic, & {
-    min-height: 24px;
+    min-height: 22px;
     padding: 4px 8px;
     border: 1px solid;
     border-radius: $small_radius;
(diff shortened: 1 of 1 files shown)


Communitheme/gtk-3.0/_common.scss
```diff
@@ -511,8 +517,8 @@ $_dot_color: if($variant=='light', $selected_bg_color,
 %button,
 button {
   @at-root %button_basic, & {
-animation: shrink 1s linear infinite;
-    min-height: 24px;
```
nana-4: I think the button `min-height` and `min-width` should hold to `24px`. Since some buttons have `24px` icons, using less than `24px` sometimes results in irregular button heights.

Instead we can reduce `padding`.

clobrano:
@godlyranchdressing what do you think about @nana-4's review?

godlyranchdressing:
@nana-4 is more well-versed than me so I'll take their word for it. I'll just make a new PR only changing padding.

PR #139, opened 2018-02-01
Improve tab sizing

nana-4:
I thought current tabs are a bit too large, so I fine-tuned that as follows:

- make `notebook > header` height same as Adwaita.
- 4px margins for `tab`.
- 4px margins for `tab > button`.
- 1px margins for `notebook > header > button` as with Adwaita.
- removed/commented out unnecessary codes.

**Before:**
image

**After:**
image

This PR is only size changes, no (actual) design changes.

What do you think?


godlyranchdressing:
Looks good to me!

PR #146, opened 2018-02-07
Implemented button design rules on dark

### LINKED ISSUE #136: Follow button design rules on dark
- Only buttons when clicked or permanent clicked have inner-shadow
- Focus is always a surrounding orange line (should be present everywhere in the system)
- Connected buttons and breadcrumbs and similar have a grey back, selected areas/folders have a bottom orange line for feedback/orientation. The inner shadow could make (visual) problems I guess in some circumstances, so I suggest simple flat design. And they have a padding or background color border between to look separated.

aa60ec53b6454bc457914e8a20a8ff65e068cdbe
b2
b3
b4

clobrano:
closes #136

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1549,12 +1549,16 @@ headerbar {
     }
 
     button, .linked button {
-      @each $state, $t in ("", "normal"), (":hover", "hover"),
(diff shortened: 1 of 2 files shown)


clobrano:
I was sure to have put already some picture about this to help review. Here some

dark-style-headerbar-button

Issue #128, opened 2018-01-30
Off or inactive switches in header bars look out of place (too bright)


ya-d:
…and are barely distinguishable.

<img width="419" alt="headerbar_switch-off" src="https://user-images.githubusercontent.com/30781876/35592248-b275ac5c-060c-11e8-8810-445e5ce74736.png">
<img width="451" alt="headerbar_switch-inactive" src="https://user-images.githubusercontent.com/30781876/35592295-d3d8f908-060c-11e8-959d-3bd3c29a2438.png">

clobrano:
Switches are discussed also in #119

PR #147, opened 2018-02-08
Implemented transparent switch

### LINKED ISSUE #119: Deactivated switches and switches in off state looks the same

clobrano:
closes #119 #128


galgalesh:
Looking good :ice_cream:

PR #150, opened 2018-02-09
Removed hover effect from sidebar active row

### LINKED ISSUE #125: Active sidebar element shouldn't have hover effect
An active sidebar element (orange state) should have no hover effect as this darkens the orange to be weirdly brownish and has no necessary meaning. Although it can be clicked technically, the element is already active and clicking it has no further consequences. It has no hover on inactive windows.

See Files/Nautilus, Settings, Tweaks for example.

clobrano:
closes #125


galgalesh:
:+1:

PR #153, opened 2018-02-11
Styled message dialog to match dark look

### LINKED ISSUE #130: Style application dialogs to match interactive dialogs
It seems inconsistent with three different kinds of dialogs. Especially the white dialog with rounded corners _(that nothing else in the UI has)_ jumps in my eye. 

Could we change the white dialog design to match the dark look of system menu, calendar, brightness, ...?

dialogs

clobrano:
closes #130 

below the new look for message dialog

image

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -3910,48 +3910,72 @@ calendar {
  * Dialogs *
  ***********/
 messagedialog { // Message Dialog styling
+  $_bg: transparentize(darken($jet, 10%), 0.1);
   .titlebar {
-    min-height: 20px;
+    background-color: $_bg;
     background-image: none;
-    background-color: $bg_color;
-    border-style: none;
-    border-top-left-radius: $large_radius;
-    border-top-right-radius: $large_radius;
+    border-bottom: none;
+    border-left: 1px solid $inkstone;
+    border-right: 1px solid $inkstone;
+    border-top: 1px solid $inkstone;
+    border-top-left-radius: $medium_radius;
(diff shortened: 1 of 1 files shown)


galgalesh:
I'm not convinced this is a good idea because we need to differentiate between a dialog that "the shell" shows vs a dialog that "the application" shows.

I think we should follow upstream behavior on this, how do these dialogs look in Adwaita?

galgalesh:
Hm, I think we need to follow upstream on this. With wayland and snaps, this will even become a security issue: "is the shell asking for my password or is the application asking for my password"

galgalesh:
@didrocks, what do you think about this?

didrocks:
It's tricky. Indeed, my first natural argument would be like @galgalesh. However, if we only block this on "security feeling", nothing prevents any malicious application to style in the GNOME Shell style way their dialog asking for password…

The only thing (but that people won't know/notice) is that the entire screen turns darker with the bright hollow at the center. In wayland word, applications can't do that (or they are fullscreen and immitate the entire screen, which isn't possible for all configs ;)).

I think thus that the answer is based on: "do we want the Shell dialogs be different than apps dialog, in general?". My first answer was no, but after a couple of days thinking about it, I think that having them differentiated actually help: "is it something from the system or is it something specific to the app?". (without talking about security, just user's hint here).

Wdyt?

galgalesh:
> I think that having them differentiated actually help: "is it something from the system or is it something specific to the app?". (without talking about security, just user's hint here).

I agree; regardless of security, it will help users understand what's happening better. The shell being dark is also consistent with other UI elements such as the calendar, activities and the drawer.

Issue #158, opened 2018-02-12
Inconsistent window titlebars / headerbars


CDrummond:
The titlebars/headerbars are inconsistent between normal and dark variants. Dark variant has a light line under the buttons - is this an artifact from the general buttons in the dark variant (which also suffer from a ligher line under buttons)? Dark variant also has a 1 pix dark border around windows. This dark border is a nice addition (and should be in the lighter variant too) as it helps to distinguish the titlebars of overlapping inactive windows.
 
 Seeing as 'light' variant has dark titlebars, should they not be the same - and share the same code/css?

Dark titlebars. Here it is easier to distinguish the inactive titlebars:
![titlebars-dark](https://user-images.githubusercontent.com/1141694/36096386-16d089f2-0fee-11e8-9032-c5189e26b4dd.png)

Normal/light titlebars. Still possible to distinguish inactive titlebars, but not as clear. Still there is a difference between this and dark which should not be there.
![titlebars-light](https://user-images.githubusercontent.com/1141694/36096387-16e957ca-0fee-11e8-9577-aa5beeaa4ab9.png)

CDrummond:
This issue is **not** completely fixed. There is still a darker border around dark variant windows than those of the standard variant.  Whilst I like this border, it should either be on both variants or neither. ie. the titlebars of both variants should be the same. Can they not share the same CSS file?

clobrano:
@CDrummond could you please tell me which application still shows the issue?
Thanks a lot!

CDrummond:
The dark 1pix border from the screenshots is from gnome-terminal set to use the dark variant. But other apps that default to using the dark variant have the same issue - e.g. Photos, Videos, gtk3-widget-factory (when set to use 'Dark Theme')

godlyranchdressing:
Looks like the highlight effect (done through box-shadow) for dark buttons leaking through. Adding box-shadow: none for titlebuttons should fix the problem.

clobrano:
Maybe the update isn't available in ppa yet?
This is how it looks Videos in my system

![image](https://user-images.githubusercontent.com/2883614/36210570-0849ba4a-119f-11e8-9725-13e19efe690b.png)

![image](https://user-images.githubusercontent.com/2883614/36210592-1a06f39c-119f-11e8-953e-328fd3962f80.png)

clobrano:
I got it know, I was thinking only at headerbar buttons, sorry!

CDrummond:
[...]

PR #165, opened 2018-02-13
Explicitly unset button box-shadow in flat button

clobrano:
In button mixin, box-shadow is set explicitly only in non-flat buttons.
Unsetting it explicitly on flat button too, to close issue #158 where
box-shadow appears on flat button too.

Note: button in active state needs box-shadow even if flat to have
a "pressed" effect.


galgalesh:
Looks good, thanks!

PR #166, opened 2018-02-14
No sidebar hover when in backdrop state

godlyranchdressing:
Removes hover effects when a window is in backdrop

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -3776,11 +3776,11 @@ separator {
  *********/
 list {
   color: $text_color;
-  background-color: $base_color; // $sidebar_bg_color;
+  background-color: $base_color;
   border-color: $borders_color;
 
   &:backdrop {
-    background-color: $backdrop_base_color; // $backdrop_sidebar_bg_color;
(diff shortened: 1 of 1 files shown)


clobrano:
LGTM 👍

ya-d:
Thanks! Was there a discussion about *having* hover effects in backdrop windows? Firefox does this for example or MacOS. Seems more logical to me as buttons can still be pressed.

clobrano:
I think the shared idea is that since it is in backdrop state, it shouldn't take the user attention, that's why has been removed any hover effect when dropped.

ya-d:
Thanks for the reply. I would argue its not actively seeking attention but reacting to user interaction: move the pointer on a (still active) button. While backdropped windows look greyed out and inactive, they actually aren't: a single click has the same consequences as if the window was focused and can even close the window itself. This should be indicated by a hover effect IMHO.

No hover effect would make sense to me only if there was a *first* click required to bring the window to focus and activate it, and a *second* click to interact with its contents/controls. Just my two cents.

(PS for the sake of consistency: StackSwitcher and PathBar currently *do* have a hover effect in backdrop state which is also colored differently.)

PR #168, opened 2018-02-14
Reverts the stackswitcher

godlyranchdressing:
- The stackswitcher looks a little buggy right now so this reverts it to the old style until a decision is made on how to proceed/alter it (if even)
- Slider-buttons at the end of the path-bar use regular button radius size

screenshot from 2018-02-14 10-02-55

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1608,7 +1608,7 @@ headerbar {
       // underline_focus using the appropriate headerbar colors
       // this is used for the stackswitcher and pathbar as there are pathbars NOT in the headerbar
       @extend %underline_focus_effect;
-      background-color: lighten($headerbar_bg_color, 10%);
+      background-color: lighten($headerbar_bg_color, 5%);
 
       &:not(:checked) {
         color: $headerbar_insensitive_color;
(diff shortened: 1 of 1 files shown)


galgalesh:
Can you give a screenshot of what looks buggy?

godlyranchdressing:
Path-bar buttons aren't continuous and stackswitcher buttons aren't rounded, with a background color that doesn't change when hovered/pressed.
stacks
pathbar

clobrano:
That was the original design. We should ask @madsrh and Stefan (doesn't he have any github account?) if they agree

https://github.com/Ubuntu/gtk-communitheme/issues/136

galgalesh:
The issue with the old design was that it wasn't clear if the arrow was disabled or enabled. Maybe make the stackswitcher and pathbar continuous and round the stackswitcher?

godlyranchdressing:
It doesn't help that the arrows and the clickable folder names are all the same color. How about this? There's more of a difference between active, inactive and disabled text colors.

path

madsrh:
+1 from me. 
I liked the old design, although I would like remove the gray line underneath on hover. Just changing the label to white when hovering should be enough.

I think @godlyranchdressing suggestion looks good. As mentioned, the arrows are rarely used (unless you go seven folders down), so they should only be discreetly indicated when disabled IMHO.

clobrano:
LGTM 👍 
The path-bar is dead, long live the path-bar! 😄

PR #169, opened 2018-02-14
Decrease element padding

godlyranchdressing:
- Decreases element padding so buttons are smaller
Closes issue #85


clobrano:
Window controls seem a bit bigger than the other headerbar buttons
reduce-headerbar-issue

godlyranchdressing:
Updated.

clobrano:
LGTM 🎉

PR #170, opened 2018-02-14
Improve scrollbars

nana-4:
- Make the transition more smoother
- Use `rgba()` colors instead of `opacity` for more smoother transition
- Adjust scrollbar position/size

out


clobrano:
Awesome! Thanks 👍

PR #171, opened 2018-02-14
Make title fonts bolder (font-weight: 500)

godlyranchdressing:
- Uses 500 as the Ubuntu font is already a little wide
Closes issue #127


clobrano:
Looks good 👍 
Just a note: titles and button text still look similar, but at least now they differ a bit, so I think we can go with that

ya-d:
Like it!
Will this be the default for title bars as well? I played around with fonts so I can't currently reproduce.

Issue #157, opened 2018-02-12
Entry fields should use $fg_color


CDrummond:
There seem to be at least 2 different shades of grey used for active text. The labels in dialogs, buttons, are dark, and have good contrast. But the text in Nautilus' views, in menus, and for entries is lighter - with less contrast. Should all enabled text not be the same shade of grey? IMHO, the darker text in the dialog looks clearer.

Screenshot showing darker text in dialogs:
![text-shades-dialog](https://user-images.githubusercontent.com/1141694/36096290-a8517cd4-0fed-11e8-924c-d4fd36c42e4a.png)

Screenshot from gtk3-widget-factory showing the different shades of text between an entry field and a button:
![text-shades-entry](https://user-images.githubusercontent.com/1141694/36096291-a869c15e-0fed-11e8-8072-7e06547da628.png)

Screenshot of lighter text in Nautilus view:
![text-shades-view](https://user-images.githubusercontent.com/1141694/36096292-a8847ec2-0fed-11e8-8f93-ebd82e466464.png)

CDrummond:
I guess this can be closed now, due to #172

Issue #156, opened 2018-02-12
Text for inactive tabs looks disabled


CDrummond:
The same shade of grey is used for the text of inactive tabs as that of disabled widgets - making the tabs look disabled. The active tab is drawn bold, and has a lighter 'tab' - making it easy to see which tab is active. Therefore, changing the font shade is not required to indicate inactive tabs.

The following screenshot of Nautilus' file properties dialog demonstrates this:

![nautilus-file-properties](https://user-images.githubusercontent.com/1141694/36096194-2dfa800c-0fed-11e8-833d-ed804f4bfe60.png)

CDrummond:
Closed by #172

PR #111, opened 2018-01-26
Removed gap in Nautilus sidebar rows

clobrano:
Removed external top margin and added some more top and bottom padding
to avoid shrinking the rows too much.

closes #108

PR #172, opened 2018-02-14
Use $fg_color for entry fields, inactive tabs use $text_color

godlyranchdressing:
Closes issue #156 
Closes issue #157

### DIFF
--- a/Communitheme/gtk-3.0/_colors.scss
+++ b/Communitheme/gtk-3.0/_colors.scss
@@ -6,7 +6,7 @@
 $base_color: if($variant == 'light', #FFFFFF, #292933);
 $bg_color: if($variant == 'light', #FAFAFA, #1C1B21);
 $fg_color: if($variant == 'light', $inkstone, $porcelain);
-$text_color: if($variant == 'light', $slate, $silk);
+$text_color: if($variant == 'light', #2D2D2D, $silk);
 //
 $selected_fg_color: #ffffff;
 $selected_bg_color: $orange;
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2453,7 +2453,7 @@ notebook {
       min-width: 24px;
       padding: 4px 12px;
       outline-offset: -1px; // -5px;
-      color: $insensitive_fg_color;
+      color: $text_color;
       border: 1px solid transparent;
 
       &:hover {
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -37,7 +37,7 @@
 // @function entry_focus_shadow($fc:$selected_bg_color) { @return /*0 0 2px 2px rgba(0, 0, 0, 0.025),*/ 0 0 2px 2px transparentize($fc, .85); }
 @function entry_focus_shadow($fc:$selected_bg_color) { @return none; }
 
-@mixin entry($t, $fc:$selected_bg_color, $c: $base_color, $tc: $text_color, $edge: none) {
+@mixin entry($t, $fc:$selected_bg_color, $c: $base_color, $tc: $fg_color, $edge: none) {
 //
 // Entries drawing function
 //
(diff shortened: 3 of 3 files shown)


nana-4:
I feel we should use `$inkstone` for `$text_color` instead of using `$fg_color` for entry fields. (Although I personally prefer more darker color than `$inkstone` for text colors.)


The original #157 reporter also refers to the low contrast of **views** and **menus**, and I agree with that. But only changing entry fields will not solve other contrast lowness.

godlyranchdressing:
@nana-4 `$fg_color` uses `$inkstone` already. Maybe `$inkstone` for `$text_color` and `$jet` for `$fg_color`?

screenshot from 2018-02-15 11-27-23

I think `$jet` is a little too dark though. I'd like to stick to the color palette that's available but maybe we should try using another color. This is `$inkstone (#3D3D3D)` for `$text_color` and `#2D2D2D` for `$fg_color`:



We also have cool grey (#333) and text grey (#111) from the [Ubuntu palette](https://design.ubuntu.com/brand/colour-palette/) to work with.

nana-4:
@godlyranchdressing I mean, I suggested using `$inkstone` for both `$fg_color` and `$text_color`. :)

Yeah, I agree `$jet` is a little too dark. Personally, a slightly darker color than `$inkstone` (like `#333333`) is just right to me. But I also agree to stick to the color palette that's available.

IMHO it would be good to update the color palette and make `$inkstone` a little darker, but it may be necessary for more discussion among designers about which color (or brightness) is most suitable for text colors.

clobrano:
Sorry, what's the state of this PR? Can be merged?

godlyranchdressing:
@clobrano I guess. I'm just wondering if to darken $fg_color and $text_color or stick to the palette. I like #3D3D3D and #2D2D2D seen [here](https://user-images.githubusercontent.com/27529229/36264892-58a33a48-1244-11e8-9c83-0d712bcb0850.png) but I wonder if the difference between the two shades is too little, or if we even need a distinction between "text" and "fg". 

It did _come with Adwaita_ but is that enough reason to keep both around? I'm on board with just using $inkstone for both $fg_color and $text_color.

@Luxamann and @madsrh , what do you guys think?

clobrano:
They look different enough to me
image

> It did come with Adwaita but is that enough reason to keep both around? I'm on board with just using $inkstone for both $fg_color and $text_color.

It's ok to me to keep just one color, but to make it easier to manage future rebase on Adwaita project, I would keep the names in common.scss and assign them the same value on our color.scss

What do you think guys?

nana-4:
> It did come with Adwaita but is that enough reason to keep both around?

FYI, Adwaita didn't have `$text_color` until several years ago.
See https://bugzilla.gnome.org/show_bug.cgi?id=726316 for details.

madsrh:
After [discussing font-weight](https://github.com/Ubuntu/gtk-communitheme/issues/181) with @nana-4 yesterday, I did [a bunch of tests](https://www.dropbox.com/s/g97f1zs7leg5c3z/fontcolors.gif) and I agree that if there's a color we can use that sits between Jet and Inkstone it would be perfect. #2d2d2d seems to fit here.

clobrano:
Ok, so I'd wait @godlyranchdressing for the last fixes (if he wants to) and then we can merge this

godlyranchdressing:
Sorry, I'm a little behind. So we're agreeing to change $fg_color and $text_color. What values are we using for each?

clobrano:
> I'm just wondering if to darken $fg_color and $text_color or stick to the palette. I like #3D3D3D and #2D2D2D seen here but I wonder if the difference between the two shades is too little, or if we even need a distinction between "text" and "fg".

wasn't this one the only doubt remained?
If I got it right, @madsrh is OK with #2D2D2D

madsrh:
Yes, this is perfect, but I would like to suggest another color change. Should I merge this and start another issue? I'm guessing it's a quick fix.

Change disabled hover color for text in lists to **#ee886e** as the current (the one in the bottom) is very hard to read.

See the comparison here:
colors_

Issue #43, opened 2018-01-12
Folder highlight in nautilus is too much


godlyranchdressing:
Selected folders in Nautilus look red with the Suru icon theme.

clobrano:
Mads opened a bug on Nautilus to support a better folder selection https://gitlab.gnome.org/GNOME/nautilus/issues/202. The good is that there is an experimental feature (selectable in settings) that improves a lot this experience

![image](https://user-images.githubusercontent.com/2883614/34866641-840f9870-f77d-11e7-92a4-0bc2776a7961.png)

didrocks:
If it's only a gsettings key to change, we can do this in the ubuntu session without any problem.

clobrano:
Yes! It's

gsettings set org.gnome.nautilus.preferences use-experimental-views true

didrocks:
Ok, I've looked at the code that is enabled via this option. I'm not really confortable enabling this on the LTS on the default session (if we switch this theme by default) as the code is indeed still new, not really tested and quite different from what we tested on for long.
No worry to do that in the separate session though. Let me add a PR for this.

madsrh:
The current design has a white overlay on hover that doesn't work well with textfiles and other bright icons. Perhaps the orange filename outline on hover could work?

The selected icon is still colored outside the icon. @godlyranchdressing would this be possible?
![image](https://user-images.githubusercontent.com/3986894/35128733-6790f672-fcb8-11e7-988d-9cfda95bd223.png)

madsrh:
Like this - only with changes to the hover state:

![hover](https://user-images.githubusercontent.com/3986894/35306159-b653054c-009c-11e8-88f3-c7c4b4e7726c.gif)

As the issue title says, the colored icon is too much. Would it be be easy enough to spot the selected file without it? (it seems that icons on the 17.10 desktop works this way)

didrocks:
+1 on not "oranging" the icon. I think as well that on some icons, the selected color overlay can render especially badly.

madsrh:
@clobrano and @godlyranchdressing I see that Nautilus has changed as [suggested by Carlo](https://github.com/Ubuntu/gtk-communitheme/issues/43#issuecomment-357177917). Do you think we could give [the above a try](https://github.com/Ubuntu/gtk-communitheme/issues/43#issuecomment-359972944)? In the gif it's none, hover _(1 or 2px not sure)_, selected

clobrano:
fixed by #173

[...]

PR #173, opened 2018-02-16
Reduces nautilus' folder highlight effect in iconview

godlyranchdressing:
- Workaround for issue #43
- It was under the wrong class name


clobrano:
Very good 👍

PR #175, opened 2018-02-17
Let dark variant use the same backdrop border color as the light one

### LINKED ISSUE #158: Inconsistent window titlebars / headerbars

clobrano:
closes #158

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4492,7 +4492,7 @@ decoration {
   // this needs to be transparent
   // see bug #722563
   $_wm_border: if($variant=='light', transparentize(black, 0.77), transparentize($borders_color, 0.1));
-  $_wm_border_backdrop: if($variant=='light', transparentize(black, 0.82), transparentize($borders_color, 0.1));
+  $_wm_border_backdrop: transparentize(black, 0.82);
 
   box-shadow: 0 3px 9px 1px transparentize(black, 0.5),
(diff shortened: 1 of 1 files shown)


CDrummond:
Perhaps I have not merged correctly, but I see no change. However if I also change the preceding line to
```
$_wm_border: transparentize(black, 0.77);
```

Then it looks as it should.

CDrummond:
Now the two variants match. Of course, this means all variants have non-sharp window borders...

clobrano:
We put some effort in making backdrop windows visible changing backdrop colors only (so that it's clearly visible in a 13'' screen like the one I'm using now) and to me the difference is still enough. 

image

CDrummond:
The windows look great! And so does the theme overall. I'm just nitpicking with the _slightly_ blurry windows.

PR #177, opened 2018-02-18
Remove treeview header button's outline-radius

godlyranchdressing:
Treeview header buttons are square, but they inherit the squircle outline radius from regular buttons, this commit corrects that


clobrano:
LGTM 👍

Issue #160, opened 2018-02-12
Bread crumbs and StackSwitcher: rounded corners, vertical separators and consistency


ya-d:
1. Both elements should use rounded corners as do regular buttons. (Arrows in bread crumbs use an even larger radius, while StackSwitcher is not rounded at all.)
2. Bread crumbs and some StackSwitcher elements like in Widget Factory or Rythmbox use vertically separating lines between connected buttons, while others do not (Software). Looked much cleaner without separators to me. In any case: stay consistent! (They even are of different width in bread crumbs and StackSwitcher.)
3. Active selection uses bold font in bread crumbs but regular in StackSwitcher. I prefer bold. (See #127 #138 )

Also see #155 .

madsrh:
1. The discussion for the breadcrumbs / stackswitcher buttons is [found here](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/244).

2. Can we try changing the separator color to #383535 like the slider-button background? I agree that the current design looks less clean.

3. I can't find any reference to text color in Stefan's post, but we should change the selected button text to white (currently #cecece) and IMHO **NOT** bold. I would argue that an orange line and white text makes it pretty easy to spot where you are. Also the bold font doesn't render that well.

nana-4:
I wonder if breadcrumbs / stackswitcher really should have a brighter background.

Because **stackswitcher** has underline marks but does **not** have arrows, [it's hard to round the corners technically](https://user-images.githubusercontent.com/9556384/36337218-e6327022-13d4-11e8-8b2e-65d0175cca86.png). (Even with the original mockup, stackswitcher corners are not rounded, though.)

Or we can round [only the upper corners](https://user-images.githubusercontent.com/9556384/36336944-7e9eb3b6-13d0-11e8-9dd5-ab6b0311d99a.png), but it may seem odd.

Or I think it's also okay to remove the background only from stackswitcher, as #168 has already done.

nana-4:
Sorry, but I take back my previous comment because I came up with a way to round stackswitcher well.
I will send the PR.

ya-d:
Cool! 🎉 Thanks for the effort!

I'm probably the only one who would like active stack to be bold?! I suggest weight 500 for this as well as for path bar (seems a bit heavier at the moment), compare #171.

nana-4:
[...]

PR #180, opened 2018-02-19
Improve stackswitcher and pathbar

### LINKED ISSUE #174: Thin line between GtkRevealer in GtkFileChoserDialog is hard to see
I guess this css part comes from the NautilusPathBar theming.
This "same" line in a white version is now in the GtkFileChoserDialog, too.

bildschirmfoto-20180215205828-1283x1005
bildschirmfoto-20180217021654-1494x973

Imho it is really hard to see - maybe it would be better to remove the thin line in nauiluspathbar & gtkrevealer again

nana-4:
- Make the stackswitcher same style as pathbar (#160)
- Round the stackswitcher corners by using additional `border` (#160)
image
- Fix style of pathbar not in the headerbar (#174)
image
- Improve the buttons behavior slightly
image
- Simplify the code

Closes #174

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1617,10 +1617,6 @@ headerbar {
       // this is used for the stackswitcher and pathbar as there are pathbars NOT in the headerbar
       @extend %underline_focus_effect;
 
-      // gets rid of the radius
-      border-radius: 0;
-      -gtk-outline-radius: 0;
-
       // lighten($c, 5%) spits out the same bg-color as include button(normal, $headerbar_bg_color)
       // not using include button, since we don't need all the other styling
       $c: lighten($c, 5%);
@@ -1637,7 +1633,7 @@ headerbar {
         color: $selected_fg_color;
         box-shadow: inset 0 -2px $selected_bg_color;
 
-        &:hover { box-shadow: inset 0 -2px mix($selected_fg_color, $selected_bg_color, 10%) }
+        &:hover { box-shadow: inset 0 -2px mix($selected_fg_color, $selected_bg_color, 10%); }
       }
 
       &:backdrop {
@@ -1654,7 +1650,7 @@ headerbar {
     .horizontal.linked.stack-switcher,
     buttonbox.horizontal.linked {
       // round the stackswitcher corners by using border
-      border-width: 0 6px;
+      border-width: 0 8px;
       border-style: solid;
       border-color: $c;
       border-radius: $small_radius;
@@ -1666,6 +1662,11 @@ headerbar {
     buttonbox.horizontal.linked button {
       // style the stackswitcher
       @extend %underline_focus_headerbar;
+
+      // gets rid of the radius
+      border-radius: 0;
+      -gtk-outline-radius: 0;
+
       margin: 0;
       min-width: 100px;
       border: none;
@@ -1906,7 +1907,7 @@ headerbar { // headerbar border rounding
     color: $fg_color;
     box-shadow: inset 0 -2px $selected_bg_color;
 
-    &:hover { box-shadow: inset 0 -2px mix($selected_fg_color, $selected_bg_color, 10%) }
+    &:hover { box-shadow: inset 0 -2px mix($selected_fg_color, $selected_bg_color, 10%); }
   }
 
   &:backdrop {
@@ -1942,9 +1943,6 @@ headerbar { // headerbar border rounding
     padding-left: 0;
     padding-right: 0;
     box-shadow: none;
-
-    &:first-child { border-radius: $small_radius 0 0 $small_radius; }
-    &:last-child { border-radius: 0 $small_radius $small_radius 0; }
 
     &:hover { color: $selected_bg_color; }
     &:disabled image {
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1616,67 +1616,75 @@ headerbar {
       // underline_focus using the appropriate headerbar colors
       // this is used for the stackswitcher and pathbar as there are pathbars NOT in the headerbar
       @extend %underline_focus_effect;
-      background-color: lighten($headerbar_bg_color, 5%);
 
-      &:not(:checked) {
-        &, label { color: $headerbar_text_color; }
-        &:hover { &, label { color: $headerbar_fg_color; } }
-        &:backdrop { &, label { color: $backdrop_headerbar_insensitive_color; } }
-      }
-
-      &:hover { box-shadow: inset 0 -2px $headerbar_text_color; }
-      &:checked:not(:indeterminate) {
-        &, label { color: $selected_fg_color; }
-        &:backdrop { &, label { color: $backdrop_headerbar_fg_color; }}}
-    }
-
-    %basic_underline_styling {
-      // makes the widget flat
-      // uses the underline effect
       // gets rid of the radius
-      @extend %underline_focus_headerbar;
-      @extend %undecorated_button;
       border-radius: 0;
       -gtk-outline-radius: 0;
+
+      // lighten($c, 5%) spits out the same bg-color as include button(normal, $headerbar_bg_color)
+      // not using include button, since we don't need all the other styling
+      $c: lighten($c, 5%);
+      background-color: $c;
+      color: $headerbar_text_color;
+
+      // &:hover { color: $headerbar_fg_color; }
+
+      &:hover { box-shadow: inset 0 -2px transparentize($headerbar_text_color, 0.5); }
+
+      &:active { box-shadow: inset 0 -2px $headerbar_text_color; }
+
+      &:checked {
+        color: $selected_fg_color;
+        box-shadow: inset 0 -2px $selected_bg_color;
+
+        &:hover { box-shadow: inset 0 -2px mix($selected_fg_color, $selected_bg_color, 10%) }
+      }
+
+      &:backdrop {
+        background-color: mix($c, $headerbar_bg_color, 50%);
+        color: $backdrop_headerbar_insensitive_color;
+
+        &:checked {
+          color: $backdrop_headerbar_fg_color;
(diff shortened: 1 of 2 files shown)


nana-4:
I made the PR because I came up with a way to round the stackswitcher.
But if you don't like the design, please close this.

clobrano:
Hey @nana-4 this looks very good, thanks a lot!

Unfortunately, this can't be applied to current master, because just yesterday I accepted @godlyranchdressing  PR #168 (I tested your PR removing this last merge first).

But, your PR is very close to #168 and moreover fixes another problem @madsrh saw (hover underlined too bright in path-bar), so if for them is ok, I could revert last merge and let this PR be merged instead

clobrano:
I just saw a little problem with rounded borders in gnome software

image

madsrh:
Great work!
Although the rounded stackswitcher looks really good and fits with the rest of the design, I wonder if we should consider making the bottom square and only the top rounded, because of [this comment](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/265):


> In Ubuntu it’s much more common to see a button’s focus ring (as shown in these screenshots, for example). Therefore, the kookiness of a focus ring having three square corners, but one circular one, would be much more of a problem.

I'm in doubt here because we already use this in the shell dialogs. I can't tag @Luxamann but I'll point him to this issue.

nana-4:
Hi, @clobrano 

> Unfortunately, this can't be applied to current master, because just yesterday I accepted @godlyranchdressing PR #168 (I tested your PR removing this last merge first).

> if for them is ok, I could revert last merge and let this PR be merged instead

I made this branch after #168 was merged. :stuck_out_tongue_winking_eye:
See: https://github.com/Ubuntu/gtk-communitheme/network

And this github page says,
> This branch has no conflicts with the base branch

So I think there is no technical problem to merge this. ;)

> Selected border is still squared.

I'll investigate whether the focus rings can be rounded.

clobrano:
> I made this branch after #168 was merged

uhm so maybe I did something wrong because nothing changed when I applied your PR to current master

nana-4:
HI, @madsrh 

> I wonder if we should consider making the bottom square and only the top rounded

Excuse me, but what are you talking about? In this PR, all four corners of stackswitcher are rounded.

nana-4:
I found a way to round the side of the focus ring, but I'm not sure which one is more slick. 🤔

current:
image

rounded side corners:
image

clobrano:
In my opinion, current looks incomplete. New rounded side corner improves it a bit, even if that step between focus ring and underline doesn't feel right.

I'd say to wait @madsrh and luxamman to clarify previous observation (I must admit I didn't get the point of above quotation myself :smile:), even because with squared bottom corner we going to avoid that step

ya-d:
Could those rounded "endings" be larger (like the arrow pieces in Nautilus)? So that it would be clear that the buttons aren't expected to reach the border and thus won't look "incomplete". Focus highlight could than still be square.

larger endings

nana-4:
@ya-d Thanks for sharing your idea. :)

> Could those rounded "endings" be larger (like the arrow pieces in Nautilus)?

It's possible, but large "endings" may look odd.

`12px` "endings":
image

`20px` "endings":
image

I think `12px` is not bad, but instead it wastes spaces. (Note that current "endings" are `6px`.)

> So that it would be clear that the buttons aren't expected to reach the border

FYI, in the new "rounded side corners" (see my above comment), the "endings" (border) will be also clickable. (But before pushing that, I'm also waiting @madsrh and luxamman to clarify previous observation. :smile:)

clobrano:
@ya-d in pathbar is fine that the selection is rounded because there are buttons at the sides, in this case it won't be good, I think.

> FYI, in the new "rounded side corners" (see my above comment), the "endings" (border) will be also clickable.
Those are part of the first/last buttons right?

Do you have time to try bottom corner squared? To see how the selection looks like

nana-4:
> Those are part of the first/last buttons right?

Yes, that's right.

> Do you have time to try bottom corner squared? To see how the selection looks like

Is it like this? (The rightmost button is focused.)

image

madsrh:
@nana-4 This is definitely the best solution so far IMO.

nana-4:
IMO the _bottom squared_ `stackswitcher` lacks consistency a bit when comparing to the _all rounded_ `pathbar`.

I understand both `stackswitcher` and `pathbar` are **containers** that contain page switcher buttons. So I think both should have same (rounded corner) style.

After spending some time and testing, now I'm thinking `8px` "endings" (with all squared focus rings, based on @ya-d's idea) is the best solution for the `stackswitcher`. It's most consistent and looks good IMO:
image

About the focus ring;

- Unless you're an active keyboard user, it's rare that the `stackswitcher` / `pathbar` buttons are focused (except for GNOME Software. It actually uses `buttonbox` instead of  `stackswitcher`). (_So the focus ring style is not a big issue here IMHO._)

- Since the arrow buttons on the `pathbar` are frequently inactive, in many cases all the button focus rings are square:

So even if the first/last button focus rings on the `stackswitcher` are square, it still has consistency IMO:


So, I vote for `8px` rounded "endings" so far.  But I'm an outsider, so I'll obey the designers' decision eventually. :)

nana-4:
> I can't tag @luxamann but

I guess @Luxamman is he. I would like to know what you think. :)

madsrh:
+1 @nana-4 As you say this 
image
looks good, but my concern was for the focusring. 

image

Either of these two solutions will keep me up at night, but at the same time I don't feel like we've found the holy grail. Let's see what @Luxamann says...

Luxamman:
Hey,
thanks for thinking things thru, I jump in a little late^^

This is again a little tricky because the pathbar is definitely something different than the stackswitcher in meaning IMO. So the pathbar is fine for me like it is now, also because that whole element is hard to style right because of how it is built and work and it needs the connected background more than the stackswitcher do. 

But - we already use a style we can adapt to that problem, I guess. The simple orange line without background - like in the top-bar, like in LibreOffice - and also like the stackswitcher when choosing wallpapers. 

lines

I already tried some designs, because the consent would be to let the orange line touch the bottom element, but maybe the headerbar is too high to look sleek. 

bottouch

I'm not sure if it is possible (or look too odd) to touch the bottom element, otherwise, I would suggest just a simple orange line without background (like it is now in "choosing wallpaper"). The focus is a little compromise, but as long there is no rounded corner background, they can be like they are. 

This would keep things simple, looks consistent with other elements and is also referring the Suru theming. 

Hope MHO is helping here :)
What do you all think about it?

ya-d:
First things first: I'm an amateur outsider so I don't expect anybody to listen 😄 – also I'm not really comfortable talking about design topics in English. But: I'll give you my two cents non the less. 😝

Without having seen it in action really I have two gripes with this design: (1.) Proximity/Context; (2.) Similarity/Concepts.

(1.) (a) Within headerbar there is much more vertical space compared to top bar or menu bar which is why I'm going to argue that StackSwitcher *is* much closer too PathBar than to those two. As can be seen in your designs, the orange underline is either unconnected to the window content (which is why you want to move it towards the bottom) or to the text label (when you do). (b) Also there is much more horizontal space which leaves the three stacks/pages unconnected and hence unrelated to each other. (And makes them indistinguishable from regular buttons, leading to…)

(2.) (a) While menu or top bar entries open out an element that is connected to the entry itself and lives *in front* of the current screen (in general they are drop down menus with the special case of Activities which spreads the whole screen but still is (kind of) an overlay), StackSwitcher lets you swipe through different pages that change the whole screen of the application. (b) If you "hop" through folder hierarchy in PathBar, what's the difference to switching stacks? Conceptually those are much closer to me than to menus. (c) As the main point of interaction with the app stacks should within a glance be easily recognizable and distinguishable from regular text buttons.

nana-4:
Thanks @Luxamman and @ya-d :smile: 

IMHO, semantically, elements in the top-bar and menubar (LibreOffice) are _menu-shown-button_ but elements in the stackswicher and pathbar are _page-switcher-button_ (as @ya-d stated in detail). So, regardless of whether we remove the background or not, I think both stackswicher and pathbar should have a consistent style.

Personally I also like the without background, but IMHO the touched bottom line is not very good UX because it'll have a different button height than other headerbar buttons.

---

For your reference; Maybe we also should consider the [GtkPathBar improvements](https://wiki.gnome.org/Projects/GTK+/Roadmap) that may be done in the future. The mockups can be found here:
- https://www.youtube.com/watch?v=CvedN8mE02k
- https://github.com/gnome-design-team/gnome-mockups/blob/master/theming/widgets/breadcrumbs/breadcrumbs-wires.png

madsrh:
@nana-4 Is your PR for what you posted in comment #1? I'll say, let give it a try and evaluate later if necessary. As you suggest, we might want to try 8-10px rather than just 6px.

Like @ya-d I also see stackswitcher and pathbar to be releated. Without the background it look too loose, especially when the buttons ([like Cancel](https://user-images.githubusercontent.com/20417555/36585434-2f502d8c-187e-11e8-8a22-41fbc955b297.png)) _currently_ don't have any borders - it kind of messy.
Let's get as close as possible to the pathbar design (except the bold for selected). I'm going the merge this PR. OK @Luxamman ? @clobrano is the conflict you mentioned earlier fixed?

I like the Gnome mockups as the current pathbar looks very outdated. We tried [something like that](https://community.ubuntu.com/t/design-mockups-updates-designer-only/2053/31) in earlier mockups.

clobrano:
I agree with @madsrh. I'd like the solution with only upper corners rounded for the focus ring, but I could accept also the last one with squared focus ring.

nana-4:
Now I've changed the stackswitcher first/last spaces to 8px (and added some minor fixes).

clobrano:
LGTM :+1:

Luxamman:
Sure, let's try!
And it's funny, for me the stackswitcher are very different in meaning than the path bar - sure, they have a similar way of working, but the path bar is for me much more like the address bar in a browser or like the file browser in windows. One is a path, the other one is an area/section - but that is MHO ;) 
I always like to get different point of views^^

Issue #178, opened 2018-02-18
Border color when hovering on flat buttons are too dark


godlyranchdressing:
I'm making an issue since I don't know why it was changed but I'm guessing it was to deal with headerbar buttons?

`border-color: if($flat==true, $_pressed, _border_color($_bg));`

![flat2](https://user-images.githubusercontent.com/27529229/36346873-fda87fa4-141e-11e8-85d7-9408c4869eb0.jpg)

`border-color: if($flat==true, $_bg, _border_color($_bg));` 

![flat](https://user-images.githubusercontent.com/27529229/36346872-fa0f679a-141e-11e8-9430-3ab499615722.jpg)

The border is a little too dark for flat buttons not in the headerbar. If we want a darker border color for headerbar buttons then maybe we should do:

`border-color: if($flat==true, if($c == $headerbar_bg_color, $_pressed, $_bg), _border_color($_bg));`

So flat buttons not in the headerbar won't be affected.

ya-d:
I hijack this issue to add a similar issue for the close button in tabs (see Widget Factory page 3, tabs in gText), which has too strong of a border for both foreground and background tabs. (Also (almost?) none other button changes icon/text color on hover but that might be necessary due to light colors and low contrast. Just wanted to add it for consideration.)

![tab close - dark border](https://user-images.githubusercontent.com/30781876/36350388-62ff7678-1497-11e8-8d06-c0122f52cef9.png)

clobrano:
closed by #182

PR #182, opened 2018-02-20
Fix flat buttons having a too dark border

godlyranchdressing:
The darker border will only be used for buttons that use $headerbar_bg_color
Closes issue #178


clobrano:
LGTM :+1:

PR #187, opened 2018-02-22
Update GTK2 theme

godlyranchdressing:
- Check and radio buttons
- Gray menu hover color
- Treeview header button
- GTK Scale slider
- Trough/progressbar


clobrano:
hi @godlyranchdressing could you suggest me some gtk2 apps that show the changes?

godlyranchdressing:
No problem. There's a GTK2 widget factory I use that's great for testing -> https://github.com/valr/awf
The only thing missing from it is checkboxes in menus, which you can see with GIMP/Inkscape.

awf

Some stuff still needs to be finetuned.

clobrano:
Wow, great!
I'm sorry, but have you seen #188? You might need to update checkboxes and radiobuttons if possible

godlyranchdressing:
Yeah, I realized. :satisfied: 
It'll just be some color changes since they're using the right design instead of Adwaita's.

clobrano:
yeah! :+1:

PR #188, opened 2018-02-22
Applied Canonical design to checkbox and radiobutton

### LINKED ISSUE #162: Update check boxes/radio buttons to match design intended by Canonical
Disabled check boxes/radio buttons should probably stay grey, while with the latest update they became green. (No choice means no "positive" selection.) Maybe use the lighter green for partially selected instead. (I personally prefer ditching yellow for mixed sate.)

check boxes

clobrano:
closes #162

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2912,9 +2912,26 @@ radio {
 %check, %radio, check, radio {
   & {
     // for unchecked
-    @each $state, $t in ("", "normal"), (":hover", "hover"), (":active", "active"),
-      (":disabled", "insensitive"), (":backdrop", "backdrop"), (":backdrop:disabled", 'backdrop-insensitive') {
-      &#{$state} { @include button($t, $dark_fill, $flat:true); }
+    @each $state, $t in ("", "normal"),
+                        (":hover", "hover"),
+                        (":active", "active") {
+                          &#{$state} {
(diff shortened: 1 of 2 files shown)


godlyranchdressing:
Checks and radios in treeviews behave strangely (a radio button even loses its radius in GTK3 Widget Factory and I have no idea why).

I played around and the `background-color: $dark_fill` line around line 3003 and the `box-shadow: 0 0 0 1px white` line around line 3060 need to be removed. That fixes the problem with the treeview. 

There was an issue (that I can't find right now) where checkboxes would get clipped in Software Updater prompts. I say we just merge it and deal with those issues when/if they pop up.

clobrano:
Wow, nice you spotted it, I'll fix that, thank you!

clobrano:
Sorry, It's my fault. I gave Mads green light for this just a second before. I'm going to open a new ticket and fix this in another PR

clobrano:
> I played around and the background-color: $dark_fill line around line 3003 and the box-shadow: 0 0 0 1px white line around line 3060 need to be removed. That fixes the problem with the treeview.

@godlyranchdressing I have a `box-shadow: **inset** 0 0 0 1px white;` at 3066, that needs to be removed, but it's not enough. To avoid radiobutton to get squared on select I need to split the `treeview.view radio, treeview.view check {` around line 3058 and add a `border-radius: 100%` to the radio one part only :unamused:

PR #208, opened 2018-02-26
Use green for selection-mode

godlyranchdressing:
Also fixes some buttons appearing as a regular headerbar button. See below:

screenshot from 2018-02-26 15-43-56
screenshot from 2018-02-26 15-43-43


clobrano:
LGTM :+1:

Issue #199, opened 2018-02-24
Scrollbar trough extends past edge of Terminal window


Xyaneon:
I'm using the Communitheme session on Wayland in Ubuntu Bionic Beaver. One thing I noticed is that when I open a Terminal window, the trough for the scrollbar extends past the edge of the window, instead of being flush. Terminal is the only app I have noticed this in.

I think the same thing happens with Ambiance, but to a much lesser degree.

See the screenshot below:

![communitheme terminal scrollbar past window edge](https://user-images.githubusercontent.com/3268365/36632915-a7899400-195a-11e8-96cb-940c26a05427.png)

clobrano:
Hi @Xyaneon thanks for taking time to report this!
Yes, someone have been affected by this previously, and then disappeared. Apparently it's back :disappointed:

clobrano:
Still unable to reproduce it

nana-4:
It seems not to be the theme issue, but a Ubuntu on Wayland specific problem:
https://bugs.launchpad.net/ubuntu/+source/gnome-terminal/+bug/1720651
https://bugs.launchpad.net/ubuntu/+source/gnome-terminal/+bug/1750889

Xyaneon:
@nana-4 I changed my session to Communitheme on Xorg, and indeed the issue is not present there. I will mark myself as affected on those bug reports. Thanks!

clobrano:
Oh that's why! I always work on Xorg so that I can use Peek :smile:

nana-4:
However, this can be fixed by adding a workaround with the theme. :-)
I'll send the PR.

ET-Bent:
Any updates about merging 47069ba7c3ef6afce0c16b7e6823056558c299ba ?

clobrano:
> Any update about merging 47069ba ?

It has been committed already, May 31st

https://github.com/ubuntu/yaru/blob/836fd8b1975aa6aefb44c7fd5fa31076a7ac9e57/gnome-shell/src/gnome-shell-sass/_common.scss#L1140

PR #214, opened 2018-02-28
Avoid scrollbar protruding from the terminal window on wayland session

nana-4:
This is a workaround for #199


clobrano:
Awesome! Thanks :+1:

PR #219, opened 2018-03-02
GTK2 theme now depends on adwaita and pixbuf engines

didrocks:
As the GTK2 theme is based on Adwaita, it's using libadwaita + svg assets.
Ensure the package dependencies reflect this.


clobrano:
Thanks! :+1:

Issue #206, opened 2018-02-26
Wrong hover color in Gnome Logs


yawzub:
I'm pretty sure it's supposed to be darker.
![gnome-logs](https://user-images.githubusercontent.com/36858800/36687768-9a0ea058-1b3b-11e8-95fd-aef3454d18dd.png)

unicyn:
Same for gnome calculator.
![peek 2018-02-27 09-36](https://user-images.githubusercontent.com/5579957/36718457-2590e954-1ba2-11e8-92fc-6cee498a6c1f.gif)

PR #220, opened 2018-03-02
Fix headerbar popup buttons

godlyranchdressing:
Programs that use a popup button in the headerbar (GNOME Logs, Calculator & Calendar) would use the regular button bg color instead of the headerbar button bg color

Closes issue #206


clobrano:
LGTM :tada:

Issue #205, opened 2018-02-26
Gnome Disks menu


yawzub:
This menu is inconsistent with the rest of the theme (separators color, padding)
![gnome-disks](https://user-images.githubusercontent.com/36858800/36687708-6dab6212-1b3b-11e8-8c4d-c017e86e0eee.png)

PR #221, opened 2018-03-02
Fix GNOME Disks using dark separator in a headerbar button's menu

godlyranchdressing:
GNOME Disks seems to inherit a dark separator bg color from the headerbar. Since no unique classnames or ids are being used, I tried to make the rule as specific as possible so it won't affect any other program.

Closes issue #205


clobrano:
LGTM :+1:

Issue #194, opened 2018-02-22
Inconsistency of mouse-over on inactive windows


CDrummond:
Most widgets with communitheme do not react to mouse over events in inactive windows. However some widgets still react/change appearance when the mouse hovers over them in inactive windows - e.g. tabs, 'link button' (URL style label), scrollbars, nautilus pathbar, and the page selector button in gtk3-widget-factory (sorry, don't know the widget's name - but the one in the middle of the header bar that is used to change pages).

IMHO, **all** widgets should still react to mouse-over in inactive windows - as they still accept button press events. e.g. press one of the 'togglebutton' widgets in gtk3-widget-factory in an inactive window - the window is raised, and the button is toggled, therefore the widget is active.

ya-d:
Also discussed in #150.

CDrummond:
Yes/no - this report mentions other widgets that still react when in inactive windows.

But, as stated, I agree that **all** widgets in inactive windows should react - as they still accept the mouse click. If the click simply raised the window, and the widget did not respond - then the current behaviour would be fine. But they do react, when they look like they would not - which is bad. Stating that this is the way Adwaita works, does not mean that it is a good idea.

ya-d:
[I'm totally with you](https://github.com/Ubuntu/gtk-communitheme/pull/166#issuecomment-367441512), just wanted to provide the context to ease management and discussion.

clobrano:
@Luxamman, @madsrh

Luxamman:
The passive windows are tricky, I had to rethink them a little because the separation between active and passive windows is very harsh. 

This is especially irritation when working with two windows side by side (something very common I think) - because passive windows look almost like deactivated or like they hung up.

Like in other systems, it is good to give a hint which window is passive (like windows control orange getting grey and headerbar little lighter, and content colors getting brighter or grey), but maybe we should not separate the whole content and buttons too hard. 

When windows are stacked, they have separation lines and a shadow to be visually separated - so I'm with you to reduce the overall passive look - including buttons and content.

mivoligo:
I'm probably going to repeat what was already mentioned: if the buttons of a background window does not react on hover, I would not expect them to work when clicked. Rather the click will get the window to the foreground.

madsrh:
@CDrummond moving the discussion back here 😉 

[...]

Issue #212, opened 2018-02-27
Separation between active and passive (backdrop) windows is very harsh


ya-d:
As posted by @Luxamman (https://github.com/Ubuntu/gtk-communitheme/issues/194#issuecomment-368857266):

> The passive windows are tricky, I had to rethink them a little because the separation between active and passive windows is very harsh.
> 
> This is especially irritation when working with two windows side by side (something very common I think) - because passive windows look almost like deactivated or like they hung up.
> 
> Like in other systems, it is good to give a hint which window is passive (like windows control orange getting grey and headerbar little lighter, and content colors getting brighter or grey), but maybe we should not separate the whole content and buttons too hard.
> 
> When windows are stacked, they have separation lines and a shadow to be visually separated - so I'm with you to reduce the overall passive look - including buttons and content.

ya-d:
(Quickly wrote down, so probably incomplete…)

*Main issue for me:*
- active entry fields and inactive buttons/fields swap colors: from white to grey/transparent or vice versa

*Minor issues:*
- Terminal changing background colors is rather intrusive
- sliders are not recognizable as active or inactive (proposal coming)
- …?

*Proposal:*
- white stays white (like new check/radiobuttons)
- inactive doesn't need to change at all
- active elements should be recognizable by higher contrast
- does the background really need to change? (same goes for Terminal background)
- maybe all colors should become the same grey value (see new check/radiobuttons) [currently their different even for the same color – maybe resulting from making them more transparent which makes the final color dependent of the background]

+ reconsider/check all colored buttons/elements in their various states

![window in focus](https://user-images.githubusercontent.com/30781876/36734230-29d14244-1bd3-11e8-94fd-bfbb7b98a911.png) 
![window in backdrop](https://user-images.githubusercontent.com/30781876/36734234-2f6a748c-1bd3-11e8-8f58-f2cb7614eef3.png)

ya-d:
Quick and dirty mockup:
![window in backdrop mockup qnd](https://user-images.githubusercontent.com/30781876/36734320-6375d596-1bd3-11e8-9a64-934286d29d01.png)

madsrh:
...Or we could keep the colors and behavior, but add a gray overlay. Here it's Silk #cdcdcd

![backdrop](https://user-images.githubusercontent.com/3986894/36781353-edfc051c-1c75-11e8-8184-458c6ade3266.jpg)

CDrummond:
[...]

PR #222, opened 2018-03-02
Remove tab hover effect in backdrop

godlyranchdressing:
- Removes tab hover effect in backdrop
- Dims icons in backdrop - just barely does anything but it's better than nothing

This doesn't do anything but remove the hover effect as further discussion is being continued in  #212

Closes issue #194


clobrano:
:+1:

PR #234, opened 2018-03-06
Add separators to Nautilus list view

nana-4:
Possibly this may not look good in Communitheme. So I separated this from #233.
image


clobrano:
LGTM :+1: thanks a lot :wink:

PR #248, opened 2018-03-09
Decrease entry field height and image button padding

godlyranchdressing:
- Decrease image button left and right padding to make image-buttons 1:1 after previous size adjustment
- Decrease entry field height to match button height


clobrano:
LGTM :+1:

PR #249, opened 2018-03-09
Update terminal colors

godlyranchdressing:
The old Unity 8 terminal colors were being used in some places, this just darkens and lightens the new (purple) $terminal_bg_color for use for $terminal_borders_color and $terminal_base_color


clobrano:
When on focus, tab has the same color as terminal, what do you think of doing the same when in backdrop?

image

clobrano:
Just one more note. I actually like the current color for inactive tabs (taken from solarized), do you agree to keep it? :smiley:

godlyranchdressing:
It should be the same color. They both use _backdrop_color($terminal_bg_color).  :weary:

clobrano:
Alright, not sure if you did something or the new backdrop colors did, but now it's good :smile: 
image

PR #254, opened 2018-03-12
Switch to bionic

didrocks:
Change as well gtk2 dep with the version in bionic as libadwaita has moved packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
   - docker
 
 env:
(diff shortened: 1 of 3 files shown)


didrocks:
Note: CI will now stall as tzdata is waiting for an answer:
```
debconf: falling back to frontend: Readline
Configuring tzdata
------------------
Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.
  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 
```
We should ignore it for now. I'll have a look to force debconf no-ui next week.

clobrano:
Thanks a lot!

PR #263, opened 2018-03-14
[Issue 261] insensitive_border_color 

Feichtmeier:


clobrano:
Hi @Feichtmeier thanks a lot for your work.
I think you spotted the right place for the issue, but have you considered that all the places where $insensitive_borders_color is used are also affected by this change?

Until now, we know that the issue is in headerbar only, so I suggest you to limit the fix to that section.
Moreover, transparentize(color, 1.0) is equivalent (and more meaningful) to use just none as color for boder-left

Feichtmeier:
Like this?

// More inline toolbar buttons
toolbar.inline-toolbar toolbutton,
toolbar.inline-toolbar toolbutton:backdrop {
  > button.flat { @extend %linked_middle; }

  &:first-child > button.flat { @extend %linked:first-child; }

  &:not(:first-child) > button.flat {
    // linked buttons will have more defined border colors
    border-left-color: $borders_color;
    &:disabled { border-left-color: none;}
  }

  &:last-child > button.flat { @extend %linked:last-child; }

  &:only-child > button.flat { @extend %linked:only-child; }
}

clobrano:
This does not seem to fix the issue. I thought more something like

```
.linked:not(.path-bar):not(.stack-switcher):not(buttonbox) button, .linked:not(.vertical) entry {
      // force linked buttons and entries to be separate
      // this needs a lot of testing
      &:focus {
        & + entry { border-left-color: _border_color($graphite); }
        & + button { border-left-color: transparent; }

        & + entry,
        & + button {
          &:disabled { border-left-color: none; }//$insensitive_borders_color; }
          &:backdrop { border-left-color: _border_color($headerbar_bg_color); }
        }
      }
```
which is inside the header target and then will affect only linked buttons inside headerbar

clobrano:
LGTM :+1: thank you

PR #274, opened 2018-03-22
Sunken entry with orange glow on focus

godlyranchdressing:
Entry field has a sunken look using box-shadow. Focused entry field has a soft orange glow.


clobrano:
Very nice! LGTM :+1:

Issue #251, opened 2018-03-10
Style searchbox


madsrh:
I really don't know what this is called so feel free to rename this issue (or enhancement / styling).

![search](https://user-images.githubusercontent.com/3986894/37240564-9fdc40ba-244d-11e8-920a-bbec3a2d4d63.gif)

Perhaps something like this:

![search](https://user-images.githubusercontent.com/3986894/37240688-3db86b5a-244f-11e8-84b4-a583de70b7ee.jpg)

clobrano:
closed by #275

PR #275, opened 2018-03-22
Fix gedit search field padding

godlyranchdressing:
Gives gedit search entry some padding and a border, otherwise it's right under the headerbar.

Close issue #251


clobrano:
Mockup on #251 adds also a kind of background that I'd like to see. Do you think it's possible?

PR #276, opened 2018-03-22
Rework/backdrop state

### LINKED ISSUE #212: Separation between active and passive (backdrop) windows is very harsh

clobrano:
New backdrop state in order to reduce the difference against normal state.

image

current normal state is
image

closes #212

### DIFF
--- a/Communitheme/gtk-3.0/_colors.scss
+++ b/Communitheme/gtk-3.0/_colors.scss
@@ -64,7 +64,7 @@ $backdrop_fg_color: mix($fg_color, $backdrop_bg_color, 50%);
 $backdrop_insensitive_color: if($variant == 'light', darken($backdrop_bg_color, 15%), lighten($backdrop_bg_color, 15%));
 $backdrop_insensitive_dark_fill: $insensitive_dark_fill;
 $backdrop_selected_fg_color: if($variant == 'light', $backdrop_base_color, $backdrop_text_color);
-$backdrop_borders_color: mix($borders_color, $bg_color, 80%);
+$backdrop_borders_color: darken($borders_color, 3%); //mix($borders_color, $bg_color, 80%);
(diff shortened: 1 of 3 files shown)


godlyranchdressing:
This looks much nicer! I kind of miss the desaturation of colors (orange, green, etc.) though, but I could live without it.

clobrano:
I replaced desaturation with darkening (is it a valid word?) so that it could be used in more places, simplifying the change and improving consistency

madsrh:
Great work @clobrano (that was fast). I can't wait to try this :clap:

didrocks:
Couldn't say better :)

madsrh:
@clobrano Any chance that a color overlay could be added to the icons as well? Currently they have the same color as the active window:

backdrop

clobrano:
Not sure, I'll check this

clobrano:
Sorry for the late reply. I tried using old "active" png for this, but are we sure we want more orange now, considering current trend? :smile:

madsrh:
If that was for me @clobrano, I didn't understand it. I was asking about darker icons in backdrop.

clobrano:
Oh my, too much jumping from bug to bug, totally forgot what you were asking for :smile:

PR #279, opened 2018-03-23
Rework buttons

godlyranchdressing:
- Rework buttons
- Add `outline-offset: -3px` to buttons
- Return button hilight


clobrano:
LGTM, but if we can fix the toolbar button hover and the border hightlight, I'd prefer the previous version

PR #287, opened 2018-03-24
Round out dialog boxes and remove headerbar

godlyranchdressing:
Remove the dark titlebar from dialog boxes
Round out the bottom corners
Add a border for better window distinction

dialog

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1528,7 +1528,7 @@ headerbar {
     background-color: $backdrop_headerbar_bg_color;
     border-color: transparentize(_border_color($backdrop_headerbar_bg_color), 0.7);
     transition: $backdrop_transition;
-    & label { color: $backdrop_fg_color; }
+    &, label { color: $backdrop_headerbar_fg_color; }
   }
 
   .title {
@@ -3978,23 +3978,26 @@ calendar {
  * Dialogs *
  ***********/
 messagedialog { // Message Dialog styling
+  border: 1px solid $borders_color;
+  border-width: 0 1px 1px 1px;
+  border-bottom-left-radius: $small_radius;
+  border-bottom-right-radius: $small_radius;
+
   .titlebar {
-    background-color: $headerbar_bg_color;
+    background-color: $bg_color;
     background-image: none;
-    border-color: transparentize(_border_color($headerbar_bg_color), 0.7);
-    border-style: solid;
-    border-width: 1px;
+    border-style: none;
     border-top-left-radius: $small_radius;
     border-top-right-radius: $small_radius;
-    min-height: 32px;
+    min-height: 0px;
+    border: 1px solid $borders_color;
+    border-width: 1px 1px 0 1px;
 
-    &:backdrop {
-      background-color: $backdrop_headerbar_bg_color;
-    }
+    &:backdrop { background-color: $backdrop_bg_color; }
   }
 
   &.csd {
-    .dialog-vbox { margin-top: 14px; }
+    .dialog-vbox { margin: 2px 0; }
     .dialog-action-area button {
       @extend %button;
       margin: 0 5px 5px;


clobrano:
Why removing the dark titlebar?

godlyranchdressing:
darktitlebar

Looks like a waste of space. The seamless look that even Adwaita has looks much nicer.

adwaita

Feichtmeier:
Though I followed the discussion and the poll on this topic, I agree that the black header bar is a bit pointless. On the one hand it is on par with the rest of the non-shell windows, but on the other hand there is nothing to click, and it is indeed wasted space. Also, I can imagine that people who did not follow the decision could think that there is something broken. Headerbars usually atleast have *something* inside of them to click :)

CDrummond:
Without the headerbar, you cannot (easily) move the dialog. (I know you can use meta+left click, but not as convenient).

clobrano:
I understand your point, but at the same time we specifically put that dark titlebar to have the same look for all the windows. @madshr?

madsrh:
@godlyranchdressing Please backlog [the discussion here](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/262).

I suggested this change because the white dialog doesn't fit well with the rest of the theme and adding a headerbar seems more visually consistent to me.

I still believe that, but I've also had second thoughts about this: My biggest concern, after pushing for this, is that every other OS (_OSX, Windows, Android,..._ ) uses the plain white box 🤔

Can we leave this in for a while and see if we(read I) will be burned on the forum? 🔥I would like to discuss this and [the dark dialogs](https://github.com/ubuntu/gtk-communitheme/pull/153) with Canonical design / UX team, but I think they will be too busy until 18.04 is out.

@clobrano What do you think after using this for 2-3 weeks? I'm guessing you @godlyranchdressing are **-1** 😉 

@Feichtmeier It was very wrong of me to create a poll about this 😞I just wanted to know what people thought would match the theme, but it seems some people believed that we would implement the design with the most votes.

Feichtmeier:
Hm, shouldn't we start from the default settings though? Only because one can move the dialog with a setting the design should not be centered around this option. Those two button dialogues are there for two actions. Accept or cancel. I believe the gnome designers had exactly this in mind when designing those adwaita theme settings.

clobrano:
> I still believe that, but I've also had second thoughts about this: My biggest concern, after pushing for this, is that every other OS (OSX, Windows, Android,... ) uses the plain white box

Is it possible to know whether those plain white box are system notification or theme notification? We still have plain (black) boxes for system ones

> @clobrano What do you think after using this for 2-3 weeks? I'm guessing you @godlyranchdressing are -1 wink

I think I get used to it, and I like it better now

> Without the headerbar, you cannot (easily) move the dialog.

This is my main concern for this change, actually. If you select this option and move the main window what happens? The main window moves and the modal dialog does not?

CDrummond:
@clobrano With these dialogs open, you cannot move the main window - regardless of whether `Attach modal dialogs` is set or not.

clobrano:
Alright then, to me, if the system allows you to move any window, the theme should support it. In fact originally the window was a white box, but with a titlebar

CDrummond:
I would imagine this is not possible to do but... Perhaps if `Attach modal dialogs` is enabled (the default setting), then there should be no titlebar. If its disabled, then there should be a titlebar (and more shadow)?

But, seeing as the theme is CSS, I'd image this is impossible to do.

I'm coming round to the idea of no titlebar. Because, with the default setting of `Attach modal dialogs`, if you attempt to move the dialog then the **whole** window moves.

Feichtmeier:
With the default setting if you move the dialog the window AND the dialog move in one:

peek 2018-03-25 12-15

nana-4:
I really can understand @Feichtmeier's points of his first comment, but somehow I like the current style better than the plain white box.

How about simply reducing the titlebar height?

image

This can save space and should resolve the concern (from @Feichtmeier's comment):

> can imagine that people who did not follow the decision could think that there is something broken.

Feichtmeier:
I have to admit that I still think that the previous version which looks like the shell dialogs, without free space between the buttons and with the wide buttons looked "better" and more to the point. At the moment they look almost like there are some elements missing (to me). I don't know why, maybe I've looked to long on the prev version/shell dialogs :)

madsrh:
@godlyranchdressing please see https://github.com/ubuntu/yaru/issues/698

PR #300, opened 2018-03-31
Fix backdrop dark text color on headerbar

### LINKED ISSUE #292: Titles of inactive non-headerbar windows unreadable
The titlebar text of any window non use GtkHeader bar (e.g. GNOME terminal, any Qt app, any Gtk2 app) is set to a very dark shade of gray - making it impossible to read its title.

Screen shot of GIMP main window and prefences window. As you can see, the title of the main GIMP window is unreadable.

gimp

Note: This is not a Gtk2 issue - the titlebars are drawn by the window manager (which used the Gtk3 theme)

nana-4:
Fixes #292


clobrano:
LGTM :+1: and thanks!

Issue #183, opened 2018-02-20
Menus should have rounded corners to fit suru-style


ghost:
Please add "design"-label.

Menus should have rounded corners to fit the suru-style. An example can be found in the pop-theme.
![menus](https://user-images.githubusercontent.com/36449926/36436975-07ef0d36-1666-11e8-9ea9-667523f7ac03.png)

madsrh:
Well spotted @NusiNusi  +1 from me!

NusiNusi:
is it possible to make the highlight with round corners too? i think i remember a mockup from you with round highlights @madsrh

Feichtmeier:
![screenshot from 2018-03-24 10-47-20](https://user-images.githubusercontent.com/15329494/37862652-e3d1c444-2f50-11e8-9d35-65705d8a7c5c.png)
No matter what I try the border always bleeds out of of the background. Tried several google solutions like

-moz-background-clip: padding; 
-webkit-background-clip: padding-box;
background-clip: padding-box;

Nothing removes the border bleed :( 

Thats' what I did

menu,
.menu,
.context-menu {
  margin: 4px;
  box-shadow: 0 2px 5px transparentize(black, 0.75);
  background-color: $menu_color;
  color: $text_color;
  border: 1px solid $borders_color; 
  border-radius: 10px; // <----------------------------------- That

madsrh:
The United theme has rounded corners as well, so @godlyranchdressing

Feichtmeier:
Ah cool I'll look into that and see how he does it :dagger: I guess he is faster anyways but let's see :)

NusiNusi:
@Feichtmeier The Canta-theme has round menus too: https://www.omgubuntu.co.uk/2018/03/canta-gtk-theme-rounded-corners?utm_source=slideshow

NusiNusi:
Another thing is: the color of checkboxes should be inverted, like in the rest of the theme (box should be green and checkmark should be white). not sure if i should open a new issue for this or we can talk about this here.
![bildschirmfoto vom 2018-03-25 22-06-55](https://user-images.githubusercontent.com/36449926/37935254-1b83410e-3151-11e8-8638-0a8f8a594ca6.png)

madsrh:
@NusiNusi I don't want to derail the conversation to checkboxes 😄 but perhaps we should rather remove the inverted checkboxes? This looks better IMO:

[...]

PR #301, opened 2018-03-31
Make the menu corners round

nana-4:
See #183

This PR also fixes the wrong icon source of arrows in menu.


nana-4:
Please note that the checkboxes and radios in menu are not touched in this PR.

clobrano:
Very nice! Thank you

jnlucian:
The contextual menus should be more rounded to be more consistent with the suru round styling as it has macOS, that would give the communitytheme a more elegant air let's say.
mac-best-menu-bar-extras
pwoer-menu-for-finder-002

Issue #288, opened 2018-03-24
Revert latest button changes?


CDrummond:
Any chance to revert 7579503954598f9fd913b3f0a7654d3e8876ef6f ?? I **really** much preferred the look of the buttons before.

1. The side and top border of buttons is lighter now, reducing contrast
2. There's a very dark line on the bottom - which looks odd
3. On hover the button gets darker. If/when the background is darkened (#259), then I think this would make the hovered button blend in more with the background
4. The previous highlight changed the border, which was sort of consistent with the focus rect, and focus of entry fields.

Now:
![gtk3-now](https://user-images.githubusercontent.com/1141694/37867755-91792f5e-2f94-11e8-9484-b99927d204b8.png)

Before:
![gtk3-before](https://user-images.githubusercontent.com/1141694/37867760-96c97ae0-2f94-11e8-8fe8-cf2fb85bd4a0.png)

clobrano:
Ok, I'm working on another branch to try to fix the border highlight for linked buttons. As @godlyranchdressing said it's complicated :disappointed:

CDrummond:
But the border highlight for linked buttons seems to be just as broken now as it was before these changes. So, I'm not sure what that has to do with reverting this.

Obviously, if more people prefer the current look then fair enough.

clobrano:
> But the border highlight for linked buttons seems to be just as broken now as it was before these changes.

That's right. Of course, having a change of the background on hover, we could remove the change in the borders (which didn't happened in that PR).

So, before showing the current state of this new branch, I'd like to explain why this fix is complicated.
As @godlyranchdressing said, in linked buttons, one of the borders (the right one) does not actually exists, so that buttons look connected. Only the last button on the right has all the 4 buttons, that's why on hover only 3 borders are highlighted.

Possible solutions
A. make the "missing" border appear on hover
B. let the shared border of the next button be highlighted as well
C. mimic border with box-shadow

Solution A and B require the linked buttons to be siblings, which does not happen always (like in page 1 of Widget Factory for instance). Using those fixes effect will cause to have randomly again an incomplete border highlight or,  worse, buttons that **change size** on hover.

![button-border-hover05](https://user-images.githubusercontent.com/2883614/37995564-2cf4ca7e-3215-11e8-9c73-7bb248ed03a8.gif)
![button-border-hover04](https://user-images.githubusercontent.com/2883614/37995566-2f052fe8-3215-11e8-97b5-0492a61c03f0.gif)


[...]

PR #304, opened 2018-04-01
Change the linked border color according to the button states

nana-4:
See #288

Note that due to a technical limitation, we cannot change the border color of combo & tool buttons with this solution.

Some minor things are still lacked (e.g. for linked `osd` buttons), but I push this as a proof of concept.

Feel free to close if you don't like this.


nana-4:
I'm considering another solution. Please wait to merge this.

PR #310, opened 2018-04-03
change color for text selection in GtkTextView

Feichtmeier:
Implementation of https://github.com/ubuntu/gtk-communitheme/issues/307

@nana-4 @godlyranchdressing is there a way to not uncomment this complete line and the text to this class is still changed? Tried many things but this seemed to be the only way - please review and change it if needed & possible

### DIFF
--- a/Communitheme/gtk-3.0/_colors.scss
+++ b/Communitheme/gtk-3.0/_colors.scss
@@ -33,6 +33,7 @@ $success_color: $green;
 $destructive_color: darken($red, 10%);
 $neutral_color: $blue;
 $focus_color: transparentize($selected_bg_color, 0.4);
+$text_selection: #d2effd;
 //
 $osd_fg_color: $porcelain;
 $osd_bg_color: transparentize($jet, 0.3);
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -113,6 +113,7 @@ $button_transition: all 200ms $ease-out-quad;
 
       border-radius: 3px;
     }
+    
   }
 }
 
@@ -121,7 +122,8 @@ textview {
   text {
     @extend %view;
 
-    selection { &:focus, & { @extend %selected_items; }}
+    selection { background-color: $text_selection;
+      /*&:focus, & { @extend %selected_items; }*/}
   }
 }
 


clobrano:
I think you can limit the change to this

``
textview {
  text {
    @extend %view;

    selection {&:focus, & { background-color: $text_selection; }}
  }
``
but let's wait for more suggestion from @nana-4

nana-4:
Are you talking about coding style? If so, IMHO it's okay with simply:

```scss
.view,
textview {
  text {
    @extend %view;

    selection { &:focus, & { background-color: $text_selection; }}
  }
}
```

I think you also need to change other `selection` nodes like:
https://github.com/ubuntu/gtk-communitheme/blob/master/Communitheme/gtk-3.0/_common.scss#L178-L179
https://github.com/ubuntu/sgtk-communitheme/blob/master/Communitheme/gtk-3.0/_common.scss#L301

For Firefox, it seems you need to change here:
https://github.com/ubuntu/gtk-communitheme/blob/master/Communitheme/gtk-3.0/_common.scss#L112

_However_, IMHO the light blue looks quite odd. At least it looks too light/soft to me. Am I the only one who thinks so?

godlyranchdressing:
I've softened up to the blue a lot. Though softening the orange is an option too:
text

madsrh:
As mentioned [here](https://github.com/ubuntu/gtk-communitheme/issues/308), I have nightmares about the toned down orange. It's SO close to salmon on the way to pink. 
It's just my opinion, so if everyone else votes for salmon - sorry, orange, then I'm fine with that. It does match with the theme.

Feichtmeier:
I would prefer a light blue. Might be a bit stronger than this though.

Feichtmeier:
So? Any vote for any color or leave it as it is and close this? :)

- this blue?
- a darker blue?
- leave the current orange?

madsrh:
Anything that isn't the current orange gets my vote - even "samon" is a better choice IMO :D

Feichtmeier:
I vote for light blue (could be the exact same blue as the volume sliders)

nana-4:
> @nana-4 I quiet like the light blue, but I see your point. There's another blue in the new Ubuntu color palette. Would that, in you opinion, work better?

Sorry, I forgot to reply to you. I have no idea which color is the best for now. BTW, did you try this branch?

image

At least for me the combo of orange row selection and light blue text selection looks quite odd.

I want to be cautious for this. :P

clobrano:
> At least for me the combo of orange row selection and light blue text selection looks quite odd.

Probably there won't be many places where this combo can actually exists.

+1 for try it and get feedbacks

madsrh:
> BTW, did you try this branch?

No, so I'm +1 for merging this

Feichtmeier:
@godlyranchdressing are you okay with trying this out and receive feedback?

chrisjbillington:
Here is some feedback from a random user (me)!

The blue seems pretty out of place to the point where I assumed it was as bug when I noticed it. It's true that the previous orange was pretty intense, but only compared to other oranges. It doesn't need to go full salmon to be sufficiently toned down - the orange used in Ambience seems pretty good to me, and I wouldn't mind that remaining the highlight colour.

PR #321, opened 2018-04-10
Build snap version for each commits and PR

didrocks:
Change README to follow new install instructions.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr321`.

PR #323, opened 2018-04-10
red close button only on hover and pressed

Feichtmeier:
Implementation of https://github.com/ubuntu/gtk-communitheme/issues/319

peek 2018-04-10 13-33

@clobrano please review, old code is commented

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4636,11 +4636,12 @@ button.titlebutton {
   min-width: 24px;
   padding: 3px;
 
-  @each $state, $t in ("", "normal"),
+  @each $state, $t in //("", "normal"),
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@clobrano uhm, this is strange :D 
So after recent commit every application does
- light grey on all buttons only on hover
- red cross only on close button only on hover

EXCEPT gnome-tweaks:

peek 2018-04-10 15-54

clobrano:
> What window state is only for gnome-tweaks?

There is no such thing :smile: 

Have you used gtkinspector on gnome tweaks maybe? That would avoid the application to acquire new changes on the theme.

Feichtmeier:
Yes ofc, just though it would be good to get back! WIll do

Feichtmeier:
Merging after email communication

Issue #327, opened 2018-04-11
gthumb has white background where black should be used


pojntfx:
I just noticed the following issue after trying out the latest update (which looks really, really good btw!):

![screenshot from 2018-04-11 22-47-42](https://user-images.githubusercontent.com/28832235/38642509-93e027ee-3dda-11e8-88f9-5cb218b1a284.png)

Due to the white color that's being used the text is hardly readable.

Feichtmeier:
Thansk for the issue could you find out what Gtk class they use there? With the gtk inspector

godlyranchdressing:
It's due to the hardcoded $sidebar_bg_color from f1315c4fb9d48c2d0075c943c86f9e58cf6c616e. @clobrano, you can try using the [adjust-color SASS function](http://sass-lang.com/documentation/Sass/Script/Functions.html#adjust_color-instance_method). 

I think we could probably just restore the old line and add:

`$sidebar_bg_color: adjust-color($sidebar_bg_color, $red: 2);`

That'll add 2 to the red value.
```
$sidebar_bg_color: darken($bg_color, 5%);
$sidebar_bg_color: adjust-color($sidebar_bg_color, $red: 2);
```

Feichtmeier:
Fixed with https://github.com/ubuntu/gtk-communitheme/pull/329

PR #329, opened 2018-04-12
Fix sidebar color

godlyranchdressing:
Uses a function to warm the sidebar color instead of a hardcoded value so the dark theme wouldn't be affected. (#327)


Feichtmeier:
Nice fix:

screenshot from 2018-04-13 11-33-33

PR #332, opened 2018-04-12
Fix window close button

### LINKED ISSUE #314: close button in chrome is scaled up and blurry
community-theme-closebutton-chrome

The community theme currently uses an image to render the close button, which in chrome causes it to appear blurry, as it is being scaled up. Other themes like Pop or Ambiance do not have that issue, as they draw the button without using an image.

### LINKED ISSUE #326: Red highlight of window control close button becomes on (second) hover an ellipse
Red highlight of window control close button becomes on (second) hover an ellipse:
peek 2018-04-11 20-43

@Feichtmeier voila!

nana-4:
- Fix `padding` size
- Remove unnecessary styling
- Stop using `backgorund-size` to avoid a chrome rendering bug
- Replace PNG assets with CSS `-gtk-gradient`

Closes #314
Closes #326

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4634,22 +4634,19 @@ button.titlebutton {
   border-radius: $small_radius;
   min-height: 24px;
   min-width: 24px;
-  padding: 3px;
-
-  @each $state, $t in (":hover", "hover"),
-                      (":active", "active") {
-                        &#{$state} { @include button($t, $c, $tc, $flat: true); }
-  }
+  padding: 2px;
 
   &.close {
-    background-position: center;
-    background-repeat: no-repeat;
-    background-size: 20px;
+    $button_size: 24px + 2px * 2;
+    $circle_size: 20px / $button_size / 2;
 
-    @each $state, $t in (":hover", "-hover"),
-                        (":active", ""),
-                        (":backdrop:hover", "-backdrop") {
-                          &#{$state} { background-image: -gtk-scaled(url("assets/close-button#{$t}@2.png"), url("assets/close-button#{$t}@2.png")); }
+    &:hover:not(:backdrop),
+    &:active:not(:backdrop) {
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@nana-4 Oh hm maybe I merged this too early.
It seems like you also removed the backdrop hover grey squircle + grey circle on X button
Any reason for this? Was this the cause of the bug? =)

nana-4:
@Feichtmeier Oh, was the backdrop hover style intentional? I thought it was also a bug, since all other widgets don't have the backdrop hover style.

Should I/we revert the grey circle? I personally feel it's inconsistent, though.

Feichtmeier:
@nana-4 yes xD 
But I personally don't care. Adwaita has no backdrop hover, but Ambiance has one!
It's a nice addition when you accidently hover off the active window because you are drunk and see "oh no its the backdrop window" and don't close it. Or when you really want not to alt+tab into the backdrop and are a heavy mouse user and just hover over the backdrop window and close it this way without changing focus :dancer: 
I would better cast a wild @madsrh in here and see what he says xD

madsrh:
@Feichtmeier You make a good point, but IMO we want the same as we had before removing the orange circle - no hover effect in backdrop. I believe that was what @clobrano meant too.

madsrh:
@Feichtmeier @nana-4 We're getting sidetrack here. Should we move this discussion to #194?

I'm torn here, because I think the current way is very calm, "uncluttered" or distraction free - whatever you want to call it - but I'll have to admit that my intention [here ](https://github.com/ubuntu/gtk-communitheme/issues/212#issuecomment-370606475)was to have everything behave the same way as the active window (except for the color-overlay). I'm fine with both solutions 😃 

@clobrano @godlyranchdressing ?

clobrano:
I'd like to keep these issues separated. It might mean double work for this close button, but otherwise we will be forced to add backdrop hover effect to other elements as soon as possible.

PR #336, opened 2018-04-17
Remove entry field glow

### LINKED ISSUE #335: Removing orange glow-effect on enterfields
bildschirmfoto vom 2018-04-14 10-15-34

Reasons for removing the glow (the orange border-line stays):
- The theme uses orange mostly in thin clean sharp lines (sidebar soon hopefully too). Removing the glow would make it fit more to the rest of the themes style. See:
bildschirmfoto von 2018-03-29 18-08-05a
- The glow seems to combine orange and transparency. On a white or bright gray background it looks more like salmon or even pink. Same effect (but of course not this strong) we had with selected text color. See: #307 -> this means it looks not so "orangey" how it is intended.

godlyranchdressing:
Closes #335

### DIFF
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -61,11 +61,11 @@
     color: if($tc != $fg_color, transparentize($tc, .1), $fg_color);
     border-color: if($c != $base_color, transparent, $borders_color);
     background-color: $_bc;
-    @include _shadows(entry_focus_shadow(transparentize($fc, 1)), $_entry_edge);
+    @include _shadows(/*entry_focus_shadow(transparentize($fc, 1)),*/ $_entry_edge);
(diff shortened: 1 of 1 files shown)


NusiNusi:
thank you!

Communitheme/gtk-3.0/_drawing.scss
```diff
@@ -61,11 +61,11 @@
     color: if($tc != $fg_color, transparentize($tc, .1), $fg_color);
     border-color: if($c != $base_color, transparent, $borders_color);
     background-color: $_bc;
-    @include _shadows(entry_focus_shadow(transparentize($fc, 1)), $_entry_edge);
+    @include _shadows(/*entry_focus_shadow(transparentize($fc, 1)),*/ $_entry_edge);
```
fitojb: It is not a good practice to comment out code. Just remove it! Trust the version control system.

Issue #289, opened 2018-03-25
Reducing the orange - making orange more elegant


NusiNusi:
Do not worry. I am not really asking for a different highlight-color. But the orange could be reduced. The theme is taking a very classy, clean and mature direction. But this is how the theme can look too:
![bildschirmfoto vom 2018-03-25 22-04-54](https://user-images.githubusercontent.com/36449926/37879430-6896dba0-3079-11e8-89ba-11bb45c9966b.png)
I think it is too much of one color.
The highlight in the nautilus-sidebar is orange. Could it behave more like a clicked button?
![bildschirmfoto vom 2018-03-25 22-05-44](https://user-images.githubusercontent.com/36449926/37879448-95b062fa-3079-11e8-9235-c470880f058c.png)
That would mean a stronger grey. It would be more consistent to how buttons behave even when the clickable elements in the sidebar are no buttons.
The menus have a grey highlight too:
![bildschirmfoto vom 2018-03-25 22-06-55](https://user-images.githubusercontent.com/36449926/37879504-6084b8fa-307a-11e8-911e-9fea2d6e1744.png)
The nautilus-sidebar is the only element with this behavior (getting orange when clicked).

CDrummond:
Other sidebars are orange too - Settings, Tweaks, file chooser.

NusiNusi:
Yes, sorry. I meant sidebars but used the example for nautilus-sidebars. You are right. All sidebars use orange highlights. For consistency: Shouldn't they behave more like pressed buttons?

CDrummond:
I'm not a designer, but (IMVHO) no they should not look like buttons. To me they're like elements from a list, where one is selected - and selected list items have an orange background.

madsrh:
I know that this is the number one criticism when people try the theme. We discussed this earlier and [tried a few things](https://community.ubuntu.com/t/design-mockups-updates-designer-only/2053/10) but landed on this pure orange.

I'll leave it up to the other guys whether this issue should be closed or needs further discussion. IMHO this is a _won't fix_

Feichtmeier:
Most themes, including adwaita which is the base for this theme, don't theme the sidebar as selected toggle buttons. 
____________________________
I would also encourage everyone to not start an issue for every bit that comes into their mind. There is already a lot of stuff to fix. Let the designers do the design and let us try to fix the issues or report actual issues. This would be just my suggestion since we all want to stay on the road, don't we? Closing this now.

[...]

PR #339, opened 2018-04-17
Use orange strip for selected sidebar rows

godlyranchdressing:
To help resolve issue #289

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -176,7 +176,7 @@ label {
     &:backdrop { color: $backdrop_text_color; }
   }
 
-  row:selected &,
+  //row:selected &,
   &:selected { @extend %nobg_selected_items; }
 
   selection {
@@ -3894,14 +3894,30 @@ row {
     &:active { box-shadow: inset 0 2px 2px -2px transparentize(black, 0.8); }
 
     &:not(:backdrop):selected {
-      &:active { box-shadow: inset 0 2px 3px -1px transparentize(black, 0.5); }
+      &:active { box-shadow: inset 3px 0 $selected_bg_color, inset 0 2px 3px -1px transparentize(black, 0.5); }
 
       &.has-open-popup { background-color: mix($fg_color, $selected_bg_color, 10%); }
 
     }
   }
 
-  &:selected { @extend %selected_items; }
+  // &:selected { @extend %selected_items; }
+
+  &:selected {
+    // use orange strip for selected sidebar item
+    background-color: $base_active_color;
+    color: $fg_color;
+    box-shadow: inset 3px 0 $selected_bg_color;
+    font-weight: 500;
+
+    &:backdrop { box-shadow: inset 3px 0 _backdrop_color($selected_bg_color); }
+
+
+    label {
+      transition: padding 100ms linear;
+      padding-left: 3px;
+    }
+  }
 }
 
 
@@ -4100,6 +4116,8 @@ stacksidebar {
       padding-right: 6px;
     }
 
+    &:selected > label { padding-left: 9px; }
+
     &.needs-attention > label {
       @extend %needs_attention;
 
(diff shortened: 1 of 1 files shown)


Feichtmeier:
The implementation looks good to me!
I'd say let's try this. @madsrh ?
screenshot-20180417211441-876x597

NusiNusi:
Shouldn't the text be bold?

NusiNusi:
okay, already merched... too late. @godlyranchdressing necessary to open an issue for the bold text + black symbol? Or will you add it next time automatically?

godlyranchdressing:
@NusiNusi, are you not using the Ubuntu font?

madsrh:
@NusiNusi Without haven't see it, I would say the bold font is too much. IMO we should tweak the colors and not the font weight. Anyway, we can easily experiment with that later 😃

NusiNusi:
We already discussed this:
https://user-images.githubusercontent.com/9556384/38165696-6829f764-3552-11e8-9c98-5e9c4c53f35d.gif
This way it is the perfect equivalent to selected item in stackswitcher. The bold black text and black symbol really should be added.

godlyranchdressing:
@NusiNusi it looks like that for me. I just tested some other fonts. I didn't think through my use of `font-weight: 500` instead of `font-weight: bold`. I was sticking to that because the bold Ubuntu font is super wide, but it looks like if another font family is used without that weight then the font will look the same.

madsrh:
> This way it is the perfect equivalent to selected item in stackswitcher. The bold black text and black symbol really should be added.

@NusiNusi No, the bold font is only used in Nautilus because it is hardcoded. The stackswitcher (like in Software center) **isn't** bold.

We can try the bold font, but I cannot find a single other theme that has bold in the sidebar and again, off the top of my head, I would say it would be screaming for attention - we might as well make it orange 🤣

NusiNusi:
Ah sorry! I just meant it should be as "fat" as the text in stackswitcher or just like the mockup @nana-4 made. It seems i am using the wrong vocabulary :(

Feichtmeier:
I have to admit, that I miss the orange box too allready :| But let's see.
I've just seen the version of the theme some which is some weeks old on my fathers pc and I also miss the orange close button :| Anyone else has this very subjective, sensitive faery feelings? =D

madsrh:
@Feichtmeier Not sure if we're getting off topic here, but IMO the close button ok when the orange block (don't know what to call it) is in the sidebar. So, I share your concerns, but at least we have until october to revert the changes 😆

madsrh:
Is it too early to create a _"revert changes for sidebar"_ issue 🤣 

Jokes aside! As mentioned I'm **not** a fan, but perhaps it's just a matter of tweaking. IMO the current gray used looks slightly disabled. Here's a few quick ideas with less gray:

sidebar

I also see a minor issue with list that has disabled switches like dconf:

New:
image

Old:
image

NusiNusi:
I like the white background with the black text. This way it looks connected to white field the folders are in.
The gradient looks good to. But i would say the gray should be on the left and the white on the right.

Feichtmeier:
Well, no it's not too early. Since this PR is already merged (shame on me -.-) We need another issue and another PR after.
We need atleast an issue which indicates this bugs shown in your gifs and screenshots because otherwise they will get lost!
*(((super offtopic, but please check the community side didi has an important question there regarding the icon size)))*

NusiNusi:
I am still waiting for the update on my snap-install. You too? Or is something broken on my install?

NusiNusi:
I could test it now and... wow... it is amazing how professional and clean it looks:
bildschirmfoto von 2018-04-19 19-03-06
I would only recommend making the text and the icon pure black. @godlyranchdressing made a PR with white background and i am looking forward to test that one too. At the moment IMHO it looks very good and fits to the rest of the style. I would really only tweak with black text and symbol. But lets check the white background first. Thank you @godlyranchdressing

Issue #316, opened 2018-04-05
Context Menus have wrong border-radius


pojntfx:
In Nautilus, context menus have a wrong border-radius sometimes:

![screenshot from 2018-04-06 01-03-36](https://user-images.githubusercontent.com/28832235/38396070-5cc3c494-3936-11e8-89bc-1db3081b6108.png)

This also applies to the desktop (look closely at the top left and right of the hovered item):

![screenshot from 2018-04-06 01-04-12](https://user-images.githubusercontent.com/28832235/38396096-80795764-3936-11e8-9f2c-9a248281506a.png)

clobrano:
Not sure if I understood, what do you mean by "wrong" radius

pojntfx:
Bottom image as example:

How it should be:
Non-rounded | Non-rounded
Rounded | Rounded

How it is:
Rounded | Rounded
Rounded | Rounded

In detail:
![screenshot of problem](https://ibb.co/iZSaEx)

clobrano:
I understood now, you meant on hover :+1:

CDrummond:
I guess its related to this, or eve caused by the fix for this? But the 1st item in a combobox's popup menu has the same issue - in that the bottom corners are now rounded.

clobrano:
uhm, could you give me an example to where to look at?

CDrummond:
From gtk-3-widget-factory:

![combo](https://user-images.githubusercontent.com/1141694/38498683-1c60e7e6-3bfd-11e8-9640-d032d7456231.png)

pojntfx:
Nautilus issue seems to be fixed, but now there is another problem in `gnome-terminal` (same as in Nautilus before):

![screenshot from 2018-04-09 15-15-13](https://user-images.githubusercontent.com/28832235/38499791-f1b1b262-3c08-11e8-832e-a87b69314b51.png)

I can also confirm @CDrummond's problem in GTK3 widget factory.

CDrummond:
Should this issue not be re-opened? Or shall I create a new issue for the combobox menu problem?

Feichtmeier:
Seems fixed here, pulled today:

![screenshot from 2018-04-15 23-50-30](https://user-images.githubusercontent.com/15329494/38783814-e94cc46e-4107-11e8-887e-bd2e384ec214.png)

CDrummond:
[...]

PR #340, opened 2018-04-17
Fix context menu border-radius on hover

godlyranchdressing:
Not sure if the altered line was an isolated case, but it resulted in the wrong radius being used in some corners in most cases. #316 

Before:
wrongradius
After:
merge

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -43,6 +43,13 @@ list.tweak-categories separator {
   notebook > stack grid:not(.nautilus-list-view) {
     padding-top: 15px;
   }
+
+
+  menu menuitem:first-child:hover {
+    // context menu in nautilus has its last-child labeled as its first child lol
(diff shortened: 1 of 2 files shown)


clobrano:
It looked like the context menu had first child at the bottom of the list. This was the lines changed with this PR. There was a ticket about it i could not find now

clobrano:
With this change we fall back into #316: in nautilus the hover has wrong borders in the row at the bottom. It look like some menus uses last-child as top row and some others do the other way around.

godlyranchdressing:
@clobrano I added an exception for nautilus. I did some testing and I didn't run into any problems in any other menus in nautilus either so it looks like a safe fix.

godlyranchdressing:
@clobrano I added an exception for nautilus. I did some testing and I didn't run into any problems in any other menus in nautilus either so it looks like a safe fix.

clobrano:
Awesome! Accepting it then :+1:

Issue #302, opened 2018-04-01
Toggled item in header bar not clearly visible


CDrummond:
The currently toggled item in a GtkHeader bar is not clearly visible/highlighted.

e.g. in this screenshot the magnifying glass icon has been pressed to produce the search field. However, its state is not indicated very clearly - dark on dark is not the best indicator. Perhaps it should have a lighter background?

![screenshot from 2018-04-01 09-58-48](https://user-images.githubusercontent.com/1141694/38171533-61f5288c-3593-11e8-8017-f24e82ecfd8c.png)

CDrummond:
This issue was originally raised by Kvantum's author. He's currently working on a matching Qt style. To mimic headerbars he's suggesting that the main toolbar in a Qt's app have a dark background. Hence the issue with toggled items not being clear. Currently he's implemented the following solution:

![screenshot from 2018-04-01 14-15-13](https://user-images.githubusercontent.com/1141694/38173503-555422d0-35b7-11e8-96a5-1d7be8b9422f.png)

where the toggled items have the orange underline. Not sure if this matches up with waht Communitheme is aiming for?

tsujan:
I completely agree with @CDrummond; `Ambiance` had a similar problem. The header-bar is elegant. The hover color of buttons is OK too. But the pressed state can't be seen clearly. I don't know whether that could cause a problem in GTK3 apps (can their header-bar buttons have a toggled state?) but a matching Kvantum/Qt theme with that pressed state would be a disaster ;)

madsrh:
I don't know if you followed [this discussion](https://community.ubuntu.com/t/mockups-new-design-discussions/1898/220) earlier, but yes I agree that current header-bar buttons still isn't very easy to spot when pressed. 

I suggested a orange line [back in november](https://community.ubuntu.com/t/design-mockups-updates-designer-only/2053/13).

Another option would be to keep the gray hover color and simple add the inner shadow / 3D look when pressed. I'm sorry for this very rude mockup, but hopefully you get my point:

![asdf](https://user-images.githubusercontent.com/3986894/38178630-21c9e55c-3617-11e8-9487-84c9e1110d64.gif)

tsujan:
> I suggested a orange line back in november.

That's very nice.

> Another option would be to keep the gray hover color and simple add the inner shadow / 3D look

IMHO, this isn't elegant but tastes differ.

Feichtmeier:
[...]

PR #342, opened 2018-04-17
Just add shadow to active headerbar buttons when active

godlyranchdressing:
As requested [here](https://github.com/ubuntu/gtk-communitheme/issues/302#issuecomment-382148039) for dealing with issue #302 

peek 2018-04-17 17-36


madsrh:
Awesome! 👍 
Wouldn't you agree that the black border should be removed too @godlyranchdressing ?

godlyranchdressing:
@madsrh I don't know. Isn't it important to retain the pushed button effect?

clobrano:
I think it would be better without the black border in both pressed and active state, or with a black that blend more with headerbar black

PR #353, opened 2018-04-22
Use extend for selected text

godlyranchdressing:
Mentioned [here](https://github.com/ubuntu/gtk-communitheme/issues/307#issuecomment-378414227). Includes a tiny fix for checks and radios sometimes using $text_color when hovered on in treeviews.

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -109,10 +109,8 @@ $button_transition: all 150ms $ease-out-quad;
 
   &:selected {
     &:focus, & {
-      @extend %selected_items;
-      //Trying blue for the text selection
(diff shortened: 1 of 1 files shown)


clobrano:
Hi @godlyranchdressing, this LGTM, but just one question, LibreOffice text selection color is the one below. Is it the best match we can have with the light blue?

image

godlyranchdressing:
@clobrano I think so. It looks like LibreOffice is doing something of its own to make the selection color lighter.

clobrano:
Fair enough :+1:

Feichtmeier:
@godlyranchdressing something went really wrong in this PR or the one before. Could you please double check the view selection in nautilus again? Its dark orange with black text now

PR #355, opened 2018-04-23
Added style for default button

clobrano:
In dialog widows, default button is styled with a darker border.

Moreover selection ring style changed from solid to dashed so that when a
button is both selected and default the superposition of the two
lines (orange and ash) is not too strong.


Feichtmeier:
Pretty straight forward easy change no reason to not merge it =)

PR #359, opened 2018-04-24
Made more evident when a button is active in headerbar

clobrano:
Removed box-shadow from headerbar active button


clobrano:
Closing it since new design landed very well :)

PR #360, opened 2018-04-24
Added insensitive state to special buttons

### LINKED ISSUE #152: Save and open dialog header buttons
Please, set save and open dialog header button's color to green (Open, Save) and gray (Cancel) because it's not easy to find the only text buttons.

clobrano:
This commit gives success and destructive button a darker background
when in insensitive state

closes #152


clobrano:
Can't see the error log

Issue #260, opened 2018-03-13
Borderless buttons


CDrummond:
Some buttons (toolbar only?) do not have a border on hover/mouse-over or when pressed. For example see LibreOffice's toolbars, or the hamburger menu in Geary's message view (screen shot below). This makes the pressed/selected buttons look slightly blurry.

![borderless-button](https://user-images.githubusercontent.com/1141694/37336382-1536535c-26a9-11e8-9b81-b44ba81ef2a2.png)

This also affects other buttons - such as the `Reset to Default` in nautilus' preferences dialog (in the `List Columns` tab)

In Qt speak these buttons would be called `Auto Raise` - i.e. buttons that initially have no border, background, etc, until hovered or pressed.

Feichtmeier:
I am not that sure, but I think this is wanted. @madsrh  ? :)

CDrummond:
Are you stating that the effect of **not** having a border is wanted? Why would that be? **Every** other widget has a border, so why should this not?

clobrano:
> Some buttons (toolbar only?) do not have a border on hover/mouse-over or when pressed.

Flat buttons do not have borders. It happens that most of flat buttons are in toolbar and headerbar.

When pressed, the border is given by box-shadow, when hovered there should be a color change. Now, since most of flat buttons are in dark headerbar and toolbar those have been tested, while the ones on bright backgrounds might need some more refinements

CDrummond:
Why do flat buttons have no border? I can understand whilst they are flat, but once they are pressed, or mouse-over, then why not just draw as per non-flat buttons.

The button in the screenshot above looks awful (IMHO), and would look nicer with a border.

clobrano:
The border might ruin the pressed effect, but let me double check  with some tests. (of course, if you can develop something, an help is welcome)

CDrummond:
I wish I could - if it was coding then I'd gladly help. Unfortunately, I'm next to useless with CSS :-(

How would a border ruin the pressed effect? It'd just look the same as a non-flat pressed button, no?

madsrh:
I'm also looking into this at the moment

godlyranchdressing:
Maybe this is a bit too much? I just made:

```
@if $flat {
      border-color: if($c == $headerbar_bg_color,$_pressed, $_bg);
    }
```

```
@if $flat {
      border-color: $_pressed;
    }
```

[...]

PR #366, opened 2018-04-25
New pushed button effect

godlyranchdressing:
[Seen here](https://github.com/ubuntu/gtk-communitheme/issues/302#issuecomment-384310559). 

It's still open for discussion and finetuning. No rush, since I'm not sure if this fits into what @ya-d was looking for and for some reason line 254: 

`border-color: if($c==$headerbar_bg_color, lighten($_border, 5%), $_border);` 

Doesn't seem to work and I'm not sure why. $c _is_ $headerbar_bg_color.


Feichtmeier:
Maybe set a second variable? Not sure about how powerful sass is there since you use the lighten function in the if

clobrano:
> for some reason line 254:
>     border-color: if($c==$headerbar_bg_color, lighten($_border, 5%), $_border);
> Doesn't seem to work and I'm not sure why

What is the expected behavior? Considering $_border definition at the top

    $_border: if($c == $button_bg_color, $borders_color, darken($c, 10%));

I would expect $_boder to be like `darken($headerbar_bg_color, 5%)`

Feichtmeier:
God I love this, really. Sorry for for much emotion but I think this gives the theme the final cut 
peek 2018-04-26 12-33

@godlyranchdressing that review: carlo is the one for this, I don't even understand everything you did there =)

CDrummond:
I like these changes too - I even think that with this change both #260 and #288 can be closed. Especially #260 - as with less of a shadow, the blurry-ness goes.

godlyranchdressing:
Nevermind, I didn't spot $c being reassigned to something else and making my if statement conditions wrong. I just moved my statement earlier and it's working now. This should be okay to merge if everyone is fine with the outcome:
b
a

PR #371, opened 2018-04-27
Added window vertical border in maximized state

clobrano:
This change completes https://github.com/ubuntu/gnome-shell-communitheme/pull/125

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1767,7 +1767,7 @@ headerbar {
   .tiled-left &,
   .maximized &,
   .fullscreen & {
-    box-shadow: inset 0 1px $inkstone;
+    box-shadow: inset 0 1px $inkstone, inset 1px 0 $inkstone;
 
     &:backdrop, & {
       border-radius: 0;


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr371`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr371` channel to test your snaps with those changes.

godlyranchdressing:
If this is to separate the dock from the maximized window then I think it should be part of the Shell's CSS. You can have the dock on the left, right or bottom of the window and a window has no way of knowing which side the dock is on. The Shell has .left #dash, .right #dash etc. and the box-shadow can be drawn under those classes instead.

clobrano:
You're right, but this is a completion of the line we use for the top panel and it's part of the headerbar. At the bottom it's not needed, eventually we might think of adding it at the right side

Feichtmeier:
@clobrano it looks like you missed some of the headerbars in this great design improvement:
(look at the line in the headerbar:)
Geary [working]:
image
Software [not working]:
image
Videos/Totem [not working]:
image
Tweaks [working]:
image
Music [working]:
image

I fail to see any connection since they are all headerbar :face_with_head_bandage: 
Maaaaybe: the ones that work have a symbolic button right  in the corner of the headerbar ? Oo

clobrano:
Videos is probably the only one using the dark variation

clobrano:
Yes, Videos refers to gtk-dark.css where border-left is 0px. I'm on it

image

PR #379, opened 2018-04-28
Round titlebuttons, orange close button

godlyranchdressing:


Feichtmeier:
I guess there are still commits coming?
image
I pretty hot for this :O

godlyranchdressing:
What the hell. That wasn't happening just a minute ago.

Feichtmeier:
@madsrh how about trying the following pressed header bar button in the change/thisPR of circle close/max/min buttons (so squircle/circle won't be there):
screenshot-20180429102941-1252x894

Ads20000:
After reading [this post](https://design.canonical.com/2016/05/colour-palette-updates/) (and other Unity 8 design posts) I'm not sure about making the orange always visible, it is rather attention-grabbing, hmm... Up to you lot I guess, just offering thoughts... I think orange (or potentially red) on hover makes sense... but the background grey squircle and orange circle doesn't make much sense to me. How about attempting an orange squircle? :)

godlyranchdressing:
@Feichtmeier I've solved the backdrop issue so we can merge this if everyone is okay with it.

Feichtmeier:
Yes let's try it

clobrano:
We might have lost a bit of pressing area here, which makes press these buttons less easy

madsrh:
Could someone post a screenshot?

> We might have lost a bit of pressing area here, which makes press these buttons less easy

We want touch friendly sizes 😃

godlyranchdressing:
It's the same size as the other buttons. 

Using box-shadow to show the actual size of the button:

title

yousuc:
I personally prefer the rounded square vs the circles. Looks to osX'ish and too much like current ubuntu default.

Issue #376, opened 2018-04-28
Bring back highlight for selected items


zappedfish:
I hope this would not be a duplicate of another issue...

Some sort of highlight is really needed for two reasons:
- Right now it's hard to spot what exactly is selected (takes a little bit more time to spot that orange stripe)
- If you use keyboard for navigation it becomes very confusing (see screenshot below)

![screenshot from 2018-04-28 18-17-11](https://user-images.githubusercontent.com/36744029/39397455-68a92128-4b10-11e8-806a-d30bf182a0e7.png)

It looks like "Default Applications" item is active, while it's actually "About"

Feichtmeier:
While many agree with you for bringing it back - the bug or wrong behaviour you mentioned does not occur on my installation.

Feichtmeier:
@zappedfish np the general issue comes to the right time =D Because there is some hot discussion going on about bringing it back for other reasons

ya-d:
I can confirm this issue for Files (updated 4-27).
* Files: pressing <tab> moves the selection highlight but content stays where it was. (As seen in first post for Settings.)
* Settings: pressing <tab> does nothing for me. Only search in the top left corner receives the keyboard focus ring. (But it seems to show the same issue as Files in @Feichtmeier's gif.)
* Tweaks: hitting <tab> moves selection highlight _and_ content view -- that's fine.

So this is clearly a bug for Files and Settings. Although it's not entirely caused by Communitheme but upstream Gnome or Apps! (Happens with Adwaita as well.)
(a) <tab> should only move keyboard selection highlight, _not_  active selection highlight (or at least content should move too). => Upstream, should be reported there
(b) the focus ring should be closed on all sides => Communitheme
(c) I wouldn't mind trying @godlyranchdressing's proposal of keeping the grey hover highlight next to the vertical orange line for selection highlight before deciding to go back to plain orange. => Communitheme

Feichtmeier:
Ah thanks for diving into this @ya-d 
So @zappedfish could report this upstream and change the text of the initial post a bit so it fits to a discussion about the sidebar row selection

PR #380, opened 2018-04-28
Add bg color to selected list row

godlyranchdressing:
Addresses #376

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -3858,26 +3858,7 @@ row {
     }
   }
 
-  // &:selected { @extend %selected_items; }
-
-  &:selected {
-    // use orange strip for selected sidebar item
-    // background-color: $white;
-    color: $fg_color;
-    box-shadow: inset 3px 0 $selected_bg_color;
-    background-color: transparentize($base_hover_color, 0.05);
-    font-weight: bold;
-
-    &:backdrop { box-shadow: inset 3px 0 _backdrop_color($selected_bg_color); }
-    &:hover { background-color: $base_hover_color; }
-
-    image, image.sidebar-icon { opacity: 1; }
-
-    label {
-      transition: padding 100ms linear;
-      padding-left: 3px;
-    }
-  }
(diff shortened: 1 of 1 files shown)


Feichtmeier:
We need a solution for all of these selected sidebar rows. They are also used in the software-center as screenshot thumbnails:
image

clobrano:
@Feichtmeier that is probably due to the image being too close to the border of the row. I feel safe in accepting this change, selected row is at more visible now. What do you think?

Feichtmeier:
Sure I like all of the selected rows we had (including the orange "Blob", the orange stripe and now this)
However: how should we fix this problem? I think it would be rather un-smart to Style the sidebar in a way that does not fit for every place it is being used. Otherwise we end up theming for specific apps
That's why I thought this would be a good timing or targeting a solution that works everywhere

godlyranchdressing:
> We need a solution for all of these selected sidebar rows. They are also used in the software-center as screenshot thumbnails:

:sob: 

We can use `.image-list row` to make the software center ignore the orange strip, but who knows if there's some other app that will have a similar problem. The only other sensible alternative is to bring back the orange background color for selected rows.

> I think it would be rather un-smart to Style the sidebar in a way that does not fit for every place it is being used

I agree. It should "just work"

clobrano:
> who knows if there's some other app that will have a similar problem.

That's the point. Probably, the easiest solution (orange selection) would be the one that fit best

Feichtmeier:
We could use the full force selection again, yet with a different tone of orange? Like proposed in the selected text discussion?

clobrano:
That is a possible solution, but maybe this time I'll try to produce a list of mockups to compare each other before going on the real thing

godlyranchdressing:
@Feichtmeier I think it'll be safe to use the regular orange (#E95420) for the selected list row. If we so prominently use a different shade of orange then that'll open up the "then why don't we use it here too?" question and we'll get stuck again.

Feichtmeier:
As long as we do not have a better solution that really works for every app I agree! @madsrh ? +1 -> merge
image

Issue #363, opened 2018-04-25
Color choice for selected elements


godlyranchdressing:
Just a semi-serious proposal. 

Selected treeview elements using $base_hover_color:

![treeview](https://user-images.githubusercontent.com/27529229/39254548-531291a4-4878-11e8-83c4-c951a2e96521.gif)

clobrano:
I agree that in treeview orange might be unnecessary

Feichtmeier:
I would make the selected text bold :octopus:

godlyranchdressing:
@Feichtmeier I hate that this is becoming my motto recently but: I tried, but it looks like you can't. 

`treeview { font-weight: bold }` works, but `treeview:selected { font-weight: bold }` does nothing.

Feichtmeier:
@godlyranchdressing not that big of a problem though imho :)

Feichtmeier:
![image](https://user-images.githubusercontent.com/15329494/39370924-7608b7f0-4a40-11e8-901c-c770dbfdeccc.png)

Geary uses this, too =) Looks kinda weird

godlyranchdressing:
Orange sidebar row + gray treeview row it is. :grin:

Luxamman:
Like already discussed (in 307) and said, it looks just deactivated in some circumstances. If we like to go for the blue selection, we should go for the blue selection. 
Also like said before, I would still like the blue to be more saturated because the contrast to white is sometimes not big enough (will make problems on cheaper displays, sunlight and so on).

clobrano:
I've never expressed an opinion about that, but can we develop a bit more the idea of shades of orange for selection? Agree that salmon is not good, but we are constantly proceding towards no orange anymore in the theme...

Feichtmeier:
I would recommend if we check the stuff on more apps before we change anything.

This was my personal failure (agreeing at the top here, reverting the PR then because I've seen where this change would hit) but I think you guys are also walking into these traps. 

A lot of GtkClasses are used in the most exotic places you would never think they exist.
((For example the new/proposed sidebar which is used in nautilus and the settings app is also used in the software center in the screenshot thumbnails and there the border is being cut off. I will open/reopen an/the issue for this))

___

On topic: 
A) leave it as it is. It does not look too bad on most gtk classes. <--- f
B) Blue @Luxamman mentioned --> recognizable yet unobtrusive

godlyranchdressing:
[...]

PR #388, opened 2018-04-30
Lighter orange for selected text

godlyranchdressing:
Uses #FA794B. 
Addresses #363. 

Seen [here](https://github.com/ubuntu/gtk-communitheme/issues/363#issuecomment-385286238).


madsrh:
@godlyranchdressing Didn't you just comment this:

> I think it'll be safe to use the regular orange (#E95420) for the selected list row. If we so prominently use a different shade of orange then that'll open up the "then why don't we use it here too?" question and we'll get stuck again.

Do you still want to suggest #FA794B for selected text?

godlyranchdressing:
@madsrh yep. The orange in text selections aren't as prominent IMO.

PR #389, opened 2018-04-30
Infobar buttons

godlyranchdressing:
Also use blue instead of purple for some prompts:

screenshot from 2018-04-30 11-45-10
screenshot from 2018-04-30 11-44-51

trash

trash2

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4270,9 +4270,9 @@ paned {
  **************/
 
 infobar {
-  @each $t, $c in ("", $purple),
-                  (".info", $purple),
-                  (".question", $purple),
+  @each $t, $c in ("", $blue),
+                  (".info", $blue),
+                  (".question", $blue),
                   (".warning", $warning_color),
                   ('.error', $error_color) {
     &#{$t} {
@@ -4291,13 +4291,14 @@ infobar {
       }
 
       button {
-        $c: darken($c, 10%);
+        $c: $c;
         $tc: white;
           @each $state, $t in ("", "normal"), (":hover", "hover"),
             (":active, &:checked", "active"), (":disabled", "insensitive"),
             (":backdrop", "backdrop"), (":backdrop:active, &:backdrop:checked", 'backdrop-active'),
             (":backdrop:disabled", 'backdrop-insensitive') {
-              &#{$state} { @include button($t, $c, $tc, $flat:true); }
+              @if $t == "normal" { $c: darken($c, 5%); }
+              &#{$state} { @include button($t, $c, $tc); }
           }
         }
       }
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4270,9 +4270,9 @@ paned {
  **************/
 
 infobar {
-  @each $t, $c in ("", $blue),
-                  (".info", $blue),
-                  (".question", $blue),
+  @each $t, $c in ("", $purple),
+                  (".info", $purple),
+                  (".question", $purple),
                   (".warning", $warning_color),
                   ('.error', $error_color) {
     &#{$t} {
@@ -4297,7 +4297,6 @@ infobar {
             (":active, &:checked", "active"), (":disabled", "insensitive"),
             (":backdrop", "backdrop"), (":backdrop:active, &:backdrop:checked", 'backdrop-active'),
             (":backdrop:disabled", 'backdrop-insensitive') {
-              @if $t == "normal" { $c: darken($c, 5%); }
               &#{$state} { @include button($t, $c, $tc); }
           }
         }
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -282,8 +282,11 @@
         $_bg: $bg_color;
     } @else if $c == $headerbar_bg_color {
         $_bg: $c;
-    } @else {
-        $_bg: darken($c, 20%);
+    } @else if $c == $red, $green, $yellow, $blue, $purple {
+        $_bg: transparentize($c, .1);
+    }
+      @else {
+        $_bg: darken($c, 20%); 
     }
 
     background-color: $_bg;
@@ -292,7 +295,7 @@
       border-color: $_bg;
       box-shadow: none;
     } @else {
-      border-color: $insensitive_borders_color;
+      border-color: if($c!=$button_bg_color, $c, $insensitive_borders_color);
       box-shadow: 0 1px transparent;
     }
 
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -219,7 +219,7 @@
 
     @else {
       border-color: $_border;
-      border-bottom-color: darken($_border, 20%);
+      border-bottom-color: darken($_border, if($c == $headerbar_bg_color or $c == $button_bg_color, 20%, 12%));
       box-shadow: 0 1px transparentize(black, 0.9);
     }
   }
@@ -239,7 +239,7 @@
 
     @else {
       border-color: $_border;
-      border-bottom-color: darken($_border, 20%);
+      border-bottom-color: darken($_border, if($c == $headerbar_bg_color or $c == $button_bg_color, 20%, 12%));
       box-shadow: 0 1px transparentize(black, 0.9);
     }
   }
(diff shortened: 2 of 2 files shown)


Feichtmeier:
Let @madsrh double check uppon this so we don't have to revert anything if it's wrong =D

madsrh:
@godlyranchdressing Isn't the disabled button taking a lot of focus? How about making it a bit more discreet?

image

Feichtmeier:
Me too. I think the purple was not bad. But the yellow is ofc unthemed

madsrh:
@Feichtmeier Yellow looks a bit odd here, but you it's so normal [for info bars](https://github.com/ubuntudesign/vanilla-design/blob/master/Notifications/notifications.png).
Adwaita has no background at all. What about something like this or is it too gray, dull and sad?

image

Feichtmeier:
:crying_cat_face: please let's not go the grey road again
Maybe I wrote this a bit strange: what I meant was -> yellow is ofc the wrong color -> let's replace it with purple and have all info bars purple. Purple is not used at all otherwise. Despite the log in screen. Blue is used for progress-bars and sliders. But I am bad at this, I don't want to decide this - you are the designer!!111 ;D

clobrano:
Sorry for the silly question, but are the borders and shadows black/grey? If so, I think they might be more soft if using a dark blue/purple/yellow

godlyranchdressing:
@madsrh that gray looks sad (which might be appropriate for files about to be deleted, but still). Some notes for possible color options:

The infobar for Nautilus' trash bin is `infobar`, which I think doesn't use any color at all in Adwaita. Plain infobars using a color is new to the communitheme. Right now `infobar`, `infobar.info` and `infobar.question` use purple while `infobar.warning` and `infobar.error` use yellow and red respectively. 

Part of the reason why I like the purple is _because_ it isn't used anywhere else in the theme but is still an "Ubuntu color" so it also isn't totally removed.

@clobrano they are a dark blue/purple/gray. :laughing: I used the regular button mixin so the colors just happen to be calculated like that. Should we allow colored buttons to have lighter border colors? I do think that destructive buttons go a bit too dark. I'll post some mockups later today.

madsrh:
@godlyranchdressing I'm **perfectly fine** with the purple 👍  I've tried a lot of options and of all of them, purple is a good choice.

I still think the disabled buttons should be less intrusive (see above) 😉

godlyranchdressing:
Lighter border-bottom and border-top colors:
zpeek 2018-05-02 14-19
wpeek 2018-05-02 14-18
peek 2018-05-02 14-18

ya-d:
**1. Suggested action buttons:** Could the same changes be applied to suggested action buttons too? They are still flat which (in combination with rather low contrast for the text) makes them somewhat look disabled to me. I think the effect is now subtle enough to also have it on header bar. Or would that be in conflict with design decisions (flat buttons in header bar)?

_// Thinking one step further: Could/Should even the hover effect for regular header bar buttons have the same 3-dimensional look?_

orange text selection - save dialogue

**2. GDM buttons:** This is probably the wrong place but I didn't know where else to report it. GDM buttons could use the same treatment.
gdm buttons

Feichtmeier:
And please not more 3d. The flat hover fits very well.

Feichtmeier:
Isn't this PR finished? Even is @godlyranchdressing is a little afk atm this seems ready to merge for me? @clobrano @madsrh ?

clobrano:
**ads2000** [comment](Ubuntu Community Hub
https://community.ubuntu.com/t/call-for-participation-an-ubuntu-default-theme-lead-by-the-community/1545/1083) on the HUB made me think a little. Since the usage of any color seems a bit forcing (if I get it right), why not use colors already present in Nautilus? Like the color behind tabs or the one in header when using list view?

Feichtmeier:
Like adapt them to the app notification colors (white/porcelain)?

clobrano:
We use white also for "search", so it might be a solution, even if it is a widget inside the active view, while infobar is outside

image


 otherwise the sidebar color which we use for inactive tabs as well

image

Feichtmeier:
38277809-bbc73212-3767-11e8-878c-25370292e48c
Here for reference.
We could use a box-shadow too

clobrano:
Something like this. Just an idea though, re-reading the thread, I saw that the agreement on purple was reached.

image

Feichtmeier:
I am a bit torn.
On the one hand purple could be good choice because this is something between a notification (white) and and activiy (blue) so the unused purple could make sense.
On the other hand there is not much difference between these info bars and shell/-appnotification with buttons and a dialog so white would be the most obvious choice.
If we stick with the current color scheme, blue feels wrong though. It is no progress and no real activity.

clobrano:
In Nautilus I see this infobar not different than tabs or search bar, it is just a "toolbar" let's say. However, infobar are used elsewhere, only I don't remember where, and in other application white might be the wrong color, then sorry for the noise :)

chrisjbillington:
> On the other hand there is not much difference between these info bars and shell/-appnotification with buttons and a dialog so white would be the most obvious choice. But white is quiet hard to see.

...well minds keep changing about whether notifications and shell dialogs should be white or dark...what if they were purple? Then the info bars could be purple *and* be the same colour as the shell notifications. Don't know whether purple in shell dialogs would look completely ridiculous, but I'm quite fond of the purple info bars.

Feichtmeier:
@chrisjbillington the white notifications will stay as far as I understoog @madsrh  and @luxamman they like them. Purple shell notifications would look very strange and alien to me.

I just found out while trying the dark theme a bit more that these infobars are quiet often used in PitVi:
image

clobrano:
So, since info, warning and error have all color well defined, I'll give blue, yellow and red a chance

Feichtmeier:
@clobrano tbh you are right. Purple just does not make much sense now after we've gone very heavy in the unity8 direction other than in the GDM screen. Sorry for changing my mind so easily (when good arguments pop up why not=) ? )

ecf799316df0f1c3eb7c3c13a3d772a3290cec89

godlyranchdressing:
So we're moving from the purple to the blue?

Feichtmeier:
Imho it would make the most sense from the arguments we've seen in here. I also like the purple but it does not really fit the current color palette we've taken from unity 8

madsrh:
> So we're moving from the purple to the blue?

If so, does anyone have another suggestion than [this one](https://github.com/ubuntu/gtk-communitheme/pull/389#issuecomment-385507266)? I know it makes sense, but this isn't really pretty IMO.

🤔

Feichtmeier:
It's also used in pitivi:
https://github.com/ubuntu/gtk-communitheme/pull/444

clobrano:
> So we're moving from the purple to the blue?

Just an idea. I think however that if we don't find an agreement, better keep things as they are or we end up as with buttons :smile: 

> If so, does anyone have another suggestion than this one? I know it makes sense, but this isn't really pretty IMO.

What about [this](https://github.com/ubuntu/gtk-communitheme/pull/389#issuecomment-388790955)

ya-d:
Nothing really is wrong with purple, most of you seem to like it and no one complained afaik. As long as there is no clear agreement and coherent reasoning I wouldn't rush changes.* Purple is used for gdm and terminal so it's not alien to the theme. It's also Ubuntu's well known secondary colour.* Blue is actually in the palette but it's currently used exclusivly for activity which doesn't match this use case. It's also very cold and doesn't fit the design language very good.* White/grey could be an option but it's very natural so it might not be a good fit for info bar. Might lead to issues for some apps where info bars should be visible more prominently.Logic didn't pave a way so far imho, which is why I would stick with taste for the time being - and I subjectivly like the use of purple. 😊Am 21.05.2018 09:36 schrieb Carlo Lobrano <notifications@github.com>:
So we're moving from the purple to the blue?

Just an idea. I think however that if we don't find an agreement, better keep things as they are or we end up as with buttons 😄

If so, does anyone have another suggestion than this one? I know it makes sense, but this isn't really pretty IMO.

What about this

godlyranchdressing:
@clobrano 
> Just an idea. I think however that if we don't find an agreement, better keep things as they are or we end up as with buttons smile

I agree, looks like we're all on the same page at least for now.

I forgot that since I reversed the color change, all this PR does is give the infobar buttons a much needed finetuning so I'll just go ahead and merge this.

madsrh:
The buttons are currently really hard to spot because the blend too much with the background.
image

## I would suggest one of these:

A lighter background (something midway between the purple background and the hover color)
image

Or simply use regular buttons
image

clobrano:
Maybe the second one is simpler to work on when in hover? The first has already a quite bright purple in normal state

madsrh:
> Maybe the second one is simpler to work on when in hover? The first has already a quite bright purple in normal state.

**+1** Using a standard button would also help keep it simple 👍 

BTW, The purple color works nicely here 🎨

Feichtmeier:
That normal button is a great idea. Draws the whole look a bit back from "alien"

PR #392, opened 2018-05-02
Connect menubar with headerbar/titlebar

### LINKED ISSUE #391: Change background colour of menu titles
Creating an issue at the request of @madsrh.

Two options are under discussion:

- Tweak the grey to be the shade currently used for the out-of-focus window header bar (reasoning and mock-ups are in https://community.ubuntu.com/t/call-for-participation-an-ubuntu-default-theme-lead-by-the-community/1545/960?u=jaggers - I believe this is @madsrh's preferred approach).

- Remove the separating line and make header bar/menu titles the same colour, so they combine to form a single continuous area.  @clobrano has said that this is possible, and some thread contributors prefer this approach.

clobrano:
- removed bottom border from headerbar/titlebar
- use headerbar bg color for menubar
- added side borders to menubar

closes #391

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1510,7 +1510,7 @@ headerbar {
   min-height: 40px;
   border-color: transparentize(_border_color($headerbar_bg_color), 0.7);
   border-style: solid;
-  border-width: 1px;
+  border-width: 1px 1px 0 1px;
   border-radius: 0;
   background-color: transparentize($headerbar_bg_color, $alpha-value);
   color: $headerbar_fg_color;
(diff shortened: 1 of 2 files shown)


Ads20000:
Firefox seems to be working properly, `hamster-applet` (which doesn't seem to be in 18.04...but I got it when on 17.04 I think and [you can get it from 17.10](https://packages.ubuntu.com/artful/hamster-applet)) doesn't strangely. I guess it's unsupported but idk why this is happening.

screenshot from 2018-05-04 12-32-07

clobrano:
@Ads20000 thank you for reporting this. Any chance you could run gtkinspector and see if that's a menubar or another kind of widget?

clobrano:
It looks like all GTK2 apps

Feichtmeier:
Inkscape is gtk2 and keepassx is Qt so this is the wrong PR for this =D
We are not back for the gtk2 theming since there were so many gtk3 design changes lately
And we didn't start to include the work of the qt communitheme project yet

clobrano:
Yep, but at least I can change menubar color in gtk2 :)

Feichtmeier:
GTK3 looks fine:
screenshot from 2018-05-04 14-35-56
(((bah how can you guys survice without no-title-bar-extension)))

PR #394, opened 2018-05-02
Save dialog enhancement

### LINKED ISSUE #357: Save dialog enhancements
I don't have the answer for this one, but whenever I see the save dialog, I always think it looks a bit off 🤔 

- The textfield where you have to type the filename is almost impossible to spot if it isn't selected. Maybe the background is just inherited from the darkmode. Perhaps this should be white as it is in Ambiance?

image

- The "browser-header-stack" panel looks weird being white when there's a table header with a lightgray background below - looks like something went wrong. Here I tried to make it darker. Is it too many shades of gray? Perhaps just use the sidebar color?

save

clobrano:
closes #357

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1578,6 +1578,10 @@ headerbar {
       }
     }
 
+    .text-button:not(.suggested-action):not(:hover) {
+      border-color: $hb_button_bg_hover;
+    }
+
     button, button.flat {
       @each $state, $t in ("", "normal"),
                           (":hover", "hover"),
@@ -1732,6 +1736,10 @@ headerbar {
           }
         }
 
+      &.text-button:not(:hover) {
+        border-color: $success_color;
+      }
+
       &:active, &:checked {
         $c: darken($success_color, 10%);
         background-color: $c;
@@ -1769,6 +1777,8 @@ headerbar {
         }
       }
     }
+
+
   }
 
   // squared corners when the window is maximized, tiled, or fullscreen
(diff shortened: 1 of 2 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr394`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr394` channel to test your snaps with those changes.

madsrh:
Nice work @clobrano, but something still isn't right for me and it **isn't** because of your work. We can merge this one, BUT...

I already mentioned that the _Cancel_ button and _Name_ label is hard to tell apart.
image

Perhaps something like @Feichtmeier did with the _Clear All_ button in the calendar would be good if it was possible 😢 
image

Anyway, wouldn't you agree that the use of gray here is odd. We have: hover, entryfield and pushed button and they look close to identical 🤔 
image

clobrano:
> I already mentioned that the Cancel button and Name label is hard to tell apart.

Yep, actually here Cancel and Name are a bit of a problem, however (I don't remember where I wrote the analisys) it is a bit of a mess, because I haven't been able to find a specific widget class to style uniformly all the buttons that require that. So the effect would be to see only some buttons with text highlighted/bordered and some not.

madsrh:
> I haven't been able to find a specific widget class to style uniformly all the buttons that require that. So the effect would be to see only some buttons with text highlighted/bordered and some not.

That would be highly inconsistent (_that word is really worn out_), but I would **almost** say it would be better 🙊

clobrano:
> That would be highly inconsistent (that word is really worn out), but I would almost say it would be better 

:D I know!
Ok, let me try one more iteration and see what happens

clobrano:
I should have found a way to don't mess up with other buttons :smile: 
What do you think?

button-border-hover06

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr394`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr394` channel to test your snaps with those changes.

Issue #397, opened 2018-05-03
Hover effect for sliders, switches and spin boxes


ya-d:
Those elements currently receive a colored border on hover while any other is fully layered by a solid color. Switches have none at all. I suggest:
All of these (sliders, switches and spin boxes) should have the same grey layer effect which check/radio buttons have (or had but currently lack in edge-channel, see #384). Sliders could have colored borders while pressed only.

_Maybe @clobrano can already consider this while working on #44._

<img width="284" alt="hover check-radio" src="https://user-images.githubusercontent.com/30781876/39566244-432fb58e-4ebb-11e8-9049-fe61e38aa325.png"> <img width="107" alt="hover slider" src="https://user-images.githubusercontent.com/30781876/39566245-434d39ba-4ebb-11e8-8fe0-fdf1259f42db.png"> <img width="127" alt="hover spin button" src="https://user-images.githubusercontent.com/30781876/39566247-436adb5a-4ebb-11e8-8749-ad6038f31360.png"> <img width="102" alt="hover switch" src="https://user-images.githubusercontent.com/30781876/39566249-438a1cb8-4ebb-11e8-854d-a98e0bedff1a.png">

clobrano:
New switch style will have hover effect, the same as check/radio (which will receive a fix as well for unchecked widgets).
Spinbuttons border highlight on hover is the remaining of the "old" button style discussion, so I think we can revert it back as well to a background highlight.

PR #384, opened 2018-04-29
Added dedicated mixin function for check and radio buttons

clobrano:
In order to avoid interferences between changes on buttons and check/radio, this commit adds a dedicated mixin function for check/radio button that fixes #367
 
I decided to add only the "button" types we actually use in scss files: normal, hover, active, insensitive, backdrop, backdrop-insensitive, so no osd- or alt-. I assume this is safe, but not 100% sure.

Below the result
check-radio-rework

PR #399, opened 2018-05-03
Fixed missing hover effect to normal check/radio

clobrano:
Fixes a regression introduced with #384.

see also #397


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr399`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr399` channel to test your snaps with those changes.

PR #404, opened 2018-05-06
Made window top border bright

clobrano:
In un-maximized mode only, added a inkstone colored window top border
to stress the idea of the light that comes from the top of the
screen.

Before
image

After
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr404`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr404` channel to test your snaps with those changes.

madsrh:
WOW! What a difference when you have overlapping windows! 👍

clobrano:
Ok, so I'd say to try this without changing backdrop headerbar color for now

Issue #109, opened 2018-01-26
More space between headerbar and canvas in Nautilus


madsrh:
![screenshot-from-2018-01-18-16-23-12](https://user-images.githubusercontent.com/3986894/35419338-b54eed54-0237-11e8-9378-d47d648648de.gif)

madsrh:
The padding also moves the scrollbar. It might be the wrong container the padding is added to.

clobrano:
> The padding also moves the scrollbar. It might be the wrong container the padding is added to.

Confirmed, I am going to fix this in #405

madsrh:
@clobrano This isn't fix with #405. This (109) issue is still present, but if we can't have both, we can discard this padding issue.

clobrano:
Uhm, weird it should, I actually kept a little margin between scrollbar and top view, maybe you're seeing this one?

madsrh:
Is that the 2px? I should still look like [this ](https://user-images.githubusercontent.com/3986894/35419338-b54eed54-0237-11e8-9378-d47d648648de.gif)- with and without scrollbars. And if the scrollbars are there they should start at the top, right? 🤔

clobrano:
:man_facepalming: damn you're right

clobrano:
Unfortunately, scrolledwindows is inside the container we want to move down, so scrollbar will always move with it

clobrano:
Closing this now, until upstream issue goes its way

PR #405, opened 2018-05-06
Made overlay-indicator (scrollbar) transparent and borderless

### LINKED ISSUE #395: Background color of scrollbar looks weird
I think background color of the scrollbar should be white? In this case it looks pretty weird.

clobrano:
closes #395

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -41,9 +41,16 @@ list.tweak-categories separator {
   }
 
   notebook > stack grid:not(.nautilus-list-view) {
+    padding-top: 2px;
+  }
+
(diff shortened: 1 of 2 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr405`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr405` channel to test your snaps with those changes.

Feichtmeier:
tried the pull :+1: solves the problem and uses good in general 
peek 2018-05-06 16-42

madsrh:
Looks good, but could the scrollable area have a semi transparent background?

image

> solves the problem...

Yes, but it still gives the impression that you're not at the top, so we still want #109 also.

clobrano:
slider-semitransparent

PR #421, opened 2018-05-09
Made headerbar backdrop color brighter

clobrano:
Made headerbar backdrop color brighter to improve distinction from active headerbar


Feichtmeier:
Hey Carlo I love your new "light from above" effect upon the headerbar.
However that "light" is also being applied to that separator between two headerbars:
image

No biggie but if there would be an easy solution maybe you could tried it in this PR? If this is the wrong place -> -1 and I open an issue

clobrano:
The "light from above" affect border-top-color only

https://github.com/ubuntu/gtk-communitheme/blob/8855c8252cb61cf811b15f0724667a2f67eee4a1/Communitheme/gtk-3.0/_common.scss#L1515

I am quite sure that the problem is a side effect of 2f0c1217. In fact if I revert that, the bright line goes away. I need to investigate more

PR #425, opened 2018-05-10
Improved selection ring in treeview and row

### LINKED ISSUE #423: No indication that tree/list view has focus
Because Communitheme uses a dashed orange focus rectangle, and it uses orange for the selected/highlighted item (whether list/tree is focused or not), there is no visual indication that the view has focus.

I'm not sure if this is possible with Gtk themes, but it would be nice if the focused view had an orange border as per focused entry fields.

At the very least, the dashed focus rectangle for selected view items should be white - so that it contrasts with the selection colour.

clobrano:
Changed selection ring to white in focus treeview and row

When selected, treeview and row have at the same time orange background
and orange selection ring, so that the latter is not visible. Changing
the selection ring to white makes it visible again.

closes #423

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -410,6 +410,14 @@ entry {
   }
 }
 
+treeview, row {
+  &:focus {
+    outline-color: transparentize(white, 0.2);
+    outline-offset: -2px;
+    -gtk-outline-radius: $small_radius;
+ }
+}
+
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr425`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr425` channel to test your snaps with those changes.

PR #432, opened 2018-05-11
Reworked gnome-terminal tab style

### LINKED ISSUE #410: Terminal tabs and scrollbar colors with custom color scheme
I've set my terminal color scheme to "Tango dark", but the tabs and the scrollbar still stay the default Ubuntu purple, which clashes with this scheme and others.

I noticed that the default Ubuntu 18.04 color scheme actually uses white and gray tabs with an orange scrollbar, which seems to be a good match for most color schemes.

Is there a way that I could customize these colors as an end-user or that they could be changed to something that matches better with the default color schemes available in the terminal?

This screenshot illustrates my problem with Communitheme and the "Tango dark" scheme:
communitheme

Here is the default theme's terminal:
default-ubuntu1

Here is the default theme's terminal with the "Tango dark" color scheme:
default-ubuntu-2

clobrano:
Restyled gnome-terminal tab using headerbar's color palette to get along
with different terminal color schemes.

closes #410

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -79,56 +79,63 @@ terminal-window {
     }
   }
 
+  notebook > header.top > tabs > tab {
+    &:checked {
+      color: $terminal_fg_color;
+      border-bottom: 2px solid darken($orange, 10%);
+    }
+  }
+
   notebook {
     // inherits from notebook header
     > header {
       border-color: $terminal_borders_color;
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr432`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr432` channel to test your snaps with those changes.

PR #442, opened 2018-05-14
Make icons less bright in backdrop

madsrh:
I don't know if the wildcard is considered a hacky solution.
Make icons in Nautilus less bright in backdrop.

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -13,7 +13,7 @@ list.tweak-categories separator {
 .nautilus-window {
   // keep details box visible
   // details box looks ugly when many items are selected and it floats above the orange
-  *scrolledwindow:backdrop { opacity: 0.85; }
+  *scrolledwindow:backdrop { opacity: 0.9; }
    // Makes icons less bright in backdrop
   paned box.floating-bar {
     background: $bg_color;
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -13,7 +13,7 @@ list.tweak-categories separator {
 .nautilus-window {
   // keep details box visible
   // details box looks ugly when many items are selected and it floats above the orange
-  *scrolledwindow:backdrop { opacity: 0.95; }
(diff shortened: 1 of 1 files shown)


clobrano:
Hi @madsrh, I'll test ASAP, in the meantime, are you sure you can't limit the change to nautilus' scrolledwindos?

madsrh:
No rush (although before october would be good).

> are you sure you can't limit the change to nautilus' scrolledwindos?

No, I'm not sure about anything 🤷‍♂️ I would prefer the icons to become darker, but that's not possible so this could work.

Do you mean something like:
`*.nautilus-window scrolledwindow:backdrop { opacity: 95; } `

Feichtmeier:
Imho if we tone down the opacity we should either do this slight opactiy reduction on every gtk window or on none. Otherwise it could look very confusing. This way it would enhance the effect of having two states of windows: foreground and background (active and backdrop). Otherwise there could be situations in which the user might think there are three states: active, backdrop and even more backdrop :sweat_smile:

madsrh:
> Also we can adjust opacity, since at 0.95 is hard to notice the difference, but the effect is interesting

I know - it is intentionally **very** subtle 🔎
You shouldn't really be able to tell, it just makes the icons fall slightly more in the background. I'll do a few tests with different opacities

madsrh:
This might be the worst recording ever :laughing: 

This is `opacity 0.92`
I know that we're splitting hairs here, but 0.9 seems too white to me. I like the "current" 0.94, but as you mentioned, it may be too subtle.

peek 2018-05-14 15-32

clobrano:
> I know that we're splitting hairs here

I understand you, when I tried I tested changes of 0.05 steps :smile: 
I think we can be a bit more brave :smiley:, it's a matter of point of view (and display quality) but I still think the difference should be visible.

madsrh:
> I still think the difference should be ~~visible~~ evident. 

Damn, I was really hoping you would agree to 0.92 😄 IMO it's visible, but no definitely not evident.

Do you a suggestion @clobrano?

clobrano:
Something in between 0.85 and 0.9 is too much for you? :)

PR #452, opened 2018-05-16
Bright border color in maximized window mode

clobrano:
Changed border color of headerbar, titlebar and menubar (only GTK3) when
the window is in maximized, tiled, fullscreen state.

Known limitation:
- GTK2 menubar is not able to change when a window is
in maximized state.

Dependency:
- dock top border should change to the same color if this change will
  be accepted. However the dock top border color would always be the
  same color, that is, it won't change according to window mode.

TODO:
- [x] fix [issue 752](https://github.com/micheleg/dash-to-dock/issues/752) on dash-to-dock
- [ ] wait for fix to be released officially (fix released in [ubuntu-dock-63ubuntu1](https://github.com/micheleg/dash-to-dock/releases/tag/ubuntu-dock-63ubuntu1) that will land in Cosmic
- [ ] implement and merge twin change on gnome-shell-communitheme

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1793,7 +1793,8 @@ headerbar {
   .tiled-left &,
   .maximized &,
   .fullscreen & {
-    border-color: transparentize(_border_color($headerbar_bg_color), 0.7);
+    border-color: $inkstone; //transparentize(_border_color($headerbar_bg_color), 0.7);
+
     &.selection-mode { border: none; }
 
     &:backdrop, & {
@@ -2161,6 +2162,19 @@ treeview.view {
 /*********
  * Menus *
  *********/
+menubar {
+  tiled &,
+  .tiled-top &,
+  .tiled-right &,
+  .tiled-bottom &,
+  .tiled-left &,
+  .maximized &,
+  .fullscreen & {
+    border-left-color: $inkstone;
+    border-right-color: $inkstone;
+  }
+}
+
 menubar,
 .menubar {
   -GtkWidget-window-dragging: true;


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr452`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr452` channel to test your snaps with those changes.

clobrano:
Work in progress, please do not merge it yet

clobrano:
To obtain the right effect changes are required on gnome-shell side as well. Unfortunately at the moment the solution is not totally clear, so I'll keep this in pause.

godlyranchdressing:
@clobrano how about this in the Shell's CSS?

#panel.solid {
  box-shadow: inset 0 -1px $inkstone;
}

transparency-test

The only problem there is that it's possible to activate the separator if a window is only a couple pixels away from the panel.

clobrano:
I am not totally convinced to extend the line also between dock and panel. There is a box-shadow between them, it's darker and with new jet color is to make more visible, but yours look too bright.

What I want to achieve is the following
image

but now I cannot keep the line between dock and panel and, more important to me, I cannot remove the line when the dock is transparent.
image

I am now in contact with the developer of dashtodock, hope I can find a solution with his help

clobrano:
Happy to say that dash-to-dock maintainer developed a test branch where I can apply the vertical border only on the opaque state. Now, it might take some time before having it upstream and then available in our version, but I am positive :tada: 

Thanks a lot to @micheleg for the awesome support :+1:

Feichtmeier:
@clobrano wow that's nice!

clobrano:
> @clobrano wow that's nice!

It is! Also, I broke my ubuntu-dash-to-dock, which is less nice :(

Feichtmeier:
@clobrano http://releases.ubuntu.com/18.04/
+
https://etcher.io/
+
Italian red ;D

clobrano:
This will be reopened when the dependency will be available

PR #454, opened 2018-05-16
Fix check + radio backdrop

godlyranchdressing:
Fixes some quirks like green checks using the gray $insensitive_borders_color, also lessens the box-shadow for regular buttons.

Before:
checks1
After:
check2


clobrano:
LGTM :+1: 1:

PR #455, opened 2018-05-16
Style nemo

### LINKED ISSUE #445: Nemo toolbar theming
Here's nemo with Ambiance:

nemo-ambiance

And here with communitheme:

nemo-communitheme

Although nemo doesn't use CSD, it has a toolbar pretty similar to the one from Nautilus, which is dark in Ambiance and with visible borders and separation between the toolbuttons. The buttons in the bar at the bottom are also probably a bit too large (I would't mind a dark bottom bar, while we're talking about it).

godlyranchdressing:
Closes #445

nemo

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -265,3 +265,59 @@ headerbar button image ~ window decoration ~ menu separator {
   // no unique class names or id used here so this rule is as specific as it can be
   background: $borders_color;
 }
+
+/***********
+ * Nemo *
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@godlyranchdressing very nice work. We could ofc merge this but I hope in our all long time interest, that this linux mint/cinamon theming will be a) limited to this app and b) won't need any more work as soon as they change this file manager? :)
I'd say let's merge this but make it clear that this is an exception and this theme will be designer for GNOME 3.28+ don't you agree?

PR #463, opened 2018-05-20
Separator finetune + close button in backdrop fix

### LINKED ISSUE #453: Treeview header button has no hover effect, sidebar separator is too light
The treeview header button has no hover and active effect. The sidebar separator is too light and the treeview button blends into the sidebar too much.
sidebartv

godlyranchdressing:
Closes #453

Also makes the close button in backdrop have the same color as the other titlebuttons instead of white.

Before on the left, after on the right:
trans2

Also adjusts headerbar separator so it's more visible. Before on the left, after on the right:
b4af


Feichtmeier:
Tried the PR - looks good!

PR #468, opened 2018-05-22
Fixed switch margins in headerbar

### LINKED ISSUE #462: Switch expands header bar
test

screenshot from 2018-05-18 21-45-49
(it also make buttons look weird)

clobrano:
switch margin top and bottom where defined in two different places and
in a way that made the headerbar with a switch bigger than the ones
without.

Fixed defining the correct top and bottom margins in only one place

closes #462

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1547,7 +1547,7 @@ headerbar {
 
   ~ separator, separator {
     &, &.titlebutton {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr468`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr468` channel to test your snaps with those changes.

PR #469, opened 2018-05-22
Added pressed effect on checked non-flat buttons

### LINKED ISSUE #457: Pressed toggle button has no hover effect
A toggle button that is toggled/active has no hover effect. This can be seen in `gtk3-widget-factory`. Press one of the toggle buttons (so that it looks sunken, and has a dark background) - if you move the mouse away from it then back over, the button no-longer gets lighter when the mouse passes over.

clobrano:
closes #457

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -568,6 +568,8 @@ button {
       transition-duration: 100ms;
     }
 
+    &:hover:checked { @include button(active-hover); }
+
     &:backdrop {
       &.flat, & {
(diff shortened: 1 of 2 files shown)


godlyranchdressing:
Can you post a GIF of it in action?

clobrano:
Sure

button-hover-checked

Feichtmeier:
peek 2018-05-22 22-33
The hover is very bright in the dark variant

clobrano:
> But maybe this fits this PR? If not just ignore

I don't want to ignore :D but I prefer opening a new issue for that ;)

PR #470, opened 2018-05-22
Restore treeview background color, add transition to switch

### LINKED ISSUE #467: Treeview.view.header background-color
adwaita_treeview_header
ambiance_treeview_header
canta_treeview_header
materia_treeview_header

And this is our treeview header:

communithem_treeview_header

____

The way it is colored in our theme makes it look like the header does not belong to the table. Every other theme I found colors the header of the table in the same color as the table. Especially in the dark theme:
image
(I actually used a  different color for the dark theme here but I think some other  PR has overwritten this)
I think it would be better if we color the header in the same color as the rest of the table in both dark and light themes and dark backdrop + light backdrop.

godlyranchdressing:
Closes #467

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2146,6 +2146,7 @@ treeview.view {
   background-image: none;
   border-style: none solid solid none;
   &:not(:hover), &:not(:active) { background-image: image($bg_color); }
+  &:backdrop { background-image: image($backdrop_bg_color); }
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@godlyranchdressing looks very good. Only one minor issue: I believe you have "forgotten" the backdrop color of the light version:
image

godlyranchdressing:
Of course! I just forgot about the backdrop state there.

Feichtmeier:
Works! :heart_eyes: 
image
image

PR #480, opened 2018-05-25
Entry top shadow

godlyranchdressing:
Adds a very slight shadow based on the button shadow to entries

entry
entry1


madsrh:
Funny! I was just looking at this. Yes +1

PR #481, opened 2018-05-25
Use plain buttons for infobar

godlyranchdressing:
Seen [here](https://github.com/ubuntu/gtk-communitheme/pull/389#issuecomment-391880002)


Feichtmeier:
image
image

I think they look rather no so super perfect on yellow, but a it's a nice solution overall and look very good on purple.

Feichtmeier:
@madsrh please double check above and merge/ request changes

PR #487, opened 2018-05-27
Fixed menu issues

vinceliuice:
Fixed menu issues and removed gedit search entry occurrences tag ugly boder

Edit from @feichtmeier : 
closes https://github.com/ubuntu/gtk-communitheme/issues/464

### DIFF
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -46,11 +46,17 @@ list.tweak-categories separator {
     margin-right: 9px;
   }
 
-  menu menuitem:first-child:hover {
+menu, .menu,
+.context-menu {
+  menuitem {
     // context menu in nautilus has its last-child labeled as its first child lol
     // using :hover for specificity bump
-    border-radius: 0 0 $small_radius $small_radius;
+    &:hover {
+      &:first-child { border-radius: 0 0 $small_radius $small_radius; }
+      &:last-child { border-radius: $small_radius $small_radius 0 0; }
+    }
   }
+}
 
   paned > separator{
     // separator between sidebar and main window view
--- a/Communitheme/gtk-3.0/_apps.scss
+++ b/Communitheme/gtk-3.0/_apps.scss
@@ -60,7 +60,6 @@ list.tweak-categories separator {
   }
 }
 
-
 .nautilus-desktop-window {
   .nautilus-desktop.view {
     // use white text for desktop icons
@@ -263,6 +262,13 @@ terminal-window {
     background-color: $backdrop_bg_color;
     border-color: $backdrop_borders_color;
   }
+}
+
+.gedit-search-entry-occurrences-tag {
+  color: transparentize($fg_color, 0.6);
+  border: none;
+  margin: 2px;
+  padding: 2px;
 }
 
 /***********
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2188,8 +2188,14 @@ menubar,
     min-height: 16px;
     padding: 4px 8px;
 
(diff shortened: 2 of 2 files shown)


Feichtmeier:
Yay @vinceliuice \o/ Thanks for this PR!

Sadly, the border radius of the last child of csd context menus is still present in eye of gnome:
br

Paz-it:
To avoid it we need  to leave a small gap at the top and the bottom of the menu  so the test will start where the round corner finished/where the round corner starts.

Feichtmeier:
@Paz-it yes I've did thisin a closed PR but @godlyranchdressing and @madsrh are not fans of this solution.  It would make everything a lot easier and less sensible  for errors though.  :shrug

Paz-it:
When the menu is less condensed It looks much better. I saw that you are trying different padding, over  time following the --edge channel . I think padding at top/bottom/left/tight will solve a lot of issues.
Just for example of a menu I did for openbox  to illustrate my idea : 
back_to_openbox_by_paz_1-dbkv2ev

madsrh:
@Feichtmeier I'm fine with the padding solution, I just said I would rather keep (fix) the current - but that's easy for me to say. @godlyranchdressing?

Don't let my vote tip the scale :)

clobrano:
I think the biggest problem is how different applications build the menus,. In some cases the last-child is on top and in some others is at the bottom, so to define a general rule for the rounded corners on hover it's complicated

Feichtmeier:
@nadsrh yes that would be the most beautiful  solution and very extraordinary and a real usp :D (it's free but you know what I mean)
So we could check every gnome app for how it fills the lists but that still does not fix the borderbleed :shrug:

Paz-it:
There are two factors to take in consideration in the menus/sub-menus. 
Padding and menu-overlap. in openbox code(in my example above) it looks like that : 
#Menu geometry
menu.border.width: 10
menu.overlap.x: -11
menu.overlap.y: 0
_____________________
I'm sure how to translate it to our .css file, but you guys seems to know.  
So, the main idea is to play with those 2 factors to get the result we want and prevent the borderbleed.

Feichtmeier:
@Paz-it I think your idea is basically what was done with the popovers, right?
peek 2018-05-30 00-09

The menus could end up quiet "fat" then and we would have a lot of space taken :thinking:

Paz-it:
Yes, exactly! @Feichtmeier
 It's better a bit  "fat" then cluttered,  condensed and border-bleeding . I would clean up the mess with the keyboard shortcuts as well BUT maybe it's only me :dancer:

Paz-it:
That's a work in progress of a gtk theme I created looong ago in 2014 :smile: 
You can see how clean the menu is without those keyboard shortcuts : 
coffee_w_cream_wip_by_paz_1-d7890n5

clobrano:
@vinceliuice are you willing to provide some feedback on this PR? Otherwise I think @Feichtmeier can continue with his work on the same topic

vinceliuice:
Sorry guys, Recently my job was busy. Now I know how to fix this problem

Feichtmeier:
Hi @vinceliuice 
I've tried your last commit, sadly it still does not work as expected and I believe there also joined yet another issue with nautilus (the third child now is wrong, too):
horror2
horror

Feichtmeier:
Hi @vinceliuice !
Thanks again for this PR!
Our designer decided to go with a different solution for now which you can see here:
https://github.com/ubuntu/gtk-communitheme/pull/509

Thank you very much, especially for your FireFox fix. If you have any spare time please check our issues and see if you have any idea for some of them to fix, we got still  a lot of stuff to do! :smile:

vinceliuice:
Add padding to menus and keep the border radius is excellent！ I love this design too.

PR #488, opened 2018-05-27
placessidebar_transparentize_selected_row

Feichtmeier:
Before:
before
After:
after

I think this is needed because the pure orange distracts too much. @madsrh what do you think?

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4088,7 +4088,7 @@ placessidebar {
     &:backdrop {
       color: $backdrop_text_color;
 
-      // &:selected { color: $backdrop_selected_fg_color; }
+       &:selected { background-color: transparentize($selected_bg_color,0.2); }
 
       &:disabled { color: $backdrop_insensitive_color; }
     }
@@ -4261,7 +4261,7 @@ infobar {
       selection { background-color: darken($c, 10%); }
 
       &:backdrop, & {
-        color: $selected_fg_color; 
+        color: $selected_fg_color;
         border-color: _border_color($c);
       }
 
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4088,7 +4088,7 @@ placessidebar {
     &:backdrop {
       color: $backdrop_text_color;
 
-       &:selected { background-color: transparentize($selected_bg_color,0.2); }
+       &:selected { background-color: transparentize($selected_bg_color,0.1); }
 
       &:disabled { color: $backdrop_insensitive_color; }
     }


madsrh:
@Feichtmeier Are you considering giving it a salmon look? 🤣 
Sorry, jokes aside. I know that it might be weird to do two different things, but how about making it darker and not using transparency? More precisely #dd501e (or slightly darker than that if you prefer) to follow the "original" overlay/multiply idea.

http://colorblendy.com/#!/multiply/

Feichtmeier:
image
I can hardly spot any difference and it still draws too much attention imho

madsrh:
image

>  ...More precisely #dd501e (**or slightly darker than that if you prefer)** to follow the "original" overlay/multiply idea.

Yeah, I know 😃 If you see what I see above, it's not as "shiney". Anyway, maybe we need to adjust the guideline we multiply with 🤷‍♂️ 

**#cd4b1d** ?

image

Feichtmeier:
:D Well, darkening does not make it less attention grabbing in my opinion :woman_shrugging: 
Everything loses it's "color" in the backdrop, why not that orange?
How about using opacity 0.9 again:
image

Feichtmeier:
No matter which color we chose, we would need to change settings, tweaks and logs too:
image

madsrh:
> :D Well, darkening does not make it less attention grabbing in my opinion 🤷‍♀️

I'm referring to [this ](https://github.com/ubuntu/gtk-communitheme/issues/212#issuecomment-370606475)thread again and when I look at something like [this](https://user-images.githubusercontent.com/3986894/37005828-d10ba638-20d5-11e8-9ad6-8919b7a16a9a.jpg), it's very obvious to me which window is in focus. You call it _less attention grabbing_ and I would say that the window is further away 🤷‍♂️ 

Not my best illustration, but hopefully you understand what I'm aiming for 🏹 
backdrop


> Everything loses its "color" in the backdrop, why not that orange?

It does? It shouldn't 😃 

> How about using opacity 0.9 again:

+1 I'm fine with that!

Paz-it:
+1 for transparentize it if that still mater to someone here :dancer:

godlyranchdressing:
Wouldn't the place to make this change be something like `list row:selected:backdrop`? Or is that not working for the places sidebar?

I think the _backdrop_color function could be a little more intense instead of transparentizing the orange in this one case.

madsrh:
> I think the _backdrop_color function could be a little more intense instead of transparentizing the orange in this one case.

What do you have in mind @godlyranchdressing ?

The icons transparentizing already a compromise, so it kind of breaks the "rule" of things being darker in backdrop 🙄

Feichtmeier:
I think he means that if we change it everywhere this would be the place.

madsrh:
> Was this considered an option? Could be an improvement especially if you have situations like the one posted above.

@ya-d This looks confusing because they all have the same orange, but all suggestions (opacity, darken and gray) would fix that IMO.

Feichtmeier:
Since we should discuss this first I close this pr now and finalize a decision  with though guys in that old backdrop issue :)

PR #499, opened 2018-05-29
Fixed spinbutton hover color in dark variant

### LINKED ISSUE #498: Spinbutton dark variant hover effect is too bright
previously reported [here](https://github.com/ubuntu/gtk-communitheme/pull/469#issuecomment-391134919)

image

clobrano:
closes #498

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1220,7 +1220,7 @@ spinbutton {
       &:dir(rtl) { border-style: none solid none none; }
 
       &:hover {
-        background-color: darken(white, 5%);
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Works but something happend to the spin btns there is some strange shadow in them:
yayacdscsd
What happened to the dark theme? :O

clobrano:
Do you mean the box-shadow on top? It is there in the light theme as well, it doesn't look bad to me

Feichtmeier:
Yes, previously they looked like they were a half or a quarter level above the surface. Now with the shadow it looks like they are a quarter pressed into the surface

clobrano:
Actually, since nothing actually changed and since this PR is for hover effect, I'd better merge this and open a specific issue for the other problem

PR #514, opened 2018-06-04
Added hover transition to non-flat buttons

clobrano:
- improved hover button transition
- added hover transition to non-flat button

see https://github.com/ubuntu/gtk-communitheme/issues/184#issuecomment-392477771


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr514`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr514` channel to test your snaps with those changes.

PR #516, opened 2018-06-05
Use flat buttons in toolbar only if explicitly defined by developer

### LINKED ISSUE #493: Flat combobox in evolution
Evolution's new message window contains 2 combo-boxes to choose message type, and bullet list type. With most themes these are not flat, with Communitheme they are. However, the flat look does not look right - as the drop down arrow can look separate from the text.

Communitheme:
communitheme

Ambiance:
ambiance

clobrano:
closes #493

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1439,7 +1439,7 @@ toolbar {
 
   padding: 4px 3px 3px 4px;
 
-  button {
+  button.flat {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr516`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr516` channel to test your snaps with those changes.

PR #518, opened 2018-06-05
bring_back_blue_spinner

Feichtmeier:
That's a regression, gtk process working spinner was orange for some commits. 
Fixed it back to blue:
erfsfgsdfgd


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr518`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr518` channel to test your snaps with those changes.

PR #525, opened 2018-06-07
Differentiate between checked and unchecked disabled switch

### LINKED ISSUE #238: inactive window switch does not clearly show its on/off state
Hi,

When I try to add a search engine : 
image
the "Default" button should be green.

clobrano:
When in insensitive/disabled state there is no difference between
checked and unchecked switches.

closes #238 

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr525`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr525` channel to test your snaps with those changes.

Feichtmeier:
Master:
image

This PR:
image

PR #527, opened 2018-06-07
Use transparent inset shadows instead of dashed lines for undershoots

Feichtmeier:
Closes https://github.com/ubuntu/gtk-communitheme/issues/523

### DIFF
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -766,25 +766,26 @@
 
   //We don't like the dashed line so we go for a light shadow
   $us_shadow_color: $silk;
+  $us_shadow_alpha: 0.8;
   @if $variant=='dark' {
     $us_shadow_color: $jet;
   }
 
   @if ($p == left) {
-    box-shadow: inset 2px 0 0 transparentize($us_shadow_color,0.8),
-                inset 1px 0 0 transparentize(darken($us_shadow_color,20%),0.8);
+    box-shadow: inset 2px 0 0 transparentize($us_shadow_color,$us_shadow_alpha),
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr527`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr527` channel to test your snaps with those changes.

madsrh:
It's subtle, but very slick +1 from me.

@clobrano ?

If the scrollbars didn't autohide, we wouldn't need this in the first place 😉

clobrano:
Very nice indeed. Would it be possible to try a shadow a little wider and brighter/transparent? If not, I'm glad to accept this one

Feichtmeier:
More transparent is still possible.
Wider could be hard but I try.
It's actually a "cheat" because that undershoot seems to be technically a box. If you use the blur or spread from inset box shadow the box is getting the shadow and not only the edge. So I use two shadows for each edge, 1 at pixel 1 and 1 and pixel 2. I could add a third at pixel 3. I try and post gifs before commiting

Feichtmeier:
More transparency is not really possible :D or I changed the color :thinking: 
That's 0.8 (current):


That's 0.9 and the shadow disappears:

And that's with three shadows:
3shadows

clobrano:
I think the current one looks good enough :+1:

Feichtmeier:
Okay, we can tweak it later if we have more ideas. Maybe you have one when you look into the implementation which is a bit "weird" :smile:

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr527`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr527` channel to test your snaps with those changes.

PR #531, opened 2018-06-08
Menus again: Padding top bottom & no border radius for items

Feichtmeier:
werwerwer

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2228,7 +2228,7 @@ menu,
 .menu,
 .context-menu {
   margin: 4px; // see https://bugzilla.gnome.org/show_bug.cgi?id=591258
-  padding: 4px;
+  padding: 8px 0;
   background-color: $menu_color;
   color: $text_color;
   border: 1px solid $borders_color; // adds borders in a non composited env
@@ -2243,7 +2243,7 @@ menu,
     min-height: 16px;
     min-width: 36px;
     padding: 5px 6px;
-    border-radius: $small_radius;
+    border-radius: 0;
     // text-shadow: none;
 
     &:hover { background-color: $base_hover_color; }


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr531`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr531` channel to test your snaps with those changes.

Paz-it:
I have a very weak heart, I can't believe it ..No no no...Please!  :hammer:

clobrano:
@Paz-it no need to be so harsh.
@Feichtmeier I think I prefer current version, it is more distinctive respect other themes that uses it (except the one we borrow this from :D). The thing I like less is the effect with submenus.

Feichtmeier:
This gif shows the standard how every other menu that uses round corners handles the menu entries and sub menus. I still can't decide whether our current approach is good design because no one else uses it or bad design because it looks weird

clobrano:
Alright, so why don't we take some time to get use to it?

clobrano:
Honestly, I don't see much difference in "selection visibility". I am not against this PR, but I'd like to keep the uniqueness in our menus highlight.

Feichtmeier:
Being unique is a great thing, if the design makes sense and does not scare more people than it attracts people. The current master looks like bubbles in a bubble. Where this approach looks like stripes selected from a bubble. :man_shrugging: If I am really alone with this I give up. I just can't think that this would be welcomed by "neutral" people who haven't gone through all that menu changes we have went through. I believe we could have lost our neutral view on that when merging the last change I did which is in master now. I would rather go with the good looking (osx!) standard than a questionable uniqueness

clobrano:
I didn't even participate in this change, so I think I am quite neutral :D

I like them both, if you are so strongly convinced, we can merge it

madsrh:
@Feichtmeier I think it's nice that they are similar to the headerbar's menu, but this is not an achilles heel for me - fine with both solutions 🤷‍♂️

Feichtmeier:
Na I won't force my opinion on you. Let's see what others say if they are still around:
@luxamman @godlyranchdressing

Paz-it:
If my opinion can help to keep the current, I'll be more then happy. It's unique in a good way, clean and stylish. solving the issues of bleeding borders and so much better looking from the other themes in the examples above.

Luxamman:
Well, after reviewing I would prefer the new way because this is also in line with the shell and other selected elements in lists. But should also be applied to the app menu list if changed.

Feichtmeier:
@Luxamman you are talking about the gtk popups, aren't you? This one? : 
yaop

Luxamman:
Yes @feichtmeier 
This also is not completely fitting with the actual rounden corner style it seems

Feichtmeier:
Okay. I try to adapt them to the menu

Luxamman:
Adapt to the actual or the proposed style?

Feichtmeier:
To the menu of this PR. So we can test how it would look

Feichtmeier:
image
I can remove the borde-radius for the selected modelbuttons but I can't remove the space
:| Seems like we'd need to stick with that either way. But they are something a little bit special because they can have buttons in them and other stuff

Feichtmeier:
@Luxamman this does not seem to be possible because the popovers have more tha "just" menuitem labels. Those would need to stay as they are. Is this ok for you?

Luxamman:
Can't test it at the moment, but if the others are okay with it we can test it in the field. If it is not working we still can reverse it - hopefully :)

Feichtmeier:
Ofc we can revert it. That's why we use git :dancer:

godlyranchdressing:
I much prefer this!

Paz-it:
I'll love it much more if it will change back to have the left/right padding. (If it stays like this it will be the first thing I'll change in the final theme locally.) 
Let us be unique in the good way!

PR #532, opened 2018-06-08
sidebar colors tweaks

clobrano:
Initially discussed in #407 and revived by [this comment](https://github.com/ubuntu/gtk-communitheme/pull/407#issuecomment-395594228)

But basically is because current mainview and sidebar and button (yes I know) look a bit dull.

---

**EDIT** still to be done: 
- [x] filechooser
- [x] treeview header button hover 
- [x] bug when right clicking a row in treeview 
- [x] fix sidebar backdrop

Here are a new attempt. Please make a long test, several pieces can be affected (One for all, the filechooser, that is to be fixed already).

image
image
image
image

Filechooser fixed
image

PR #538, opened 2018-06-12
Button hover slighly darker

clobrano:
This comes from #532, where buttons became so bright that no effect was visible on hover.

This PR add a slight darker color on hover. Note that it affects also suggested action buttons

button-hover-darker

### DIFF
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -235,8 +235,10 @@
   @else if $t==hover {
     // hovered button
     -gtk-icon-effect: highlight;
+
     @if $flat == true { $c: darken($c, 10%); }
-    $_bg: if(lightness($c)<35%, lighten($c, 25%), lighten($c, 5%));
+
+    $_bg: if(lightness($c)<35%, lighten($c, 25%), darken($c, 1%));
     background-color: $_bg;
     color: $tc;
 


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr538`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr538` channel to test your snaps with those changes.

madsrh:
Thanks for giving it a try - that really isn't working out!

Feichtmeier:
Not at all or the level of darkening?

Feichtmeier:
hover


We can't lighten up the bg in white variant since it's already white. But we can lighten up the textcolor.

```    $_bg: null;
    $_tc: null;

    @if $c==$headerbar_bg_color { $_bg: lighten($c, 15%); }
    @else if $c==$success_color or $destructive_color or $neutral_color or $purple { $_bg: lighten($c, 5%); }
    @else if $c==$button_bg_color {
      @if $variant=='light' { $_bg: lighten($c, 25%); $_tc: lighten($tc,20%);}
      @else { $_bg: lighten($c, 1%); }
    }

    background-color: $_bg;
    color: $_tc;```

clobrano:
I was thinking... maybe this could help #567, but I think the label should "transparentize" a bit less

madsrh:
Could you update this PR so we can give it a try? 😃

Issue #265, opened 2018-03-16
Buttons and entry fields too similar?


CDrummond:
With the recent change to the button look, buttons and entry fields look **very** similar.

In the following screenshot, its hard to see if the middle item is an entry field or a button (it's an entry field):

![entry_or_button](https://user-images.githubusercontent.com/1141694/37512563-697d8ac2-28fa-11e8-89c6-b1a9ed1c3f12.png)

The above was created using the following python code:

```
#!/usr/bin/python3

import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk

class ExampleWindow(Gtk.Window):

    def __init__(self):
        Gtk.Window.__init__(self, title="Example")
        
        box = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=10, margin=10)
        box.set_homogeneous(False)

        entrya = Gtk.Entry()
        entrya.set_text("  Is this an entry or a button?")

        entryb = Gtk.Entry()
        entryb.set_text("  Is this an entry or a button?")

        button = Gtk.Button(label="Is this an entry or a button?")
        box.pack_start(entrya, True, True, 0)
        box.pack_start(entryb, True, True, 0)
        box.pack_start(button, True, True, 0)

        self.add(box)

window = ExampleWindow()        
window.connect("delete-event", Gtk.main_quit)
window.show_all()
Gtk.main()
```

Perhaps buttons should have a _slightly_ darker background?

madsrh:
@CDrummond I agree, I think the buttons are still being worked on. [Issue 211](https://github.com/Ubuntu/gtk-communitheme/issues/211) is still open.

@clobrano Isn't there still some shadow missing for buttons? My mockup in #211 isn't the best place to look, but perhaps look to Pop for inspiration for the shadow:
![image](https://user-images.githubusercontent.com/3986894/37515739-13c50896-290c-11e8-9f6c-403a0d021375.png)

I'm not sure if the shadow becomes hard to notice when we also have a border? So perhaps more shadow than Pop uses.

clobrano:
I thought the shadow should appear only on hover. I'll fix it

madsrh:
> I thought the shadow should appear only on hover. I'll fix it

A little shadow on buttons and even more shadow if hovered ;)
Again, like Pop (although their hover shadow might be a bit too much).

clobrano:
We'd need to break the pixel in a half here :smile: , but @CDrummond's right.

[...]

PR #541, opened 2018-06-14
Added back box-shadow to button in normal state

clobrano:
peek 2018-06-14 15-00

see #265

### DIFF
--- a/Communitheme/gtk-3.0/_drawing.scss
+++ b/Communitheme/gtk-3.0/_drawing.scss
@@ -229,6 +229,7 @@
       box-shadow: none;
     } @else {
       border-color: $_border;
+      box-shadow: 0 1px transparentize(black, 0.9);
     }
   }
 
@@ -246,7 +247,7 @@
     } @else {
       border-color: $_border;
       border-bottom-color: darken($_border, if($c == $headerbar_bg_color or $c == $button_bg_color, 15%, 12%));
-      box-shadow: 0 1px transparentize(black, 0.9);
+      box-shadow: 0 1px transparentize(black, 0.85);
     }
   }
 


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr541`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr541` channel to test your snaps with those changes.

Feichtmeier:
I think I have already expressed that I prefer the current version but why don't we use other methods to
a) indicate that buttons are clickable (like a colour difference)
b) differentiate buttons from entries with a change to entries
?
This looks a bit like the buttons are detached from the surface. Like when you lay a piece of paper uppon the table

clobrano:
Plan A is the one we rejected considering that white buttons look better than gray ones
Let me try a little with plan B

Feichtmeier:
Yes white can't become any brighter. But the background could be darkened a bit. When buttons have no box Shadow and only a little color difference to the surface I understand that some might say that they can not be identified as an clickable object (note: I don't mean hover). 
In the software main view three white objects have a good contrast on the grey area

clobrano:
The problem is distinguish button from entries.

Feichtmeier:
Yes this could be achieved by a super soft border-top increase and or inset box shadow but you prbly know :dancer:

clobrano:
Here is a mockup applying to entries the same effect active buttons have

image

It looks a bit rough in the top right corner when in combo with a button and I am a little in favor of button change, but I have no strong preference

@madsrh what do you think about this?

madsrh:
Well, this is hard 🤔 

I think the entry fields looks good, but on the buttons I would prefer some kind of shadow. I like the buttons [in the PR above](https://user-images.githubusercontent.com/2883614/41413554-bcf734ac-6fe3-11e8-9124-0d546ab41dc5.gif).

Feichtmeier:
@clobrano I will not block this. Maybe everyone else is right. Usability > MostOtherStuff in the end. We got a big audience (ubuntu users) so better go for this!

PR #542, opened 2018-06-14
Improved stack-switcher button backdrop color

clobrano:
currently, stack-switcher button color in backdrop is darker than the
headerbar's one in the same state, while in normal state the color
relationship is inverted (hb darker then stack-switcher). This causes
a bad transition effect.

Using _backdrop_color function to automatically get a valid bg color
for this widget.

Before
peek 2018-06-14 15-24

After
peek 2018-06-14 15-19

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1676,7 +1676,7 @@ headerbar {
       }
 
       &:backdrop {
-        background-color: mix($c, $headerbar_bg_color, 50%);
+        background-color: _backdrop_color($c);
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr542`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr542` channel to test your snaps with those changes.

Feichtmeier:
@clobrano tested it. Looks good + reduces the transistion issue (it's still there but not on every button). :+1:

Issue #381, opened 2018-04-28
Popovermenu Boxpointer starts to high in some apps


Feichtmeier:
![boxpointer](https://user-images.githubusercontent.com/15329494/39400256-3ad3c06a-4b2d-11e8-86f4-3dfecd9bfb82.gif)

![image](https://user-images.githubusercontent.com/15329494/39400257-3c2d203c-4b2d-11e8-8a75-768d5a5c5152.png)

PR #543, opened 2018-06-14
Fix headerbar button height > width

clobrano:
On headerbar image buttons are taller than wide, except appmenu button.

Fixing this adjusting headerbar top and bottom margins.
Incidentally, this also fixes issue #381 

peek 2018-06-14 16-07

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1853,8 +1853,8 @@ headerbar {
   spinbutton,
   // separator,
   button {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr543`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr543` channel to test your snaps with those changes.

Feichtmeier:
Yeah! Nice fix!

PR #544, opened 2018-06-14
Added gray hover effect to circular slider

clobrano:
This change gives to circular sliders the same hover effect that check and radio buttons have

Note: for non circular sliders, we need to create the appropriate png

see #397

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -3130,7 +3130,9 @@ scale {
     transition: $button_transition;
     transition-property: background, border, box-shadow;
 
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr544`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr544` channel to test your snaps with those changes.

Feichtmeier:
@madsrh did you test this one?

madsrh:
@Feichtmeier yes, it works as intended, but we still need the non circular slider png or is that another PR?

PR #567, opened 2018-06-19
Treat horizontal spinbuttons as buttons, like vertical spinbuttons

Feichtmeier:
peek 2018-06-19 23-15

@madsrh @clobrano 

I've "noticed" that the vertical spinbuttons already worked like regular non flat buttons and was wondering why the horizontal don't. By adding the button functions they now work like regular buttons, too. Are you +1 with this "last-minute" change?

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1218,21 +1218,25 @@ spinbutton {
       border-color: $borders_edge;
       border-radius: 0;
       box-shadow: none;
+      @include button(normal);
 
       &:dir(rtl) { border-style: none solid none none; }
 
       &:hover {
         background-color: if($variant=="light", darken($bg_color, 5%), lighten($bg_color, 25%));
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr567`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr567` channel to test your snaps with those changes.

madsrh:
I'm +1 for them behaving alike, yes :D

I like this minimal effect on hover. Is it just me or is it a bit odd that the ckeckboxes/radios use the hover gray? or perhaps it just looks different because the background is now gray too :man_shrugging:

clobrano:
I would have expected an effect similar to this one

vert-spinbutton-hover

why is it so different?

Feichtmeier:
True, the highlight is way to soft. I'll need to try again!

Feichtmeier:
I will try later with a different PR

PR #582, opened 2018-06-29
Improved hover effect to GTK3 toolbar button

### LINKED ISSUE #571: Toolbar button hover very hard to see
The toolbar hover/mouse-over is very hard to see - it seems to be a very subtle lightening of the background. This can be see in non-headerbar apps such as Evolution, EasyTag, Synaptic, etc.

LibreOffice, however, has a much nicer hover effect - here the button frame is drawn on hover. (Although there is no effect for hovering over toggled buttons).

Perhaps all Gtk apps should follow the LibreOffice style? ie. draw frame on hover?

clobrano:
(Sorry, I forgot to PR this)

hover effect on GTK3 toolbar buttons is hard to see. For GTK2 toolbar buttons
have the same hover effect than normal buttons, which makes them easy and
pleasant to see.

Then, this commit applies the same hover effect to GTK3 toolbar button

closes #571

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1451,12 +1451,17 @@ toolbar {
   padding: 4px 3px 3px 4px;
 
   button.flat {
-    @each $state, $t in ("", "normal"), (":hover", "hover"), (":hover:backdrop", "backdrop-hover"),
-    (":active, &:checked", "active"), (":disabled", "insensitive"),
-    (":disabled:active, &:disabled:checked", "insensitive-active"), (":backdrop", "backdrop"),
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr582`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr582` channel to test your snaps with those changes.

Feichtmeier:
If the last gif on that issue is still up to date: what I dont understand is... the toolbar icons are flat buttons, yet when you hover they receive the same hover effect like normal buttons, is this wanted?

clobrano:
> the toolbar icons are flat buttons, yet when you hover they receive the same hover effect like normal buttons, is this wanted?

it is actually the change proposed, yes

PR #597, opened 2018-07-01
Added a dark border to toolbar button on hover

### LINKED ISSUE #571: Toolbar button hover very hard to see

clobrano:
New iteration of toolbar button hover. This change tries again to
mimic GTK2 behavior, this time adding a darker border and a brighter
background color

closes #571 

toolbar-hover-non-flat

- [ ] see #614 first

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -1451,19 +1451,20 @@ toolbar {
   padding: 4px 3px 3px 4px;
 
   button.flat {
-    @each $state, $t, $is_flat in ("", "normal", true),
-                                  (":hover", "hover", false),
-                                  (":hover:backdrop", "backdrop-hover", false),
-                                  (":active, &:checked", "active", true),
-                                  (":disabled", "insensitive", true),
-                                  (":disabled:active, &:disabled:checked", "insensitive-active", true),
-                                  (":backdrop", "backdrop", true),
-                                  (":backdrop:active, &:backdrop:checked", 'backdrop-active', true),
-                                  (":backdrop:disabled", 'backdrop-insensitive', true),
-                                  (":backdrop:disabled:active, &:backdrop:disabled:checked", 'backdrop-insensitive-active', true) {
-      &#{$state} { @include button($t, $c:$menu_color, $flat:$is_flat); }
+    @each $state, $t, $is_flat in ("", "normal"),
+                                  (":hover", "hover"),
+                                  (":hover:backdrop", "backdrop-hover"),
+                                  (":active, &:checked", "active"),
+                                  (":disabled", "insensitive"),
+                                  (":disabled:active, &:disabled:checked", "insensitive-active"),
+                                  (":backdrop", "backdrop"),
+                                  (":backdrop:active, &:backdrop:checked", 'backdrop-active'),
+                                  (":backdrop:disabled", 'backdrop-insensitive'),
+                                  (":backdrop:disabled:active, &:backdrop:disabled:checked", 'backdrop-insensitive-active') {
+      &#{$state} { @include button($t, $c:$menu_color, $flat:true); }
     }
 
+    &:hover { background: $menu_color; border-color: darken($menu_color, 30%); }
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr597`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr597` channel to test your snaps with those changes.

CDrummond:
Looks much better. A couple of issues:

1. There is no shadow given to the buttons - perhaps this is on purpose?
2. At first I thought there was no 'press' effect - but having looked **very** closely, a **very** slight inner shadow appears. This is exceedingly hard to spot. It makes it _feel_ as if the button has not been pressed.

As an aside, why not just theme these button as per regular no-flat buttons? ie. draw them flat when no mouse-over, pressed state, etc. But then when they are pressed, hovered, etc, draw as standard buttons. Would this not be easier? (And less 'code' ?)

clobrano:
Toolbar buttons are flat by design, this is why we don't have shadows

> As an aside, why not just theme these button as per regular no-flat buttons? ie. draw them flat when no mouse-over, pressed state, etc. But then when they are pressed, hovered, etc, draw as standard buttons. Would this not be easier? (And less 'code' ?)

Uhm, that was the version you didn't like

CDrummond:
@clobrano but when they are hovered, they are no longer flat. And if they are to be flat, then why have an internal shadow when sunken - as its now no longer flat.

> Uhm, that was the version you didn't like

? Not sure what you are referring to here. But if they were drawn the same (as in same border colour, etc), then they would be much more visible. This is how Adwaita draws these.

clobrano:
Sorry for the late reply, I double checked this and I think we were both right :)

The toolbar button, in master, is drawn the same as non-flat buttons in hover state *only*. However non-flat buttons always have a dark border, which is probably what you where talking about, which is not added nor made stronger by the hover state and so it is missing in the toolbar.

Feichtmeier:
I think it would be best if they would only receive a border and with the transparent background. This way they would not look awkward in any version (light/dark).

clobrano:
>  This way they would not look awkward in any version (light/dark).

but if we use button mixin, which already manages variants, why should be there any issue with light/dark versions?

clobrano:
So, the alternative I was talking about (button hover is not flat) is the one below

toolbar-hover

CDrummond:
As just a simple user, I'd much prefer it if the borders were the same for all buttons. At least all button-like elements (e.g. combo boxes) in a toolbar.

When you have an app that has combo boxes and flat buttons in its toolbar, then it will look odd if the combo boxes have a dark border, and the buttons have a lighter border on mouse-over. For example, look at evolution's message compose window.

Feichtmeier:
All the other elements have a very beautiful hover. So if anything should be adapted because their might be a reason to have the same hover effect, then the toolbar buttons should act like the flat headerbar buttons.

clobrano:
I can't recall any example of app with combo box in toolbar, so I can't really imagine it right now. Flat combo boxes are in Tilix and it looks good to me, of course they have a dark bg, so it might be different. 
We should remember that toolbar buttons are flat (not by our decision I mean) and our flat buttons don't have borders in normal state. Moreover having borders in toolbar will make it less clean

CDrummond:
@clobrano Evolution's message compose window has combo-boxes and flat buttons in the 'toolbar' under subject field.

clobrano:
Thanks @CDrummond. I like how we style it now. The only thing I would improve it the separator, that is not visible enough

image

image

clobrano:
Sorry for the mess. [Here](https://github.com/ubuntu/gtk-communitheme/pull/614) the other version's PR

PR #609, opened 2018-07-06
Fixes terminal tab backdrop hover

clobrano:
Give terminal unchecked backdrop tab hover proper border and background
colors

closes #591 
peek 2018-07-06 23-09

PR #613, opened 2018-07-07
Improved backdrop hover in nautilus tabs

clobrano:
Missing commit from #609 

improved hover effect also on other buttons in the tabs area that are not
part of the tab. Added also a darker border to the buttons above and to
the close tab button of unchecked tab. Checked tab close button gets a
background highlight instead.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr613`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr613` channel to test your snaps with those changes.

PR #621, opened 2018-07-11
 Fixed missing message-dialog csd window shadow

### LINKED ISSUE #600: Modal dialogs have no shadow
As can be seen in the following screenshot, the confirmation dialog has no shadow. Which is OK when the message window (its parent) is larger then it, but looks odd otherwise. Perhaps a slight shadow should always be added to these?

evolution

clobrano:
image

closes #600

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -4558,7 +4558,7 @@ decoration {
   }
 
   messagedialog.csd & {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr621`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr621` channel to test your snaps with those changes.

PR #626, opened 2018-07-14
colorswatch hover and roundness

### LINKED ISSUE #611: Flat GTK color picker selection
#MoreHairSplitting
The selection in GTK color picker has a highlight at the top and a shadow at the bottom. Could we make it flat? 

image

I'm not sure I understand the rounded corners. I really prefer this non-rounded style seen in Arc, but maybe it would be best to follow upstream here 🤷‍♂️ 

Arc:
image

clobrano:
closes #611

This is actually made of two commits, the second is optional, so wait for @madsrh to review it before merging it (EDIT: I forgot he's on vacation :D)

1st commit removes top border highlight from the widget
color-chooser-hover

2nd commit makes the color button squared 
color-chooser-square

### DIFF
--- a/Communitheme/gtk-3.0/_common.scss
+++ b/Communitheme/gtk-3.0/_common.scss
@@ -2,15 +2,6 @@
   @return unquote("alpha(#{$c},#{$a})");
 }
 
-$ease-out-quad: cubic-bezier(0.25, 0.46, 0.45, 0.94);
-$asset_suffix: if($variant=='dark', '-dark', '');
-$small_radius: 4px;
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr626`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr626` channel to test your snaps with those changes.

madsrh:
Looks good 😁

I think we should go with both 1st and 2nd commit 👍

PR #628, opened 2018-07-17
Merge all repositories into a single one

didrocks:
This is the merge of all communitheme related repositories and moved into subdirectory.

In addition, it fixes some issues in the build system (duplicated icons, wrongly assigned names) and ensure that we can change via a single "project_name" the whole directories and theme outputs.
A single source package is available creating the different binary packages (renamed as well).
The CI system is kept as well, and reshaped building the gtk-common-themes snap.
For backward compatibility, symlinks and other renaming are done in the snap itself, which will still keep the name "communitheme".


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/gtk-communitheme-pr628`.

You may want to download as well `gtk-common-themes` snap from `edge/gtk-communitheme-pr628` channel to test your snaps with those changes.

Feichtmeier:
Is the Communitheme Icon & Cursor option selected by default after installing and rebooting into the sesssion? This VM already had the snap so I needed to select the Communitheme Icon & Cursor option manually now after updating. The Suru cursor and Icon name could also be removed as an option since it is a bit confusing for new users in my opinion :)

Other than that, everything works with this PR! Great job - thank you!

didrocks:
Hum, if the VM is vanilla, nothing should need be selected again. This was after an update and a reboot? Can you try to reproduce the issue and details the exact step on a clean VM for me to reproduce?

Feichtmeier:
Alright - making a new vm now.

One more thing, could you set the default font size to 11 again, because of the issue cutting the window thumbnail description labels? It has been reported upstream and they are working on it (its "Only" if you select any other font than cantarell xD)

didrocks:
This is quite complex to qualify as a SRU, but we can try. Do you mind opening a bug against ubuntu-settings package in launchpad, following https://wiki.ubuntu.com/StableReleaseUpdates#SRU_Bug_Template?

I think the rationale and screenshot before/after, telling it's only affecting people using the communitheme snap is fine.

Feichtmeier:
> Can you try to reproduce the issue and details the exact step on a clean VM for me to reproduce?

Fresh install +
sudo snap install communitheme --channel=edge/gtk-communitheme-pr628
reboot
sudo apt install gnome-tweaks

1) Communitheme cursor + suru icons are selected (maybe we should remove the "suru" entry?)
screenshot from 2018-07-17 15-24-59
2) The font size issue (already reported on gnome gitlab):
vanilla install/session:
screenshot from 2018-07-17 15-18-37
After communitheme install/reboot:
screenshot from 2018-07-17 15-23-06
screenshot from 2018-07-17 15-25-08

So maybe you misunderstood me. You once changed this to 10 for aestitic reasons requested by @madsrh - but the issue still persists so I'd suggest to change it back to 11 (like vanilla session) to prevent this cut-off labels. Or maybe I misunderstood you and you know this all and yet I should report it on launchpad? =D

didrocks:
1. It sounds all good, even if it's set to Suru, it's our fork which is set. We can't change those names, and correct selection is applied, in a backward compatible way on purpose.  So, in a nutshell, all looks good to me :)
2. Yes, but as said when I made the package and integration for bionic, the tradeoff is that those settings are in a package in ubuntu, not in the snap. This is a change which thus need to go through the SRU process, hence asking for the bug and filing up the paperwork so that I can do that change.

Feichtmeier:
/cry
Okay, I make this later this evening then.

didrocks:
Got good feedback on the hub and here! Time to push the BGB, Big Green Button ;)

PR #647, opened 2018-07-22
Press effect only for the icon, add missing line.

Feichtmeier:
Fix missing line so the press effect is only for the icon:

peek 2018-07-23 00-35


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/communitheme-pr647`.

You may want to download as well `gtk-common-themes` snap from `edge/communitheme-pr647` channel to test your snaps with those changes.

PR #651, opened 2018-07-25
Update control

8none1:
Changed language slightly.  Not necessarily better English, but it reads better to me.


didrocks:
good! no new lintian warning, let me merge and upload to cosmic. Thanks!

Issue #474, opened 2018-05-23
Maybe headerbar text buttons shouldn't be flat?


godlyranchdressing:
This is the last time I'll bring this up. If we reach a consensus I'll drop it forever but I'm not a fan of the border around text buttons. I get the need for it, but it looks like something went wrong. So what if we excluded text buttons from being flat? In the mockup I used $hb_pathbar_bg to have them look consistent with the pathbar/stackswitcher buttons.

![finalboss](https://user-images.githubusercontent.com/27529229/40432337-73ef34ee-5e77-11e8-92bb-47a39217af3a.png)

Compared to the current:

![button-border](https://user-images.githubusercontent.com/27529229/40432686-43a6d340-5e78-11e8-99b9-a08e47a763e4.png)

madsrh:
Yes +1 I suggested this a long time ago

Feichtmeier:
Isn't the contrast between text-btns and image btns a bit harsh then?
What about using the border we have for text-btns for every btn instead?
Like a flat version of adwaita
![image](https://user-images.githubusercontent.com/15329494/40434311-654068be-5eae-11e8-8dca-6c3ce0e4cf30.png)

Paz-it:
I'm with you about it @clobrano ! Make them all flat 3d...Lets squire the circle!!! :))
I think @Feichtmeier   is right in the middle. I tend to agree with him about it. 
Another mockup maybe will help to clarify.

clobrano:
:D to be clear, I prefer them flat

godlyranchdressing:
Flatter suggestion, basically [this](https://github.com/ubuntu/gtk-communitheme/issues/152#issuecomment-368325780) but with a border:

![flatter](https://user-images.githubusercontent.com/27529229/40436073-2363496c-5e80-11e8-91ec-d6361bb2a3ba.png)

```
.text-button:not(.suggested-action):not(.destructive-action) { 
    @each $state, $t in (all the buttons) {
        &#{$state} { @include button($t, $hb_pathbar_bg, $tc, $flat:true); border: 1px solid _border_color($hb_pathbar_bg) } 
    }
}
```

Also another example since the combination of buttons in Gedit (which IIRC was what prompted the decision to add the border?) is a little odd. The headerbar doesn't look that cluttered with every app:

![settings](https://user-images.githubusercontent.com/27529229/40440397-f38897f4-5e8b-11e8-87be-638395896456.png)

clobrano:
[...]

PR #657, opened 2018-07-27
Make headerbar text buttons not flat

godlyranchdressing:
Uses $hb_pathbar_bg variables
Issue #474 

screenshot from 2018-07-26 21-34-38


clobrano:
It looks a bit confused IMO, moreover It's similar to pressed button, I think I prefer the current master

clobrano:
I guess we can close this

PR #658, opened 2018-07-27
Dim headerbar text button labels

godlyranchdressing:
Uses $headerbar_text_color so headerbar text buttons won't be as bright as a window's title, kind of like how pop os does it.

alternative to adding a border to text buttons to distinguish them

Issue #474

screenshot from 2018-07-26 21-36-39

### DIFF
--- a/gtk/src/gtk-3.0/_common.scss
+++ b/gtk/src/gtk-3.0/_common.scss
@@ -1559,12 +1559,14 @@ headerbar {
   label:backdrop { color: inherit; } // make sure to inherit the backdrop label color
 
   .title {
+    color: $headerbar_text_color;
     font-weight: 500;
     padding-left: 12px;
     padding-right: 12px;
   }
 
   .subtitle {
+    color: $headerbar_text_color;
     font-size: smaller;
     padding-left: 12px;
     padding-right: 12px;
@@ -1633,7 +1635,7 @@ headerbar {
                           (":backdrop:disabled", 'backdrop-insensitive'),
                           (":backdrop:disabled:active, &:backdrop:disabled:checked", 'backdrop-insensitive-active') {
                             @if $t == "normal" {
-                              &#{$state} { @include button($t, $c, $headerbar_text_color, $flat:true); }  
+                              &#{$state} { @include button($t, $c, $headerbar_fg_color, $flat:true); }  
                             }
                             
                             @else {
@@ -4455,7 +4457,6 @@ button.titlebutton {
   }
 
   &.close {
-    color: white;
     @include draw_circle($selected_bg_color, $close:true);
     &:hover {
       @include draw_circle($selected_bg_color, $close:true);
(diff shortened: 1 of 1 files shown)


Feichtmeier:
This grey looks like the buttons are disabled :)

madsrh:
@godlyranchdressing The title really isn't that important, so how about doing the opposite? Making the title slightly gray?

image

godlyranchdressing:
@madsrh sure

Screenshot from gedit:
dimmed title

madsrh:
That was fast @godlyranchdressing 😄 I'll give this PR a try 👍 

@Feichtmeier what do you think about this one?

Feichtmeier:
I prefer the current solution :cry

clobrano:
I think I am quite bond to current solution, so I would test this for the whole week in edge, so if you agree let's merge this tomorrow after stable release, what do you think?

Feichtmeier:
I don't understand why this is a solution to aarons initial complain, that the border around cancel buttons feels strange beside icon buttons without a border. If this is really a missmatch, which - as you know - is not my impression, that I'd rather remove the border around the cancel buttons again instead of either leveled buttons or these color changes, which remind me too much of disabled elements.

madsrh:
> I think I am quite bond to current solution...

I know what you mean. I'm happy with the current too

godlyranchdressing:
@Feichtmeier if I understand correctly, the border was added to distinguish text buttons from the window title. This is because headerbar button labels and window titles use the same text color. So to avoid confusion between the two without using a border, we can instead try making text buttons not flat, or simply changing the colors used to avoid confusion. Thinking about it, I don't think button labels and the window title should be the same color anyway.

The "disabled button" look I don't think applies here because _all buttons_ would be the new dimmed color and within context and actual use, they won't look disabled.

But @madsrh is right about the title being less important, so the title's text color is now dimmed instead.

Perhaps we can simply dim the title anyway if the border isn't removed?

clobrano:
Not sure about different label colors in the same headerbar, this is why I asked for a long try, I don't want to be biased, but the borders still have a better impact IMO. 
Was the problem the fact that only text buttons have it?

clobrano:
Originally borders have been added also to balance cases when there's only two buttons and one is green. So at the end of the day this won't be a solution for that use case

madsrh:
> Perhaps we can simply dim the title anyway if the border isn't removed?

+1, will you/someone create a seperate PR for that?

madsrh:
> I'd rather reduce the font weight.

I don't think that's a good solution. 

> We use the grey mostly for disabled / backdrop elements

It's a subtle difference. IMO it doesn't look disabled. Here I'm using #e1dfd3 🎨 

image

The more I'm experimenting with this, the more I like the current. I'm +1 for leaving things as they are 🤷‍♂️

Feichtmeier:
After some weeks of inactivity and the general agreement on don't changing the text I will close this now.

PR #662, opened 2018-07-28
Added hover effect to slider with mark

### LINKED ISSUE #397: Hover effect for sliders, switches and spin boxes

clobrano:
- changed assets.svg
- regenerated pngs with the render-assets.sh script

closes #397

scale-with-mark-hover

NOTE: when pressed, the handle gets a blue border. I thought about changing it to something else, but being a press effect and not a selection, orange is not necessary (e.g. a button pressed) and I think the blue suits better than any other color since matches with the trough, so I decided to keep it.

EDIT: I noted that the handle in scrolled windows is actually orange when pressed, so I decided to give this a try here, below the result. I am still not convinced though (moreover scrolledwindow handle is never blue)

scale-with-mark-orange-active


Feichtmeier:
Wow! Did you automatically changed the files?

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr662`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr662` channel to test your snaps with those changes.

clobrano:
> Wow! Did you automatically changed the files?

Yes, lukily it's way easier than I though. I changed the colors on inkscape and then regenerated the pngs with the script

Paz-it:
@clobrano , wow, nice! - I like them both, hard to choose :smiley:

PR #668, opened 2018-07-28
GTK2 asset updates

godlyranchdressing:
Correct colors for pressed button and scrollbar slider, frame and separators. Made the slider circular.

I'm not sure how to add shadows in inkscape and have the rendered asset show correctly: if I add a shadow to a button, the size of the button will jump from say 32x16 to 40x24 and this will get scaled down and appear fuzzy in a widget factory (see the checks and radio buttons).


eaglersdeveloper:
Is it better to use the square slider everywhere?
In the Unity 8 style?

Feichtmeier:
We changed that to round because the shell does not support squricles (we need to write this into the wiki)

clobrano:
Awesome :+1:

godlyranchdressing:
@eaglersdeveloper nope, I wish. The slider's handle is drawn using a bunch of shell-only custom CSS properties. 

```
.slider {
  height: 1em;
  color: #48b9c7;
  background-image: url("some-asset-thing.svg"); <---- this would apply to the slider's bar, not the slider itself
  -slider-height: 0.3em;
  -slider-background-color: rgba(87, 79, 74, 0.26);
  -slider-border-color: transparent;
  -slider-active-background-color: #48b9c7;
  -slider-active-border-color: transparent;
  -slider-border-width: 1px;
  -slider-handle-radius: 6px; <---- this only seems to determine the size of the handle, not the radius
}
```
It's a shame. I'd push for taking out the Shell's slider completely :laughing: and making the others the squricle, but ease of use for wide audiences, etc. would be an issue.

@clobrano you can try playing around with "background-size" for the GTK3 theme until the fuzziness is gone, that should work.

PR #675, opened 2018-07-30
Removed background-position property from gnome-shell common

### LINKED ISSUE #664: gnome-shell: Ignoring length property that isn't a number at line 186, col 24
/* Switches */
.toggle-switch {
  width: 65px;
  height: 22px;
  background-size: contain;
  background-position: right center; } <<<<<<

clobrano:
Removing background-position property since gnome-shell does not
support it

close #664

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -177,7 +177,6 @@ StScrollBar {
   width: 65px;
   height: 22px;
   background-size: contain;
-  background-position: right center;
 }
 
   @each $v in us, intl {


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr675`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr675` channel to test your snaps with those changes.

3v1n0:
Ah, I was about to fix the same...

However using pixels positioning in this case I guess is possible, as the bg-size is known (g-s doesn't support relative positioning, but it does support absolute ones).

clobrano:
Oh, I see. I was using the "accesibility" panel as test and I didn't see much difference in having the background position or not, but I'll try your suggestion in a next PR, thanks :+1:

PR #677, opened 2018-07-31
Gnome-ToDo: Remove the border radius of the taskrow button and the outline

Feichtmeier:
Because the border radius of the taskrow is hardcoded.

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr677`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr677` channel to test your snaps with those changes.

clobrano:
Sorry @Feichtmeier I can't see the difference, below is without this PR

image

Feichtmeier:
But I can ;0 look close to the button and the edges. You can see it better with the dark theme enabled in the extensions

PR #680, opened 2018-08-01
gnome-shell: merge with gnome-shell 3.29.90

3v1n0:
Removed some code cleanups (spaces and such) around that were just
creating diff with upstream version in a 3-way-merge but with no actual
need.

Kept some old definitions to make sliders and bars to continue working
in gnome-shell < 3.29.90 (although with some compromises, like we can't
show the slider background in osd-window's or that will break upstream
version).

PS: some colors (i.e `$destructive_color`) should be fixed in a different branch (or in this one if you point me the right ones) to match the designed solution.


Feichtmeier:
Hi @3v1n0 thanks for the PR

I wonder if this is a good point to create a 18.04 branch now before we merge this into master. Even if I didn't find any lines that change anything for the current version is probably safer.
What do you think @didrocks @clobrano ?

didrocks:
let's not diverge yet, the set of rules that aren't bacwkard compatible (those are my work upstream :p) are just duplicated here and are small. I would just add something like "START VERSION_SPECIFIC / END VERSION_SPECIFIC" to be easily greppable once we create a branch.

clobrano:
LGTM :+1: 

> PS: some colors (i.e $destructive_color) should be fixed in a different branch (or in this one if you point me the right ones) to match the designed solution.

what do you mean?

3v1n0:
> what do you mean?

For the overdrive color I've used the one defined by upstream (and by yaru itself), if you want to change that in future (or now) let me know.

But that is not functional to this change. Let me just push better comments for underlining the code for legacy support and it would be nice to have this merged ASAP as we're in transition to 3.29 for cosmic.

Cheers :)

clobrano:
Alright, waiting for your green light then :+1:

3v1n0:
@clobrano I pushed already. @didrocks  was fine as per #ubuntu-desktop discussion, so I think we're ready to go :)

PR #685, opened 2018-08-02
Apply MIR requested changes on packaging

didrocks:


didrocks:
Builds are passing, packaging only nitpicking changes, let's merge to unblock the MIR.

PR #693, opened 2018-08-06
Define a proper buttonbox style

### LINKED ISSUE #482: buttons in buttonbox not styled
image

Glade uses a strange box for buttons which seems to be styled wrong - compared to other buttons.

Adwaita styles them exactly like normal buttons:
image

clobrano:
Stop mimic stackswitcher with buttonbox

TODO
- [x] (gnome-software) hover on central button, when first is selected changes button size
- [x] backdrop background turns black
- [x] pressed effect is barely visible
- [x] (in glade) backdrop, last-child button border disappear (actually is just hard to see)

closes  #482

### DIFF
--- a/gtk/src/gtk-3.0/_apps.scss
+++ b/gtk/src/gtk-3.0/_apps.scss
@@ -316,6 +316,26 @@ terminal-window {
     box-shadow: inset 0 0 100px 100px $base_color;
 }
 
+// Emulate the stackswitchers look for this buttons
+.toolbar-primary-buttons-software {
+  margin: 0;
+  padding: 0;
+  border: none;
+  background-color: $hb_pathbar_bg;
+
+  &:hover {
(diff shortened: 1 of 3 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr693`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr693` channel to test your snaps with those changes.

Feichtmeier:
Nice work thank you carlo!

I think the hover looks still a bit odd:
peek 2018-08-06 17-00

The hovered gray area is a 2 or 3 pixels higher than the checked buttons :thinking:

clobrano:
> The hovered gray area is a 2 or 3 pixels higher than the checked buttons thinking

I understand what you mean, but considering that the buttons are on a different level this effect comes expected to me :thinking:

PR #696, opened 2018-08-09
Light entry and button fix

Feichtmeier:
In GDM the insensitive entry and buttons fades to a strange color 
because it was originally not designed to be light in drawing.scss.

This is fixed globally for light entries and locally for the light button now.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr696`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr696` channel to test your snaps with those changes.

PR #695, opened 2018-08-08
Default button and outline style

Feichtmeier:
Closes https://github.com/ubuntu/yaru/issues/306

PR #702, opened 2018-08-10
Revert "Default button and outline style"

eaglersdeveloper:
Reverts ubuntu/yaru#695

It looks bad with one button or/and dark theme, IMHO.


Feichtmeier:
IMHO it doesn't look bad. :man_shrugging: 

Btw the outline style didn't change. ;)

Feichtmeier:
We just merged this into master after a long conversation about this.

@madsrh @clobrano and me agreed that the black line made the buttons look to heavy and old - so I see no real reason to revert this 1 day after since prefer this design.

clobrano:
The right process would be to open a bug, not a PR :/

PR #704, opened 2018-08-10
Infobar backdrop improvements

Feichtmeier:
- make the label more greyish instead of black
- overwrite the default button backdrop bg color for infobars because the 
not(:insensitive) buttons merge their color with the purple then

Before:
screenshot from 2018-08-10 17-29-27

After:
screenshot from 2018-08-10 18-07-23

@madsrh @clobrano please check if you like it or not

### DIFF
--- a/gtk/src/gtk-3.0/_common.scss
+++ b/gtk/src/gtk-3.0/_common.scss
@@ -4213,6 +4213,8 @@ infobar {
       }
     }
   }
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr704`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr704` channel to test your snaps with those changes.

madsrh:
+1 nicely spotted

PR #716, opened 2018-08-14
Remove the headerbar border for geary new messages.

Feichtmeier:
Carlo introduced the awesome "light from above" headerbar color

Geary uses the global headerbar styling for their inapp-reply feature.
This is a minor GEARY ONLY fix which removes the headerbar border

Before:
screenshot from 2018-08-14 18-06-30

After:
screenshot from 2018-08-14 18-07-02


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr716`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr716` channel to test your snaps with those changes.

PR #719, opened 2018-08-14
Made message dialog all white

### LINKED ISSUE #698: Revert dialogues back to the upstream design
The current dialogues are styled like "classic" dialogues or to look more "generic".

While we had a big discussion about this already (and even a poll) I want to be annoying again and propose that we should revert this styling back to look more like the upstream dialogues.

Pros:
- upstream designers had big discussions about this and it is definitely a design decision to reduce the dialogues on simply the buttons, we want to be as close to upstream as possible anywhere else (I don't want to count the upstream adaptions but they should be quiet numerous) so we could do this here, too
- LibreOffice was pretty proud about having native dialogues finally implemented and with styling them back to the generic look, we somehow "ruin" this milestone
- beside upstream decisions when you look to the dialogues as if you never seen a dialogue before (to reduce the bias) then it is pretty clear what you want to do: press this button OR press that button so there shouldn't be a dire need for a titlebar since you should answer to the questions the dialogue asks you and not drag him around (EDIT: dragin is still possible with CSD dialogues)
- opinion: the upstream dialogues look more "modern"/minimal designed
- the upstream design looks like the GNOME shell dialogues - and we use the light dialogues now in the shell - having a white gtk dialogue and a white shell dialogue could be more usable for the average user who doesn't care if it is GNOME Shell or a gtk window wanting him to proceed/click

Cons:
- Some people want to drag the dialogue around (EDIT: dragin is still possible with CSD dialogues)
- Yaru could like to be on par with the rest of dialogues outside of GNOME

@godlyranchdressing @madsrh @clobrano @luxamman

Sorry for opening this barrel again, but I think if we have the possibility to stick to upstream we should do this.

Current yaru gtk dialogues: 
peek 2018-08-09 14-24
Adwaita gtk dialogues:
peek 2018-08-09 14-14
peek 2018-08-09 14-28

Yaru Shell dialogues:
peek 2018-08-09 14-26

clobrano:
closes #698

### DIFF
--- a/gtk/src/gtk-3.0/_common.scss
+++ b/gtk/src/gtk-3.0/_common.scss
@@ -3876,26 +3876,35 @@ calendar {
  * Dialogs *
  ***********/
 messagedialog { // Message Dialog styling
+  decoration {
+    border-radius: $large_radius;
+  }
   .titlebar {
-    background-color: $headerbar_bg_color;
+    background-color: $base_color;
     background-image: none;
-    border-color: transparentize(_border_color($headerbar_bg_color), 0.7);
-    border-style: solid;
-    border-width: 1px;
-    border-top-left-radius: $small_radius;
-    border-top-right-radius: $small_radius;
-    min-height: 32px;
+    border-color: $base_color;
+    border-top-left-radius: $medium_radius;
+    border-top-right-radius: $medium_radius;
+    color: $fg_color;
+    min-height: 20px;
 
     &:backdrop {
-      background-color: $backdrop_headerbar_bg_color;
+      background-color: $backdrop_bg_color;
     }
   }
 
   &.csd {
+    background-color: $base_color;
+
+    &.background {
+      border-bottom-left-radius: $medium_radius;
+      border-bottom-right-radius: $medium_radius;
+    }
+
     .dialog-vbox { margin-top: 14px; }
     .dialog-action-area button {
       @extend %button;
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr719`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr719` channel to test your snaps with those changes.

madsrh:
Looks good, but is it just me or did you also notice that it isn't centered? 📏 

untitled-1

clobrano:
That's because the text it's centered on the area below the titlebar. I'll add some padding :+1:

clobrano:
Ah, but there's a problem adding unconditional padding, since *sometimes* a title IS present and the padding isn't necessary

image

clobrano:
> Can you reduce the gap between the buttons and move them down a little bit to the bottom?

The padding between the buttons is the one used everywhere, however I actually move the buttons up respect normal, but just because I felt it looked better. Is there a specific reason you think it's better to lower them?

Paz-it:
@clobrano , I see what you mean, It make sense. I think I had the shell in mind ans thought it coould be nice to have these a bit look alike :smile  but looking at it again it's not going to be as beautiful anyway :cry

Feichtmeier:
Merging this now after general agreement on the look. 
Thanks @clobrano :tada:

PR #720, opened 2018-08-15
Change the no-notification and no-events icons

eaglersdeveloper:
Changed the no-notification and no-events icons to preferences-system-notifications-symbolic and appointment-symbolic icons respectively from Suru.
But in the notification bar remained old icons. Help wanted.

### DIFF
--- a/gnome-shell/src/no-events.svg
+++ b/gnome-shell/src/no-events.svg
@@ -13,8 +13,8 @@
    height="64px"
    id="svg3471"
    version="1.1"
-   inkscape:version="0.48.5 r10040"
-   sodipodi:docname="New document 5">
+   inkscape:version="0.92.3 (2405546, 2018-03-11)"
+   sodipodi:docname="no-events.svg">
   <defs
      id="defs3473" />
   <sodipodi:namedview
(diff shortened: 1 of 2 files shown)


Feichtmeier:
Great idea.

But I think the no notification icon is a bit misleading. Maybe just add a checkmark? Like 

No notifications
[X]

clobrano:
@eaglersdeveloper could you explain first what problem this PR should fix :)?

Feichtmeier:
Hmmm
Testing the newest cosmic at the moment (it has the yaru theme now by default after updating yay :dancer: )

Anyways,... gnome has updated the "no notifications" icon:

image

Can we close this then? I don't feel that the "no events icon" is that bad

clobrano:
+1 to close it

eaglersdeveloper:
@clobrano, current icons don't look in Suru style.

eaglersdeveloper:
I didn't see an operating system, in which in some hidden place there were old icons.

eaglersdeveloper:
If you are doing a redesign, you should take care that the remnants of the old design aren't preserved.

Feichtmeier:
@eaglersdeveloper 

But the new upstream icon is pretty neutral to our icon theme and an improvement over the current.
Maybe you could limit the PR to the no event icon only and update your initial, pretty confusing post :D 

- what is the state of this PR, would it be ready to merge?
- for what do you need help as you said "help wanted"

How about we close this and you open a dedicated issue where we track this down. I think opening PRs without issues is OK as long as you come up with a solid solution and description

clobrano:
> you are doing a redesign, you should take care that the remnants of the old design aren't preserved.

We aren't redesigning every single bit though. 
I agree with @feichtmeier, either you open a bug explaining what's the issue and say what you can and what cannot do, or open a clear PR which we are needed only for review it

clobrano:
Closing this now, waiting for a PR/BUG that follows [this](https://github.com/ubuntu/yaru/pull/720#issuecomment-414409572) indications

PR #722, opened 2018-08-16
Remove the uneven .counter-label margin because of the recent buttonbox changes

Feichtmeier:
Before:
image

The margin is uneven at the moment, after we stopped to mimic the stackswitcher in buttonboxes.

After: 
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr722`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr722` channel to test your snaps with those changes.

PR #741, opened 2018-08-23
Avoid changing progressbar bg color when selected in treeview

### LINKED ISSUE #726: Change progressbars color
I don't know what this bar is called (CellRenderProgress?), but IMO the white looks a bit odd. It's also a bit unnecessary here because of the orange selection which makes it very obvious which column is selected. Maybe we could just leave it blue?

Current:
image

Mockup:
image

Or here using the text selection color:
image

Is this perhaps the same type of bar that is found in Transmission?
image

Which would look like this with blue:
image

clobrano:
Progress bar color in treeview, when selected, gets a white color that
does not fit Yaru theme.

with this commit
- progress bar bg uses a lighter selection color for background
- progress bar fg (the label) turns white when selected. This
  will give a slight visual hint that the progress bar is part of a
  selection.
- progress bar and trough get a border to make the separation from the
  row below more evident

moreover:
- uniformed border-radius using Yaru values (small_radius)
- set border radius to zero internally to the trough, this makes more
  evident that the bar is progressing

closes #726

### DIFF
--- a/gtk/src/gtk-3.0/_common.scss
+++ b/gtk/src/gtk-3.0/_common.scss
@@ -2172,39 +2172,36 @@ treeview.view {
   }
 
   &.progressbar { // progress bar in treeviews
-    @if $variant == light { color: $base_color; }
-    $c: $neutral_color;
-    border-radius: 3px;
-    background-color: $c;
+    background-color: $neutral_color;
+    border-radius: $small_radius 0 0 $small_radius;
+    border: 1px solid $neutral_color;
+    color: $text_color;
 
     &:selected {
       &:focus, & {
-        border-radius: 3px;
-        color: $c;
-        background-color: $base_color;
-
-        &:backdrop {
-            color: $c;
-            background-color: $backdrop_base_color;
-          }
-        }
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Could we try to use a transparent trough? I think the triple color looks a bit heavy

clobrano:
> Could we try to use a transparent trough?

I don't think it's a viable solution

image

Feichtmeier:
I meant for all troughs :) but if you don't like it don't bother

clobrano:
Sorry,  I didn't get it. 
However we need to replace the grey area with something else to separate it from the background. We might use Borders and a white background, but in this case an empty progress bar would look like an entry

Feichtmeier:
We could also try the light orange from @madsrh 's mockup

clobrano:
new version

peek 2018-08-23 17-30

Feichtmeier:
I wound prefer the version :D

madsrh:
> I would prefer this last version with the lighter orange for the trough

+1

PR #752, opened 2018-08-28
gnome-shell: Don't use symbolic app icons in the top bar

jbicha:
We prefer full-color app icons in the top bar, in part because many
third-party apps don't provide symbolic icons yet so it is more
consistent to use full-color icons for all apps.

This fixes a regression accidentally introduced in 050ddce6

[LP: #1787401](https://launchpad.net/bugs/1787401)


didrocks:
Thanks a lot for fixing this Jeremy!

PR #753, opened 2018-08-28
headerbar button active state completely flat

### LINKED ISSUE #750: Improvements to (new) button box styling (or: consolidate toggled buttons)
While its a pitty that Software lost its (fake) stack switcher you came up with a universal, pragmatic and good looking solution for button boxes. 👍 I think it could be further improved by making the toggled state either weaker or stronger.

Currently there are three different styles for toggled header bar buttons which should be consolidated IMHO.

**Regular buttons:**
toggled button regular
**Text Buttons:**
toggled button text
**Button box:**
toggled button box

**1.** Since regular buttons give the clearest and most visible indication of a toggled state I would go that route and make text und box buttons the same: **strong dark shadow plus dark border**. That means removing the grey border for both text buttons and button boxes while toggled. Darkening all four sites would especially improve separation to the left and right element for button boxes.

**2.** If there is some intention behind the more subtle toggled text button style that could be another option for button boxes as I think the current grey plus dark border looks weird. I prefer option 1 though.

clobrano:
closes #750 

examples

image

image

image

### DIFF
--- a/gtk/src/gtk-3.0/_common.scss
+++ b/gtk/src/gtk-3.0/_common.scss
@@ -1741,10 +1741,6 @@ headerbar {
       &:first-child { border-width: 1px 0 1px 1px; }
       &:last-child { border-width: 1px 1px 1px 0; }
 
-      &:checked {
-          box-shadow: inset 0 1px rgba(0 , 0, 0, 0.6);
-      }
-
       &:backdrop {
         &, &:checked { border-color: lighten($inkstone, 3%); }
       }
--- a/gtk/src/gtk-3.0/_drawing.scss
+++ b/gtk/src/gtk-3.0/_drawing.scss
@@ -283,10 +283,13 @@
       $_border_top_color: darken($_border, 5%);
     }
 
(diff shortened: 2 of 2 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr753`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr753` channel to test your snaps with those changes.

clobrano:
Sure, I'll do it

Il mer 29 ago 2018, 01:25 Feichtmeier <notifications@github.com> ha scritto:

Feichtmeier:
@clobrano I am not sure if the reason for this fuzzy effect is the transparency or the color difference of border and bg?

clobrano:
I think it's that only the background is highlighted, while the border keeps the same color (the code you posted refers to non-headerbar buttons)

Feichtmeier:
@clobrano found another minor thing after using the snap on my host:
image
The active state for toolbar buttons must be diveded from the changes. I guess with asking for the headerbar_bg_color or something :dancing_men:

clobrano:
Good catch, the Semi flat toolbar

Issue #755, opened 2018-08-28
Button hover effect too strong in dark theme? [Discussion]


ya-d:
Again, not an issue. I tend to find the 3d hover effect for gtk buttons a bit too harsh (too black) in the dark variant of Yaru. It looks more subtle and "natural" in the regular light theme.

![button hover dark](https://user-images.githubusercontent.com/30781876/44751188-c5724d00-ab17-11e8-9534-2698e976c140.png)
![button hover regular](https://user-images.githubusercontent.com/30781876/44751191-c905d400-ab17-11e8-92e1-a3c56edd82eb.png)

*Not really apparent in pictures and I'm not the giffy guy. Look at it in person...* 😝

clobrano:
I think you should consider that the background in dark variant is... Dark :D so a lighter effect won't be visible enough.

ya-d:
Sure, it just feels like pure black instead of a drop shadow to me (even if its not). In the regular theme it looks more balanced IMHO which is why I would opt for a slightly less dark grey/black hue.

Feichtmeier:
just a general note: could we please chill a bit with the colors. I assign some of the new dark issues to me, but the colors are used in the strangest places. Let's not rush colors or anything only because UI freeze is soon. Atm we got a pretty solid composition, but I look into this button hover in drawing

clobrano:
can we close this? I think that colors are fine now

PR #756, opened 2018-08-29
Darken border_bottom_color less in the dark variant for btn hover

Feichtmeier:
peek 2018-08-29 11-33

Closes https://github.com/ubuntu/yaru/issues/755

@ya-d like this?

### DIFF
--- a/gtk/src/gtk-3.0/_drawing.scss
+++ b/gtk/src/gtk-3.0/_drawing.scss
@@ -251,7 +251,19 @@
       box-shadow: none;
     } @else {
       border-color: $_border;
-      border-bottom-color: darken($_border, if($c == $headerbar_bg_color or $c == $button_bg_color, 15%, 12%));
+      $_amount: 12%;
+      @if $c==$headerbar_bg_color {
+        $_amount: 15%;
+      }
+      @else if $c==$button_bg_color {
+        @if $variant=='light' {
+          $_amount: 15%;
+        }
+        @else {
+          $_amount: 4%;
+        }
+      }
+      border-bottom-color: darken($_border, $_amount);
       box-shadow: 0 1px transparentize(black, 0.85);
     }
   }


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr756`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr756` channel to test your snaps with those changes.

Paz-it:
The current is better imo.

Feichtmeier:
I am not sure. Maybe ya-d is right
Any other opinions @madsrh @c-lobrano

clobrano:
Honestly, I like this, but it introduces it a complex code, so I'd keep the current as well

Feichtmeier:
Lets wait for ya-d and mads and then eventually close it

ya-d:
I like it as it tones down the naturally black shadow. 👍 Thanks for the pr, @Feichtmeier !

madsrh:
Sorry for not being much help here, but I'm good with both solutions.

This PR looks slightly better (more elegant IMO) but I didn't notice the all black shadow before ya-d mentioned it.

clobrano:
Alright, if we agree that the effect is good, the complexity can be reduced (even later). I was thinking about doing something like that anyway after UI Freeze :snowflake:

Feichtmeier:
This way we don't need to change the border color. Because I think it is already at the contrast edge. I wonder if this clashes with the other pr because it is almost in the same part

Feichtmeier:
I could reduce the inner if else with an inline  if check. The rest is needed because we have the dark header bar and a dark and light variant with the same headerbar color. That's why we need to check this three possibilities so often in drawing. Actually we should remove all these lightness checks all together of we stay with our dark headerbar (which I hope and which I don't think anyone wants to change)

clobrano:
Yes I agree, but let's do it in another PR, taking advantage of the UI freeze time, when all the changes won't affect the general style

PR #761, opened 2018-08-31
Gnome shell upstream less diff

3v1n0:
I've given meld another chance to check differences between upstream and yaru (@didrocks!) to reduce the diffs, and setting the values to default when possible (like `.user-widget-label` that was changed in g-s 3.29.30, but not reported here).

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -190,9 +190,14 @@ StScrollBar {
   width: 65px;
   height: 22px;
   background-size: contain;
-  background-image: url("toggle-off.svg");
-  &:checked { background-image: url("toggle-on.svg"); }
 }
+
+  @each $v in us, intl {
+    .toggle-switch-#{$v} {
+      background-image: url("toggle-off.svg");
+      &:checked { background-image: url("toggle-on.svg"); }
+    }
+  }
 
 /* links */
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Great thank you. About the no the notification code. We have a pr which handles some of the code by using a shortcut to a new %light_button because we use this styling in three different places. So we would reduce the code there. Is this rather wanted or not because it creates more diff?

clobrano:
Yes, I added a `$light` variable in _colors.sccs because I found it more effective than writing `$variant=="light"` every time, but indeed it increases the diff

clobrano:
Sorry @3v1n0, a couple of merges introduced a small conflict. Could you double check?

3v1n0:
@clobrano no problem... I've rebased.

Of course there's no problem with having more diff with upstream if it's something functional for yaru, but when it's possible is better to stay closer as possible (comments, spaces and everything :)), in order to make easier to backport upstream changes.

clobrano:
button.flat and toolbar

https://github.com/ubuntu/yaru/blob/ec829c88d038ed80e5ae01a24d91997f34452565/gtk/src/gtk-3.0/_common.scss#L570

https://github.com/ubuntu/yaru/blob/ec829c88d038ed80e5ae01a24d91997f34452565/gtk/src/gtk-3.0/_common.scss#L1482

PR #762, opened 2018-08-31
Adapt flat suggested action buttons

Feichtmeier:
I have forgotten suggested and destructive action buttons with $flat==true

@clobrano I just checked if there are any suggested action btns with flat==true outside of the headerbar and didn't find any - did I miss something?


clobrano:
button.flat and toolbar

https://github.com/ubuntu/yaru/blob/ec829c88d038ed80e5ae01a24d91997f34452565/gtk/src/gtk-3.0/_common.scss#L570

https://github.com/ubuntu/yaru/blob/ec829c88d038ed80e5ae01a24d91997f34452565/gtk/src/gtk-3.0/_common.scss#L1482

clobrano:
LGTM :+1:

PR #764, opened 2018-09-01
Issue746/dark insensitive borders

### LINKED ISSUE #746: Insensitive borders does not look insensitive
It's a bit odd that the insensitive spinbutton border is darker than the active one - IMO it makes it look active. 

Current:
image

If I switch the borders (dark border on active and gray border on insensitive)
image

Here's the current (for reference):
image

Perhaps doing something like this?
image

This is just extending the already used border color from the dropdown arrow:
image

I also tried a brighter border, but that didn't work: 🗑 
image

@Feichtmeier ?

Also shouldn't the background be transparent like with the switches? Or is it fine because radios have a lighter background too 🤷‍♂️ 

image

---
EDIT: sorry if I tweaked the main post, but let's keep a list of the changes here ;)

## spinbutton insensitive
- [ ] border be the same color as insensitive button
- [ ] (vertical) number box border be the same color as insensitive button

## entries insensitive
- [ ] border be the same color as insensitive button (see comboboxes)

~## radio/check and switch insensitive~ see  #763
~- [ ] (unchecked) background be a lighter version of normal radio/check and not transparent~

clobrano:
closes #746 

EDIT:
Since CI is still failing, I added some pictures
normal

image

image

backdrop
image

image

### DIFF
--- a/gtk/src/gtk-3.0/_colors.scss
+++ b/gtk/src/gtk-3.0/_colors.scss
@@ -14,7 +14,7 @@ $selected_bg_color: $orange;
 $borders_color: if($variant == 'light', darken($bg_color, 18%), darken($bg_color, 4%));
 $link_color: $blue;
 $link_visited_color: if($variant == 'light', darken($link_color, 20%), lighten($link_color, 10%));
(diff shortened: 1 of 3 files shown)


madsrh:
@clobrano I love this 👍 Insensitive borders now look insensitive 😄

PR #779, opened 2018-09-05
Adapt selection-mode buttons in headerbar to the new headerbar styling

Feichtmeier:
Also add a transparent white border for text-buttons like we did with 
the other text-buttons

image

image

image

peek 2018-09-05 12-39

@clobrano @madsrh are you okay with this?


madsrh:
Nicely spotted @Feichtmeier

PR #784, opened 2018-09-05
Use stable docker image

didrocks:
Snapcraft tip is currently sending permission denied.
Reference: https://forum.snapcraft.io/t/permission-denied-when-building-with-snapcore-snapcraft/7186/15


didrocks:
And build pass!

PR #806, opened 2018-09-10
drawing/switch dark: made insensitive border as in spinbutton

### LINKED ISSUE #746: Insensitive borders does not look insensitive

clobrano:
closes #746

### DIFF
--- a/gtk/src/gtk-3.0/_drawing.scss
+++ b/gtk/src/gtk-3.0/_drawing.scss
@@ -660,6 +660,13 @@
   border-top-color: darken($bg, 20%);
   box-shadow: inset 0 1px 1px 0 if($bg==$dark_fill, transparentize($bg, 0.2), rgba(0, 0, 0, 0.15));
 
+  $_insensitive_border_color: null;
+  @if $bg == $dark_fill {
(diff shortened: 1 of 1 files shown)


clobrano:
image

image

image

Feichtmeier:
Tried the pull looks good. Mad likes it too. Merging

PR #812, opened 2018-09-11
apps/nautilus: fix circular button backdrop border color

clobrano:
Backdrop border color of nautilus' circular button (stop loading button)
stays dark in backdrop. This commit makes it transparent in backdrop


Feichtmeier:
screenshot from 2018-09-11 14-01-04

PR #813, opened 2018-09-11
Use small radius for nautilus widgets/canvas items

Feichtmeier:
Before: 
image

peek 2018-09-11 14-13

After: 
image

peek 2018-09-11 14-14

### DIFF
--- a/gtk/src/gtk-3.0/_apps.scss
+++ b/gtk/src/gtk-3.0/_apps.scss
@@ -111,6 +111,10 @@ list.tweak-categories separator {
       &:selected { color: transparentize($selected_fg_color, 0.3); }
   }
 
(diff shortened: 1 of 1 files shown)


madsrh:
@clobrano 
While you were merging this PR 🤣 @Feichtmeier and I was discussing this:

image


To be honest, I don't like it. I would prefer this **-1px** suggestion @Feichtmeier came up with, BUT  it cuts the label. I would say that it's fine because this is a very rare use case 🤷‍♂️ 

image

As you easily can see here, there is more top padding than bottom 📏 
image

clobrano:
> While you were merging this PR rofl @Feichtmeier and I was discussing this:

oops

PR #814, opened 2018-09-11
Use -2px offset and change outline color to white when selected

Feichtmeier:
Part two of https://github.com/ubuntu/yaru/pull/813#issuecomment-420400963

Using a -2px margin would work in the default case (no dim-label activated):

image

If one uses the dim-label option for more information it still works:
image

Only, in a corner case if you activate two or all three dim-labels the offset touches the lowest label:
image
image

But the labels are only displayed if the zoom of the view is high enough:
peek 2018-09-11 22-18

Sadly, any padding or margin does not work here :man_shrugging: 
Sorry I'd say we are good to go, last decision to @madsrh @clobrano

### DIFF
--- a/gtk/src/gtk-3.0/_apps.scss
+++ b/gtk/src/gtk-3.0/_apps.scss
@@ -113,8 +113,11 @@ list.tweak-categories separator {
 
   .nautilus-canvas-item {
     -gtk-outline-radius: $small_radius;
+    outline-offset: -2px;
+    &:selected {
+      outline-color: white;
+    }
   }
-
 }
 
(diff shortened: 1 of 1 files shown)


madsrh:
Nice work, thanks @Feichtmeier 

> Sorry I'd say we are good to go, last decision to @madsrh @clobrano

I'll let @clobrano decide if we should use this PR or keep https://github.com/ubuntu/yaru/commit/660e4c630abffc8aac462fd6d82dcf107f00aeb2

clobrano:
very nice :+1:

PR #816, opened 2018-09-12
Support dark variant

didrocks:
Add a dark variant support in the build tools for gtk3.0 directory.
Strip them from the debian and snap packages as we don't support them officially yet.
Added some cleanups and better way create generated filename for themes (+ support variants, while keeping same name for icon and cursor theme).

Once gtk2 support is ready, it only need to copy the `gtk/src/light/gtk-2.0` to `gtk/src/dark/gtk-2.0` (or symlink what can be symlink) and uncomment `subdir(gtk-2.0)` in `gtk/src/dark/meson.build`.


clobrano:
Nice and clear :+1:

didrocks:
Thanks! and CI pass (apart from the well-known "SnapRevs update failed"), merging then :)

PR #832, opened 2018-09-17
Updated url to pull gtk-common-themes from gitlab.gnome.org

kenvandine:
gtk-common-themes has moved from github to gitlab.gnome.org


didrocks:
:+1: Thanks!

PR #833, opened 2018-09-17
Adapt headerbar entries to the buttons in headerbar

Feichtmeier:
As we changed the elements in the headerbar to be flat, we have forgotten to adapt this to the entries. 

This styles the entries like buttons (flat) in foreground and backdrop.

Before:
peek 2018-09-17 10-25

After:
image

image

peek 2018-09-17 13-11

@madsrh @clobrano please check design
@didrocks this has been forgotten - so it should be considered to be a fix and thus merged into cosmic?

### DIFF
--- a/gtk/src/light/gtk-3.0/_drawing.scss
+++ b/gtk/src/light/gtk-3.0/_drawing.scss
@@ -60,14 +60,20 @@
     $_bc: if($c == $headerbar_bg_color, lighten($hb_pathbar_bg, 5%), $c);
     color: if($tc != $fg_color, transparentize($tc, .1), $fg_color);
     border-color: if($c != $base_color, transparent, $borders_color);
-    @if $variant == 'light' {
-      border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+    @if $c!=$headerbar_bg_color {
+      @if $variant == 'light' {
+        border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+      }
+      else {
+        border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+      }
+      @include _shadows($_entry_edge);
     }
-    else {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr833`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr833` channel to test your snaps with those changes.

madsrh:
I just tried the snap and it LGTM 👍

Feichtmeier:
It's so awesome that snap testing is back :)
Alright, let's see what carlo says to the code

gtk/src/light/gtk-3.0/_drawing.scss
```diff
@@ -60,14 +60,20 @@
...
     border-color: if($c != $base_color, transparent, $borders_color);
-    @if $variant == 'light' {
-      border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+    @if $c!=$headerbar_bg_color {
+      @if $variant == 'light' {
+        border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+      }
+      else {
+        border-top-color: darken($_bc, if($c != $base_color, 12%, 20%));
+      }
```
clobrano: it seems that we do the same in both light and dark branch. Are you sure we need this if/else?

Feichtmeier: Seems pretty useless. I change it

Feichtmeier: Ah no, I remember. For some reason, this would end up in this border top color for dark variant entries:

image

Whereas the current looks like this:

image

madsrh:
Isn't this the same issue in Image Viewer?

image

Feichtmeier:
@madsrh yes but this should be fixed in EOG too

peek 2018-08-06 17-00

Issue #714, opened 2018-08-13
Style GNOME Disks


madsrh:
GNOME Disks uses the default selection color for the selected disk. In Yaru's case, that means we have a orange background with dark text over it. 

Current:
![image](https://user-images.githubusercontent.com/3986894/44059169-bf0bf6d2-9f50-11e8-9cf5-3a81b7433101.png)

And in backdrop:
![image](https://user-images.githubusercontent.com/3986894/44059841-ed5fe406-9f52-11e8-8b38-1c727154bc8c.png)

Can we change only the text color of the selected disk and keep the other ones dark?
![_2colors](https://user-images.githubusercontent.com/3986894/44059382-751a5fe0-9f51-11e8-9e98-cd72844adf07.jpg)

Another solution would be to use a brighter background (like gray) and add the orange selection border:
![_border](https://user-images.githubusercontent.com/3986894/44059406-90003b5e-9f51-11e8-9e7b-6b8d18ce75fb.jpg)

This is merely an enhancement, so If this issue is too much individual application styling, feel free to close it.

Paz-it:
@madsrh , Just for competing to the dark : 
![screenshot from 2018-08-14 02-03-22](https://user-images.githubusercontent.com/37431503/44062817-b837c4ba-9f66-11e8-932c-3f1d96ee0754.png)
Hmm... The gray looks nice imo ..

clobrano:
Selected disk background is decided in gnome-disks custom css, so we cannot change it.
About the label, is it true that in Ambiance the labels are white, however it looks like in Yaru (and Adwaita as well) it uses `@theme_fg_color` which is our `fg_color` that is `$inkstone`

Feichtmeier:
I actually think, though gnome-disk-utility does not provide any style class or ID for these selected widgets, we could do the following "wildcard":

![peek 2018-09-18 11-34](https://user-images.githubusercontent.com/15329494/45678493-cc98e380-bb36-11e8-90b8-de6fe01c4bde.gif)

@clobrano 
Could you think of any negative effect for this? Replace "white" with "selected_fg_color" ofc

clobrano:
With wild cards the risk is high, but we can try a PR

PR #834, opened 2018-09-18
GNOME disks: Selected widgets should have the a light text color.

Feichtmeier:
**Testing** PR for gnome disks (https://github.com/ubuntu/yaru/issues/714#issuecomment-422325117) and every other widget that is selected (they should all have orange bg and bright fg)

Currently, gnome-disk(s-utility) has it's selected partitions colored with black text-color even if they are selected. Which makes the text hard to read with our default selected background color: orange
peek 2018-09-18 12-44

This change applies widget:selected { color: $selected_fg_color; } high up in _common.scss 
peek 2018-09-18 11-34

I can not think of a selected widget which should not have a bright text color. But we should check carefully if this does not change anything else.

@madsrh @paz-it @clobrano please test this PR if you have time and check for all the selections of UI elements and if this has a negative effect

-->  does this overwrite any selected element which should NOT have a light text color/fg color?

### DIFF
--- a/gtk/src/light/gtk-3.0/_common.scss
+++ b/gtk/src/light/gtk-3.0/_common.scss
@@ -90,6 +90,8 @@ $popover_shadow: 0 2px 5px transparentize(black, 0.75);
   }
 }
 
+widget:selected { color: $selected_fg_color; }
+
 .view,
 %view {
   color: $text_color;


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr834`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr834` channel to test your snaps with those changes.

Paz-it:
Tested! Looks much better in Disks :100: % 
Didn't spot any regression.

madsrh:
> Didn't spot any regression, yet...

+1 ...yet 😉

Feichtmeier:
Let's keep testing for some more days and then merge if nothing is spotted

Issue #835, opened 2018-09-19
Yaru builds incorrectly with Meson < 0.45


jhenstridge:
At the moment, Yaru claims it needs Meson >= 0.42.  However, if built with Meson < 0.45 it will install incorrectly.

The install directive for the icon theme uses the `strip_directory` directive [that was added in Meson 0.45](https://mesonbuild.com/Reference-manual.html#install_subdir).  So with old releases the icons get installed into `$datadir/icons/Yaru/Suru` rather than `$datadir/icons/Yaru`.

PR #836, opened 2018-09-19
build: require Meson >= 0.45 for strip_directory support, fixing #835

jhenstridge:


didrocks:
Thanks for spotting this! Do you mind updating debian/control build-dep as well?

PR #838, opened 2018-09-19
Make .inline-toolbar.horizontal.sidebar-toolbar transparent in backdrop

Feichtmeier:
We have forgotten to style this in backdrop

Before:
image

After:
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr838`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr838` channel to test your snaps with those changes.

Issue #847, opened 2018-09-21
Shell menus hard to recognize against dark backgrounds


matthewpaulthomas:
yaru-theme-* 18.10.4, Ubuntu Cosmic

In a dark-themed app, menus are dark grey, #474747:

![menu from dark-themed app](https://user-images.githubusercontent.com/19801137/45892206-56f37880-bdbf-11e8-9649-186b5fe7c023.png)

But shell menus are even darker, #1d1d1d:

![menu from shell launcher](https://user-images.githubusercontent.com/19801137/45892217-61157700-bdbf-11e8-9033-90bd246a0cc0.png)

Against a dark background, this makes it difficult to see where the menu begins and ends.

This is especially the case for an “application menu”, since it’s often short, so much of its background is a title bar and menu bar that are only slightly lighter, #262424.

![menu from shell top bar](https://user-images.githubusercontent.com/19801137/45891456-8bfecb80-bdbd-11e8-9f21-758f9acc52e5.png)

The border between them is a single-pixel semi-transparent line.

![menu from shell top bar cropped](https://user-images.githubusercontent.com/19801137/45891625-f57eda00-bdbd-11e8-9d85-4ddf0688c6fb.png)

This could be clearer in a variety of ways:
- Use the lighter grey from dark-theme app menus (#474747) for shell menus as well.
- Make the menu border thicker.
- Make the menu border brighter.
- Make the menu border more opaque.

[[Originally reported in the Yaru forum](https://discourse.ubuntu.com/t/call-for-participation-an-ubuntu-default-theme-lead-by-the-community/1545/1750).]

Feichtmeier:
The dock, top panel, popups and OSD (volume lower/higher popups) all use the same color

The popups "grow" out of the top panel and the dock.

Therefor having a different color for the popups only looks bad (we tried that some months ago). :-1: 

What we would have loved to do is, adding a box-shadow like the OSDs have (you also mentioned that on the hub)
Sadly this is not possible in the current gnome shell unless we remove the box-shadow, like for example the Matera theme does it: https://github.com/nana-4/materia-theme/blob/master/src/_sass/gnome-shell/_common-3.28.scss#L624

so we could have a trade: no arrow but with a box-shadow 

Increasing the border brightness and thickness was also tried back then in time, this somehow breaks the "elegant" "dark silk" look the shell menus/OSDs/window switchers currently have

I make a testing PR with the box shadow and we can try it

Feichtmeier:
[...]

PR #849, opened 2018-09-21
Shell popup usability experiments

Feichtmeier:
Testing PR 
Eventually
Closes https://github.com/ubuntu/yaru/issues/847

@matthewpaulthomas @madsrh @clobrano  @paz-it please test it 

- remove the box-pointer to be able to draw a shadow
- use the same shadow as in OSD popups
- switch osd_border color from fg_color to white (transparentized)
- make the border more opaque

screenshot from 2018-09-21 18-42-09
screenshot from 2018-09-21 18-42-30


clobrano:
Sorry guys, I won't remove the box pointer. My opinion is that it would be a regression other that a big deviation from upstream

Issue #853, opened 2018-09-23
usb-creator-gtk dialog misses margin


Feichtmeier:
Before
![screenshot-20180923102342-844x132](https://user-images.githubusercontent.com/15329494/45925727-61c32000-bf1b-11e8-80ab-7029daa0c63f.png)
After adding margin
![screenshot-20180923102650-922x422](https://user-images.githubusercontent.com/15329494/45925726-61c32000-bf1b-11e8-971c-95ccae052473.png)

PR #858, opened 2018-09-23
Usb creator margin and geary fix

Feichtmeier:
- Add a margin to usb-creator's dialog
Closes https://github.com/ubuntu/yaru/issues/853

- Correct geary headerbar (had a mistake in my commit back then which limited this to headerbar to a specific style class)


Feichtmeier:
Tested on my host 
image
image

PR #873, opened 2018-09-25
Fixing entry bleeding background in gtk2

clobrano:
entry's assets contour was defined as transparent in assets.svn, however
it was not actually transparent when rendered, causing the "bleeding"
background effect.

Fixing it setting the contour backgorund color to the base color #f5f6f7

before
image

after
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr873`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr873` channel to test your snaps with those changes.

Paz-it:
Well done! :clap:

PR #875, opened 2018-09-27
light gtk2: fixed radio/check buttons

### LINKED ISSUE #843: gtk2 radio button has the wrong check sign
Here is the check sign in gtk2, used in both light and dark theme

screenshot from 2018-09-20 22-44-49

here is the one used in GTK3

screenshot from 2018-09-20 22-46-25

clobrano:
Currently the asset for radio checked uses the wrong image (a tick in
place of a circle), however, since the menu radio checked has already
the right image (except for the colors), this commit:

- fixes colors of menu-check/radio buttons
- let re-use the menu-check/radio button assets outside of menus as well

closes #843

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr875`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr875` channel to test your snaps with those changes.

PR #878, opened 2018-09-30
gtk2 assets + main.rc: made scrollbars more similar to GTK3 ones

clobrano:
- background white
- border transparent
- reduced scrollbar width by 3px

before
image

after
image


Paz-it:
Good but half solution .  When you look at the top headerbar double-line is drown -one for header and one for the window frame. on the right top side, the last separator is just there hanging and the top line above the scrollbar is thinner . that way it looks unfinished imo.
drawisland

PR #892, opened 2018-10-05
Remove Shell artefacts from gtk-common-themes

didrocks:
We don't need it this snap only covers GTK assets.


didrocks:
@Feichtmeier: FYI, tests pass thanks to this (got a SnapRevs update failed again, but it's due to the store randomness and not a build failure anymore).

Issue #885, opened 2018-10-02
Sliders have a fuzzy border in the dark theme


Feichtmeier:
![screenshot-20181002231434-51x36](https://user-images.githubusercontent.com/15329494/46377587-45e91800-c699-11e8-8cdb-2d49e4a49302.png)

![image](https://user-images.githubusercontent.com/15329494/46377651-76c94d00-c699-11e8-984e-f95661292a22.png)

Possible solutions:
- remove the border
- make the border same color as the knob

clobrano:
Using the same color of the knob makes this weird effect

![image](https://user-images.githubusercontent.com/2883614/46799076-15b60f00-cd54-11e8-8678-9e970538bc06.png)

Better transparent I believe
![image](https://user-images.githubusercontent.com/2883614/46799134-3aaa8200-cd54-11e8-96da-7007c539fba0.png)

Feichtmeier:
Fixed with https://github.com/ubuntu/yaru/issues/909

PR #909, opened 2018-10-11
common/scale circular slider: made border transparent in dark variant

clobrano:
In the dark variant, gtkscale slider border looks fuzzy, probably due to
the different contrast between a dark background (in the light variant
the background is brighter and the border does not look fuzzy).

With this commit, the gtkscale slider border is set to transparent to
dark variant only to have a sharper border.

NOTE: a similar change is to be made to the 'scale with marks', however
that is a svg image and belongs to the gtk/dark folder which has not yet
landed on master.

see #885


Feichtmeier:
I'm curious: doesn't this make the whole knob smaller?

clobrano:
I thought the same at first, but the border is above the background, not
aside, so making it transparent just shows the background color. Making the
border white, instead, produce some glitches, this is why I opted for
transparent.

Feichtmeier:
Awesome!

Can we merge to master again btw?

clobrano:
I think so, however we probably need to change the CI scripts to build snaps for cosmic

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr909`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr909` channel to test your snaps with those changes.

PR #924, opened 2018-10-20
Reduce dock icons padding to 1px 2px

eaglersdeveloper:
gifmaker org_7psssm

In Unity 8, an icons padding sought to be less.
I reduced them. Now the dock no longer seems so fat.
Especially at 1366x768.

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -181,10 +181,15 @@
   &.dashtodock.right #dash {
     // don't let the first icon be too close to the shadow + panel
     background: $dash_bg_color;
-    padding-top: 2px;
-    padding-bottom: 2px;
+    padding-top: 1px;
+    padding-bottom: 1px;
   }
-
+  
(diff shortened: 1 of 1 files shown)


clobrano:
I like the idea, but as side effect the border between the dock size and the icon highlight is, maybe, too small (1px only)

image

btw the dock size can be reduced from settings

@madsrh, @Feichtmeier what do you think?

madsrh:
Everybody loves more screen space so +1
It doesn't look too cramped to me, but earlier to tell in the real

Feichtmeier:
I would like to play around with this later before we would eventually merge it.
When the horizontal padding differs too much from the vertical padding the launcher could look strange. As mads said need to try it first

Feichtmeier:
This is the current padding, which is exactly even on both sides:

image

Feichtmeier:
Ok, we discussed this and decided to not merge this, since we prefer the current padding + you can change the dock size here:

image

Thanks for the PR would be great if we could discuss something like this before you make PRs so we don't end up in situations like this one. Thanks anyways for contributing :+1:

PR #929, opened 2018-10-22
Really don't ship dark theme

didrocks:
We didn't intend to ship dark theme on cosmic until we have proper GTK2
support for it.


didrocks:
clobrano +1 on IRC

Issue #928, opened 2018-10-22
Dash to dock notification counter has wrong color


clobrano:
## Expected Behavior
notification counter is green

## Actual Behavior
the notification counter is red with green border
![screenshot from 2018-10-22 12-32-01](https://user-images.githubusercontent.com/2883614/47288702-79afc180-d5f7-11e8-8276-026633cc3834.png)

## Specifications
on Yaru master

Feichtmeier:
I had this problem once when I fiddled around with the dash to dock configs. The result were that I needed to reinstall Ubuntu, because I lacked the skills to undo what I did :)

Is this on a fresh installation?

clobrano:
So, it is probably due to the fact I reinstalled Yaru from source?
If that so, and nobody can reproduce this, I'll close the bug

Feichtmeier:
Let me check on my VM first

Feichtmeier:
![image](https://user-images.githubusercontent.com/15329494/47293168-a66cd480-d609-11e8-951c-264240f29ab1.png)

I have the dock and panel transparency difference (which is new, right?) but the color of the badge is correct :thinking:

clobrano:
Yep, on bionic at least the dock is more transparent. In cosmic is it ok, right?

Feichtmeier:
Oh no :( fresh cosmic VM and it happens here as well :(
![image](https://user-images.githubusercontent.com/15329494/47357785-09bd3c00-d6c8-11e8-8ac7-d42cf104557b.png)

Imho this is a really high prio, since those green badges are kind of iconic for the whole look

Feichtmeier:
When one uncomments the styling of the notification badge inside Ubuntu-dock/stylesheet.css the red badge overlay is gone.
So we could make a pull request at the Ubuntu dock repo or any other ideas? @didrocks @clobrano

clobrano:
I'd like to have a better understanding of the issue, first. I'll check it
again this afternoon, then I can ask upstream, they're always very
supportive

Il mer 24 ott 2018, 12:11 Feichtmeier <notifications@github.com> ha scritto:

Feichtmeier:
Okay great!
Best would be if we wouldn't use that second CSS file at all (I think this is how it was before?). To have full control over the theming

clobrano:
I understand, but ad usual, it can be installed in other OS, I think

Il mer 24 ott 2018, 13:00 Feichtmeier <notifications@github.com> ha scritto:

[...]

PR #931, opened 2018-10-25
Make the badge border temporarly dark red

Feichtmeier:
**This should not be merged into master but instead into a new branch "Cosmic" or something
to not break the bionic snap theming**

Currently there is a bug in Cosmic, which prioritizes the stylesheet.css from /usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/ over our dock theming  in the yaru shell theme (https://github.com/ubuntu/yaru/issues/928)
And we end up with a color overlay of red on our green badge which looks really messy and buggy

screenshot from 2018-10-22 12-32-01

This commit changes the badge border to a dark red to remove this "bugged" look. (the dashtodock css only overrides the background and does not style the border, thus our border style is used)

@didrocks I would call this rather important because currently it looks very broken in cosmic. 
We should revert this ASAP when a fix is applied in either dash to dock or ubuntu-dock

After:
image

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -125,7 +125,7 @@
     background-color: $green;
     padding: 0.2em 0.5em;
     border-radius: 1em;
-    border: 1px solid darken($green,10%);
+    border: 1px solid darken($red,10%);
     box-shadow: -1px 1px 5px 0px transparentize(black, 0.5);
     font-weight: bold;
     text-align: center;
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -122,7 +122,7 @@
 
   .notification-badge {
     color: rgba(255, 255, 255, 1);
-    background-color: $green;
+    background-color: $red;
     padding: 0.2em 0.5em;
     border-radius: 1em;
     border: 1px solid darken($red,10%);


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr931`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr931` channel to test your snaps with those changes.

Feichtmeier:
This should not be merged into master but instead into a new branch "Cosmic" or something
to not break the bionic snap theming

didrocks:
Hum, this is complex. Anyway we investigate rather the fix in dash to dock so that we don't workaround and have a dance to do? Indeed, it will request creating 3 branches: one for the snap, one for cosmic and one for master -> d, which will add a very big setup complication, changing travis configuration, changing as well gtk-common-themes.
Also, it means then each fix should be merge in 2 branches at least (master and snap).

didrocks:
Also, I'm maybe a little bit out of context, but I don't understand this:
- the badge was green in the past, then became red (but with this buggy green border that is defined in our css)
- your change is to make the color match the inner color of the badge.

Why is that not possible to push in the snap itself for bionic?

Feichtmeier:
Only for info real answer later because I'm on mobile
With cosmic there came a new version of dashtodock
Since that version the stylesheet.css is used
Before it was there too but it was not used

We cover all that styling in our shell theme so nothing was missed

didrocks:
No worry, take your time once you are back home! :)
I'm surprised part can be overriden (like the border-color) and not others (like background-color). Maybe, looking if another more precise property by stylesheet.css overrides our dock one would be a good lead…

Feichtmeier:
We could also do the following to not have to make new branches

Make background red too (until the fix is in Ubuntu dock in cosmic)
After make border and background green again

didrocks:
That would be my preferred approach.
We can then prepare another branch with the "proper" green color, ready to merge (once we diverge).

However, we really need to dig why one property is overriden, and not the other IMHO.

Feichtmeier:
Okay I will commit this then!

Because the border was not styled before we did it. Same as the box shadow :)

didrocks:
Ah, so everything that isn't specified can be overridden, but anything that is now specific in stylesheet.css from the dock can't?

Could be a Shell change as well to make our hack of "empty css file" not working…

didrocks:
LGTM, merging. Please reopen a new PR with the "don't merge yet" flag for reverting to green, and we really should investigate solutions as this will spread out more and more…

didrocks:
and thanks btw ;)

Feichtmeier:
Yes!

First our shell theme is used (and we style the dock in a separate file _dock.scss but meson etc.) 
Then the stylesheet.css is used and overrides everything that is styled inside this file. If it is not styled it is not overriden 

Cool thing that you can edit the file on mobile ^.^

didrocks:
Yeah, the order was the opposite in the previous release (if we put that empty file). But it was a hack anyway ;) Let's see where this lead us to.

Yeah, editing text file on mobile is really great. I have be witness of doing that as well :)

didrocks:
Ok, I have found the issue! As we changed session, the "mode name" changed as well in cosmic and so, the hack of empty file needs a new name :p

To trigger it, just `sudo touch /usr/share/gnome-shell/extensions/ubuntu-dock@ubuntu.com/ubuntu.css` (instead of existing yaru.css which is for the snap on bionic).

So, here, we have the override mecanism working (badges in green). However, the whole dock is now black as you mentioned when in maximized window mode. Do you mind looking at this?
I can SRU the empty file with correct name, but I think we should ensure Yaru is first updated to still have a Dock looking the same.
At least, progress!

clobrano:
Oh, I was checking that file, but didn't know the correct name. Let me close the bug upstream and then I'll look into this

clobrano:
It turned out that the transparency level difference between dock and panel had this same root cause.
However, there's a weird behavior in transition between opaque and transparent. It looks like the dock transitions first to a default transparency level and only then to our value. I'm about to open a PR about this

clobrano:
@Feichtmeier can I revert the badge color to green in this PR?

Feichtmeier:
Yes!!! XD

didrocks:
The new PR will only be included in d* though. I'll ensure as well that the ubuntu.css missing file is SRU in d*

clobrano:
What's "d*"?

didrocks:
next release-where-we-dont-have-a-codename-yet, but it will start with a d ;)

clobrano:
ah that's it :D

Feichtmeier:
Just curious: will the dash to dock fix of using the other empty CSS file be included in cosmic at some point with some point release?

didrocks:
I don't think there is a point to include the fix in cosmic:
- we see that there is some issues in the dock color, (@clobrano is looking at it)
- it thus requires for installing 2 packages from different sources at the same time. Not really SRU-friendly.
- we need to keep the red badge for cosmic anyway

So, sounds like better for targetting d* with this. Tell me if I missed anything obvious in that reasoning!

Feichtmeier:
Alrighty. We could keep a bionic branch with the green badges though?

didrocks:
I think we should do that:
1. Keep master with the red badge until d* opens (it's not opened yet, until it has a name)
2. Branch cosmic from master
3. Remerge green badge + add css file hack and release that on d*.

Does it make sense?

clobrano:
> we see that there is some issues in the dock color, (@clobrano is looking at it)

basically, fixing the empty css filename fixes also the badge color, but it has a side effect in the dock opaque color. However the fixes (css+dock color) are in Yaru only, what's the other package to be included?

didrocks:
> However the fixes (css+dock color) are in Yaru only, what's the other package to be included?
Oh sorry, I thought that the empty css file was in ubuntu-session (as it was in the past), but you're right. I only need to handle the package transition.

Feichtmeier:
:tada: nice! So it's us who have control over the ****.css file that is being created from the session name? @clobrano

clobrano:
@Feichtmeier, correct, it has been fixed by @didrocks already

PR #940, opened 2018-10-28
shell/common: fixed auth dialog title style

### LINKED ISSUE #937: Authentication dialog title uses Adwaita styling
## Expected Behavior

Authentication dialog title color should be using the same font color and font weight as other dialog titles, see an example in the shutdown dialog. #3B3B3B font color, font weight isn't bold:

shut

## Actual Behavior

Authentication dialog titles look like they do in Adwaita: lighter font color, bold font. This is probably the fallback styling, unless you explictly define how something looks for the Shell, the element always falls back to Adwaita styling.

aut

Open a program that requires authentication (gparted, etc.) or try to install a program from the software center

clobrano:
Made Gnome shell _authorization dialog_ title the same as the one in _close session dialog_

closes #937

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -255,7 +255,8 @@ StScrollBar {
   }
 
   .message-dialog-title {
-    font-weight: bold;
+    font-size: 135%;
+    font-weight: normal;
   }
 
   .message-dialog-subtitle {


clobrano:
Gtk and shell dialogs are not intended to be exactly the same

eaglersdeveloper:
Why?

clobrano:
Because they come from different sources. This has been discussed already
quite a lot. Last change was to move the buttons shape back to how they
look in adwaita.

Il giorno mer 31 ott 2018, 12:36 Eaglers <notifications@github.com> ha
scritto:

eaglersdeveloper:
If they come from different sources, this does not mean that they should be different.

clobrano:
Sure, but it means that the message has a different nature, gtk is application related, shell is system related. 
I think this is not the right place for this discussion though. This is a PR for a specific problem, which does not involve gtk. Please consider to open a new discussion on the HUB first

clobrano:
Good point, do you mind report it on the HUB to discuss with the others as well?

eaglersdeveloper:
Hub is discourse.ubuntu.com?

Feichtmeier:
We had the same discussion some months ago. Please stop to highjack our PRs with discussions @eaglersdeveloper . It's becoming a little bit annoying lately. 
You have a valid point, but this is THE WRONG PLACE

clobrano:
> Hub is discourse.ubuntu.com?

Yes

PR #941, opened 2018-10-28
shell/common: Made dialog background brighter

### LINKED ISSUE #938: Shell dialogs are different from MessageDialog
**We need to make Shell's dialogs similar to MessageDialog.**

This is GTK's MessageDialog:

This is Shell's PolicyKit dialog;

This is Shell's shutdown dialog:

Differences:

- [ ] **Button style**
- [x] **Background color**
MessageDialog has a dazzling white background (`#FFFFFF`)
Shell's dialog has an almost white background (`#F4F4F4`)
- [ ] **Headline**
In MessageDialog it is a bit bigger.
In shutdown dialog headline has normal font weight (in MessageDialog font weight is bold).

clobrano:
Use $light_base_color (#FFFFFF) in place of $light_bg_color (#FAFAFA)
keeping the same transparency level.

closes #938

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -782,7 +782,7 @@ StScrollBar {
 }
 
 %dialog_window {
-  $dialog_bg_color: transparentize($light_bg_color, $popover-alpha-value);
+  $dialog_bg_color: transparentize($light_base_color, $popover-alpha-value);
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr941`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr941` channel to test your snaps with those changes.

PR #942, opened 2018-10-30
Ensure min height for scale trough

### LINKED ISSUE #930: Yaru's theming for slider-widget doesn't draw a track at all when GTKAdjustment isn't provided, which makes it hard to interpret slider position in Firefox
## Expected Behavior
Yaru's slider-widget should have a clearly visible track, to make it clear what percentage the "thumb" represents.

## Actual Behavior
Yaru's themed slider-widget doesn't have a track visible at all -- it just shows the thumb, with no reference for how far it is in the range.

  1. View this URL in Firefox, in Ubuntu with Yaru theme: `data:text/html,<input type="range">`, or this JSFiddle: https://jsfiddle.net/e1s57u0z/

This just renders a detached floating white circle, without it being obvious that it's a slider on a track.
If I switch theme to e.g. "Adwaita", then I get a slider widget with a clearly-visible track.

## Specifications
Note: I didn't manually install Yaru -- it's just the theme that was there by default, in Ubuntu 18.10.  Here's what `$ snap info communitheme` says:
````
snap-id: Yd6CISPIf6tEf3ZEJ0cqSoEg9rG2VkRi
channels:                    
  stable:    0.1 (1273) 17MB -
  candidate: ↑               
  beta:      ↑               
  edge:      0.1 (1287) 17MB -
````

clobrano:
According to GtkScale documentation [1], upper and lower bounds can be the same and in this case the slider disappears. In Adwaita, only the trough is drawn, while  Yaru shows nothing, leaving the widget in an inconsistent state.

Ensuring a min-height of 2px fixes the above behavior without changing the aspect the scale has in the other use cases. 

closes #930

[1] https://developer.gnome.org/gtk3/stable/GtkScale.html#GtkScale.description


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr942`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr942` channel to test your snaps with those changes.

Feichtmeier:
This needs to be applied in the 3.20 folder now =D (as discussed)

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr942`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr942` channel to test your snaps with those changes.

PR #952, opened 2018-11-06
gtk: bundle CSS and assets into a gtk.gresource file

jhenstridge:
The PR bundles the PNG assets and generated CSS files into a `gtk.gresource` file for each version of the theme.  A stub `gtk.css` file then imports the bundled version of the CSS.

This optimises the process of loading the theme: rather than opening each individual asset, GTK will memory map the single `gtk.gresource` file, and then access resources via simple memory access (with the kernel faulting in pages if needed).

This does somewhat reduce the ability to poke around in the installed theme, but all of these bundled files are already build artefacts rather than source (either generated from an SVG asset sheet, or from SASS files).  It is also possible to unbundle the theme using the `gresource` command line tool if people really want to.


Feichtmeier:
Thanks for this PR !

> This does somewhat reduce the ability to poke around in the installed theme

Just for me, what do you mean when you write "poke around"?

jhenstridge:
> Just for me, what do you mean when you write "poke around"?

What I mean is people editing files under `/usr/share/themes` directly to experiment.  All the interesting data is bundled together into a single file.  But as what's installed isn't the preferred source (it's generated from sass and svg files), this might not be an issue.



> @jhenstridge, do we need to convert all the calls to specific files in the filesystem to calls in gresource format, like the one below?

No.  The main CSS is included in the bundle and `url()` does relative URI resolution, so there is no need to change those directives.

clobrano:
@jhenstridge, is pkg-config a dependency as well now?
The build failed with this message at the bottom

    Configuring gtk.gresource.xml using configuration
    Found Pkg-config: NO
    gtk/src/light/gtk-3.0/meson.build:202:0: ERROR: Pkg-config not found.

https://travis-ci.org/ubuntu/yaru/builds/456313012?utm_medium=notification&utm_source=email

clobrano:
And it is actually there in fact

```
Building theme 
meson --prefix=/ snapbuild
The Meson build system
Version: 0.45.1
...
Program sassc found: YES (/usr/bin/sassc)
Configuring index.theme using configuration
Program sassc found: YES (/usr/bin/sassc)
Configuring gtk.gresource.xml using configuration
Found pkg-config: /usr/bin/pkg-config (0.29.1)
Native dependency glib-2.0 found: YES 2.56.3
...
```

but this comes from "theme" part, while the error comes from a "yaru" part. Probably it's because `gtk-common-themes-parts.yaml` is missing the same dependency. I'm going to do a PR to test this

PR #961, opened 2018-11-09
Style the Shell's virtual keyboard to match the rest of the theme

godlyranchdressing:
Still a WIP

firstgo

Thoughts? @madsrh @clobrano @Feichtmeier 

Icons still need to be replaced, but I'm not sure which ones to go with yet. The language/region key especially.

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1844,12 +1844,15 @@ StScrollBar {
     &.default-key {
       border-color: $osd_borders_color;
       background-color: darken($dark_fg_color, 5%);
-      background-size: 20px;
+      background-size: 24px;
     }
     &.enter-key {
-      border-color: _border_color(#00A2E4);
-      background-color: #00A2E4;
+      $_blue: $green;
+      @include button(normal, $_blue);
+      // border-color: _border_color(#00A2E4);
+      // background-color: #00A2E4;
       background-image: url("key-enter.svg");
+      &:active { @include button(active, darken($_blue, 5%)); }
     }
     &.shift-key-lowercase {
       background-image: url("key-shift.svg");
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1847,12 +1847,11 @@ StScrollBar {
       background-size: 24px;
     }
     &.enter-key {
-      $_blue: $green;
-      @include button(normal, $_blue);
+      @include button(normal, $success_color);
       // border-color: _border_color(#00A2E4);
       // background-color: #00A2E4;
       background-image: url("key-enter.svg");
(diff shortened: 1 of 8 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr961`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr961` channel to test your snaps with those changes.

Feichtmeier:
Nice @godlyranchdressing !

Maybe this symbol? 
image

Also I'M not sure about the return key :thinking: It is very bright =)

godlyranchdressing:
I tried using that symbol, but the narrow lines within the globe ended up looking too blurry. I managed to find an Earth icon =though.

Second attempt with icons replaced:

second go

Darker blue:

darker blue

The blue here brings back that old complaint where it looks "unstyled" though, so here's one using green:

green

Feichtmeier:
I'd personally prefer the green one, awesome work!

I don't know about the light one, I think from the logic we applied to the OSD popups (volume/brightness), dark would be better, since the chance of having a bright screen in that area is rather high.

madsrh:
Nice work @godlyranchdressing 

I also prefer the green button, but I wonder what reason upstream has for coloring it in the first place.

godlyranchdressing:
@Feichtmeier I see. I just recall seeing a Unity 8 screenshot with the keyboard being white, I have no preference one way or another in this case.

@madsrh It makes it more obvious, I guess. Also, they could possibly be mimicking the virtual keyboard on other OSs. It's like that on Google Keyboard at least, so I'm sure there are others that follow.

eaglersdeveloper:
Try remove borders in keys and make keyboard not transparent.

Feichtmeier:
With the dark, transparent Jet and the border it fits 100% to our shell pop-ups and osd pop-ups so I would not try to diverge from this style. 
We have two shell styles: light is opaque and without borders and dark is transparent and with borders.
Choosing between those styles is the only option we have since a third style would really create the question if we have lost the control over consistency.

clobrano:
I finally got what you all where talking about :smile: , awesome work!

I noticed that the darker buttons, as well as the green one, do not have a hover effect, is that on purpose?

Feichtmeier:
Can't test the snap:
https://travis-ci.org/ubuntu/yaru/builds/453107045#L2381

godlyranchdressing:
@clobrano Yeah, it's like that in Adwaita so I just followed suit.

@Feichtmeier I was hoping it would fix itself eventually :laughing: but I guess not.

>fatal: not a git repository (or any parent up to mount point /home/travis/build/ubuntu)
>Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

I really don't know what to do with this, though. There's no conflicts either.

Feichtmeier:
Hmhm, now the merge into master commit fails on master, I guess I should revert this until some of you understood why it fails. We all like this change so it will land on master we only need to fix that CI build (maybe it is not even dangerous and I am panicking )

PR #962, opened 2018-11-11
Dependencies + Sudo to avoid a update-alternatives permission error

har9862:


clobrano:
Hi @har9862, thanks for contributing. 

I believe that not explicitly write the `sudo` part of a command is good practice, because it prevents people to just copy paste a the command, without noticing that they are acting like root, which is a good thing to notice instead.

har9862:
80cf43f 🤔

clobrano:
Good catch, this is something I would avoid, thanks

PR #967, opened 2018-11-15
snap: Removed version-script target from yaml

clobrano:
Removed `version-script` target to fix snap build failures

According to the discussion on snapcraft forum, yaml has either "version: git" or "version-script".

see: https://forum.snapcraft.io/t/how-we-disable-to-automatically-look-for-the-version-number-in-our-yaru-snap-build/8528


didrocks:
It seems to still fail though, no? See the Travis build output.

clobrano:
You're right, I got deceived by the other build :(

clobrano:
Alright, @kenvandine already pushed a fix in gtk-common-themes :tada:

didrocks:
\o/

PR #972, opened 2018-11-18
Added missing dependency in yaru part yaml

clobrano:
If the "theme" part yaml requires libglib2.0-dev dependency, the "yaru" part in gtk-common-themes-parts.yaml should need it too.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr972`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr972` channel to test your snaps with those changes.

PR #978, opened 2018-11-21
New desktop icons

ubuntujaggers:
Hi all,

I've updated my fork to make the following changes.

**The full colour user-desktop icon is now Yaru-looking:**

image

**It looks like all the symbolic icons are kept in a single master file, so I removed the launcher buttons from the desktop icon to make it match the full colour version:**

image

**Lastly, there's a file in my fork for user-desktop-symbolic by itself, so I removed the launcher buttons from that as well** (I wonder if this was necessary - perhaps a script extracts the individual symbolic icons from the master file?):

image

Let me know if there are any problems @Feichtmeier !


clobrano:
I does not seem to update the fullcolor icon. I see the svg of new one in src/fullcolor but not the PNG in Suru/ folder. Is it autogenerated or it must be done manually?

Feichtmeier:
Thanks @ubuntujaggers !

https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/README.md
Sam says that it works similar to the full colour icons. So the last commit was unnecessary I think. Could you revert this one?
Best regards and thanks again for your work!

Feichtmeier:
@clobrano I need to make a second PR with the PNGs extracted like we did it the last time Jaggers made that many icons

ubuntujaggers:
Think I fixed it...

clobrano:
It seemed that before the last commit the user-desktop-symbolic was similar to the fullcolor one, while now it is different

image

ubuntujaggers:
> It seemed that before the last commit the user-desktop-symbolic was similar to the fullcolor one, while now it is different

Hmm...  I definitely changed it in the source-symbolic file, you can see what it's meant to look like here: https://github.com/ubuntujaggers/yaru/blob/master/icons/src/scalable/source-symbolic.svg (right click to view full size in new tab - the user-desktop-symbolic icon is part of the second group on the top row).

The idea is that the individual icon for user-desktop-symbolic will be auto-generated from that source file somehow.  Is there still a step that @Feichtmeier has to do?  Otherwise, if we don't know how to force it to happen, I can update it manually again?

Feichtmeier:
No don't make it manual. 
Did you exchange the layers somehow? I think the script searches for the name, like with the full colour script

Feichtmeier:
Ah I know the problem

In the ruby script it looks up for inscape as a flatpak. I guess you dont use flatpaks, neither?
I make another PR and pull your changes, I think we can close this one.

PR #981, opened 2018-11-22
Opaque orange focus

Feichtmeier:
image
and
image

Instead of

image
and
image

@clobrano @madsrh


madsrh:
+1 from me. It makes sense to use the solid orange throughout the theme.

PR #988, opened 2018-11-24
Fix backdrop and checked:active

Feichtmeier:
Style backdrop and checked:active .path-bar buttons :)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr988`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr988` channel to test your snaps with those changes.

Feichtmeier:
Works
peek 2018-11-24 17-12

PR #971, opened 2018-11-17
Blue focus

Feichtmeier:
Testing PR for blue focus + solid focus ring + no margin
Basically everything, everywhere the same blue focus ring 

@matthewpaulthomas are you able to test this via snap or via the branch?

@madsrh @clobrano @godlyranchdressing @ubuntujaggers opinions, opinions, etc. :)

Never ending story :) 
Eventually closes https://github.com/ubuntu/yaru/issues/845
and 
closes https://github.com/ubuntu/yaru/issues/844

PR #992, opened 2018-11-30
Fix outline regressions introduced by PR #971

Feichtmeier:
Makes selection mode green button's outline color white
Makes **not** selected rows outline color orange


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr992`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr992` channel to test your snaps with those changes.

PR #997, opened 2018-11-30
Use background property to set insertion-cursor icon

### LINKED ISSUE #945: touch screen caret is stretched horizontally
## Expected Behavior

The caret shown in response to touch screen action in an entry widget should be displayed without distortion.

## Actual Behavior

The caret is stretched horizontally:

image

It seems that GTK expects the image asset used for the caret to have the same dimensions as the `text-select-start` and `text-select-end` assets (which is the case for Adwaita).  We're using the `slider-horz-scale-has-marks-above` asset as the caret, which is somewhat narrower.

  1. Run `GTK_DEBUG=touchscreen gtk3-widget-factory` (or any other GTK app) so the mouse is treated like touch screen input.
  1. Use the mouse to position the caret in one of the text entries with the overlay 

## Specifications

This is with `yaru-theme-gtk` version 18.10.6.

clobrano:
-gtk-icon-source uses the same properties (-GtkWidget-text-handle-width/height)
to set both text-selection icons and insertion-cursor icons, but since the two
kind of icons differ in size in Yaru, the latter looks stretched.

Using background-* CSS properties instead, to set the proper icon sizes.

closes #945

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -4561,7 +4561,11 @@ cursor-handle {
 
     &.insertion-cursor#{$s}:dir(ltr), &.insertion-cursor#{$s}:dir(rtl) {
       $_url: 'assets/slider-horz-scale-has-marks-above#{$as}#{$asset_suffix}';
-      -gtk-icon-source: -gtk-scaled(url('#{$_url}.png'),
+      -gtk-icon-source: none;
+      background-repeat: no-repeat;
+      background-size: 16px 24px;
+      background-position: top 2px;
+      background-image: -gtk-scaled(url('#{$_url}.png'),
                                     url('#{$_url}@2.png'));
     }
   }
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr997`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr997` channel to test your snaps with those changes.

Feichtmeier:
image
 
I think it works =)

clobrano:
It looks a bit misaligned

Il giorno sab 1 dic 2018, 11:12 Feichtmeier <notifications@github.com> ha
scritto:

clobrano:
caret

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr997`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr997` channel to test your snaps with those changes.

PR #998, opened 2018-11-30
Fixing entries to have a blue/dark blue selection for light/dark entries

Feichtmeier:
Regression introduced by #971 

image
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr998`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr998` channel to test your snaps with those changes.

PR #1007, opened 2018-12-02
Repair master

Feichtmeier:
Reverts my bad symlink commit ( https://github.com/ubuntu/yaru/commit/cb4c489b03c5259cbd80dfa78f5e6699be369fff ) which broke the local meson build
Also commits the symlinks correctly

### DIFF
--- a/icons/Suru/16x16/apps/update-notifier.png
+++ b/icons/Suru/16x16/apps/update-notifier.png
@@ -1,1 +1,0 @@
-aptdaemon-download.png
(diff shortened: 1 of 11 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1007`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1007` channel to test your snaps with those changes.

PR #1012, opened 2018-12-03
Firefox theming

Feichtmeier:
- remove the dedicated text selection color, because it also changes the whole accent color of firefox and thundebrird to blue
- remove the border-radius of menus without removing the shadow
- add comment reminders to remove the dedicated styling when the fixes will happen

@madsrh @clobrano 
are you okay with the first point? Thunderbird and firefox are completely blue now, since the mozilla window sucks up one color and takes it as the accent color :cry:

### DIFF
--- a/gtk/src/light/gtk-3.20/_apps.scss
+++ b/gtk/src/light/gtk-3.20/_apps.scss
@@ -531,21 +531,15 @@ headerbar button image ~ window decoration ~ menu separator {
  * Firefox *
  ***********/
 #MozillaGtkWidget.background  {
-  > widget {
-    > scrolledwindow > textview {
-      text {
-        &:selected { @extend %selected_text; border-color: $selected_bg_color;}
-      }
-    }
-  }
-
   headerbar.titlebar {
-    // Remove the round corners until firefox fixed the white border bleed at the top edges
+    // Removes the round corners until firefox fixed the white border bleed at the top edges
+    // REMOVE THIS when firefox 65 is in the ubuntu repos on the end of Januar 2019
     border-radius: 0;
   }
 
-  > menu, .menu { border: none; } // Removed menu borders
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1012`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1012` channel to test your snaps with those changes.

clobrano:
Is the firefox/thunderbird change for the download section only? If that so, we are going to have a selected download row all orange, instead that all blue, is that right?

image

Feichtmeier:
No that's for Firefox and Thunderbird using light blue as the accent colour.
I first thought I can fix the progress bar also but that's a Firefox issue, no matter which theme you use. I opened a bug on bugzilla for this.

clobrano:
> No that's for Firefox and Thunderbird using light blue as the accent colour (everywhere).

Yes, sorry I was unclear :D 
I meant, even if we change the selection color in firefox, we cannot get rid of this issue, can we?

If that so, having it blue or orange does not make any difference here, however the text selection would be orange, while in the rest of the theme is blue. Or did I miss something else?

clobrano:
I would keep it blue, at least text selection would be consistent. What do you think?

Feichtmeier:
Personally I think that we should not fix the issues of Firefox theming. The orange tab border was also quiet iconic for Ubuntu users. We exchange wrong theming in the text for wrong theming in the whole theme.
But on the other ✋ I am not a Firefox user so I don't have a strong opinion on that!

clobrano:
> Personally I think that we should not fix the issues of Firefox theming. The orange tab border was also quiet iconic for Ubuntu users. We exchange wrong theming in the text for wrong theming in the whole theme.
> But on the other  I am not a Firefox user so I don't have a strong opinion on that!

good point, let's keep it as it is in this PR

PR #1019, opened 2018-12-05
Add icon superimposing Firefox on a squircle

ubuntujaggers:
Hi all,

I don't think we ever made a firm decision on which Firefox to use, so I've done the least controversial one, which is the Firefox logo on a plain white squircle.  Let me know if you'd rather have a different one, happy to do more work on it.

image


Feichtmeier:
Nice! Can you also add Thunderbird? I think then we have them all :) 
In a future PR
I like this, let's wait for @madsrh and @clobrano

ubuntujaggers:
> Can you also add Thunderbird? I think then we have them all :)
> In a future PR
> I like this, let's wait for @madsrh and @clobrano 

Yep, no probs - whatever gets accepted/merged for Firefox, I'll do Thunderbird the same way  :)

clobrano:
It looks good :+1:

mivoligo:
I've just started Ubuntu and was thinking it's a bug. It looks kind of out of place in my dock next to other icons.
zrzut ekranu z 2018-12-06 13-04-29

clobrano:
Maybe it's also because being the squircle white the normal light effects
of the borders aren't visible

ubuntujaggers:
I think he's right, it looks a bit rubbish in that screenie. Let me try one of the colours when I get home shortly.

ubuntujaggers:
Do either of these look better:

ff pale blue

ff pale yellow

Both squircles are colours from the Suru palette.  We have to get this one right because it's Firefox, so I'm happy to do more work.

clobrano:
I prefer the second, which is close to the auto generated one

Il giorno gio 6 dic 2018, 18:50 ubuntujaggers <notifications@github.com> ha
scritto:

ubuntujaggers:
Yes, the second colour is a bit more neutral without being white or grey.  Happy to change/commit/push if others feel the same.

madsrh:
@ubuntujaggers could you post a mockup/screenshot with the icon in the dock?

ubuntujaggers:
image

Not sure I've quite nailed it...

dedricbrayon:
Android:
screenshot_20181110-121825_samsung experience home

iOS:
1200x630wa 1 1

Firefox about dialog:
screenshot from 2018-12-06 20-01-57

Feichtmeier:
I think adapting to iOS/ANdroid is a very good idea

clobrano:
I agree. That dark blue is beautiful

Il giorno gio 6 dic 2018, 20:10 Feichtmeier <notifications@github.com> ha
scritto:

ubuntujaggers:
If we go for the Android/iOS blue (which has a _lot_ of advantages in terms of respecting the brand) it might be an idea to soften the Suru highlight so it isn't so jarring on a dark squircle.

Also I had another go at the orange background, both are shown here:

image

ubuntujaggers:
@feichtmeier dark one with softer highlight?

clobrano:
>@c-lobrano, you felt it was so dark that it looked like a normal Firefox icon, I assumed that was a negative but actually I wonder if I misinterpreted you? i.e., did you like this version?

I forgot :D, you're right

ubuntujaggers:
image
?

Feichtmeier:
I would prefer the dark purple background overall other because
1) it is the same on iOS/Android
2) it has a great contrast against a) blue and b)orange in the firefox logo

madsrh:
I agree with @Feichtmeier +1 dark purple background works best IMO

ubuntujaggers:
Super - @Feichtmeier, @madsrh, @clobrano, how about this for the final version:

final dark blue ff

image

ubuntujaggers:
Okay, I pushed the dark blue version  :)

ubuntujaggers:
@Feichtmeier, do I do a new PR now or do you just merge the dark blue one?

Feichtmeier:
Yes you need to make a new Pull Request :)

ubuntujaggers:
Done! https://github.com/ubuntu/yaru/pull/1030

PR #1029, opened 2018-12-09
Remove red/yellow border of error/warning entries when ...

Feichtmeier:
.. out of focus or in backdrop

Before:
peek 2018-12-09 17-29
After:
peek 2018-12-09 17-31


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1029`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1029` channel to test your snaps with those changes.

PR #1030, opened 2018-12-09
Add dark blue Firefox icon aligning with Android and iOS icons

ubuntujaggers:
After discussion, here's the chosen Firefox icon to replace the version with the white squircle:

ff all

image


clobrano:
Thanks @ubuntujaggers, merging it :tada:

Feichtmeier:
@ubuntujaggers  could you make a new PR for Thunderbird?  :bird:

ubuntujaggers:
No probs, should I use the same dark blue again?  I have a lot of time tomorrow day so am going to look at Thunderbird and backup file.

Feichtmeier:
No, they are prbly placed as the first two apps - so a different color would be cool
Thanks for your work!

madsrh:
@ubuntujaggers wrote:

> I haven't been able to use it live yet - I can't get snaps to work since I upgraded to Cosmic, so I can't comment on this one (I might try replacing the icons manually).
>
> IMO Firefox was a hard icon to do because our hands were tied in terms of altering the colours and design even slightly to adapt them to the theme. So it was literally a case of finding the least objectionable colour for the squircle, and even that was hard because Firefox has quite lurid colours and the Ubuntu dock is dark.

I'll comment here rather than derailing the other thread or creating an issue if this isn't an issue for anyone but me 😉 

As mentioned I **LOVE** [the image above](https://user-images.githubusercontent.com/38893390/49700459-b438e200-fbd6-11e8-92e7-ad4725b53344.png)! It's SO beautiful as 256px BUT in the small sizes the blue looks almost black and it's a very "heavy"/dark look compared to the other Suru icons 🏋️ 
I think "System monitor" has the darkest background of all the suru icons, and it feels lighter that Firefox.

image

ubuntujaggers:
@madsrh Yep, I agree  :)  I think the essential problem is, we identified one colour that definitely went well with the icon itself - and was used in other OSes - which arguably doesn't sit 100% well with the rest of Suru or Yaru because it's so dark.  I think it was the least unpopular option rather than one with unanimous support.

_"As mentioned I LOVE the image above! It's SO beautiful as 256px BUT in the small sizes the blue looks almost black and it's a very "heavy"/dark look compared to the other Suru icons"_

I remember reading something about the New Coke fiasco in the 1980s...  apparently one mistake they made was relying too much on the "sip test" in their market research.  This is where you give someone a sip of your drink and ask what they think.  The problem is, sometimes a sweeter formula makes a great first impression and then gets sickly.  So, they realised that you have to give your testers a crate of the stuff to have in the fridge and drink throughout the week, to find out how nice it is in the long run, whether it goes well with meals, etc..

I think the dark blue squircle definitely passes a visual "sip test" because it's a lovely shade of blue and the source file for the icons looks fine when you view it at full size.  But you have to see it on the desktop for a while and click on it every day before you know for sure if you like it.  So, if testers have been living with it for a few weeks, it's probably worth double-checking that people are still happy with it, and plenty of time before 19.04 to try lighter shades if not  :)

clobrano:
I agree with Mads. I used the surufy icon (from the script) for some time
and when changed to the new dark icon I felt the difference.
As soon as possible, I can show here the surufied icon, if you like

ubuntujaggers:
The dark blue wasn't my favourite either. If the Surufy icon looks nice, we could just use the pipette tool to take the colours...?

ubuntujaggers:
I can make and push an svg with that exact gradient very quickly if you, @madsrh and @feichtmeier prefer it?

clobrano:
The log message with colors for background and shadow might be helpful

generating firefox.png with: size 32x32, background #F0BF25, shadow #E74530

madsrh:
I like the scripted background. Could you also try a brighter blue? As we saw with the current, we should test it in the real 😄 

It seems like the Firefox logo is bigger in the scripted version compared to the Jaggers.

Here's a pretty bad mockup:
untitled-1

Feichtmeier:
I'm fine with the golden Firefox :)

clobrano:
> It seems like the Firefox logo is bigger in the scripted version compared to the Jaggers.

I didn't notice, nice catch. I'd say I prefer the bigger logo, what do you think?

ubuntujaggers:
I only made the logo smaller because the blue was so dark, to give everything more room to "breathe". We were using a larger size with the lighter backgrounds :) so can fall back to that.

ubuntujaggers:
I'll post some options for this tonight.

ubuntujaggers:
Option 1: here's the svg with the script colours.

firefox with script colours

The Firefox logo is enlarged as per discussion - when I push the chosen option, don't let me forget to do a PR for Thunderbird enlarged by the same amount!

Feichtmeier:
I like the gold a lot

I wonder if we could use the new Firefox icon already?

ubuntujaggers:
That's the icon that was launched in December 2017 - has there been another one since?  If so I can swap it in.

Here's a version with colours from @madsrh's mockup:

firefox with madsrh colours

Feichtmeier:
firefox-logo-design-system-1-1
One of these will be the new icon. Don't know when though. Yes let's wait then

madsrh:
I like both but do prefer the blue version. To me, the yellow background is a bit too colourful.
image

ubuntujaggers:
Option 3: a lighter blue that isn't as vivid:

firefox with less vivid blue

image

clobrano:
The effect of blue squircle is very nice, however, as with the dark blue, doesn't make more evident that the firefox icon is smaller?

ubuntujaggers:
I see what you mean @clobrano, the golden background has at least one advantage of making the icon seem more expansive.

@madsrh, asap I'll mock up an option 4 which is the golden colour but maybe with a tiny bit less saturation, just to see if that works.

Also I'll look up the exact apps/order of the default launcher to remind myself...  I suppose that's the most important context for the icon to be perfect, so really I should mock these up in a default launcher, rather than with any old Suru app icons around them.

ubuntujaggers:
Here's a slightly less vivid gold - I still intend to mock up a proper default launcher, but I have to resize Thunderbird too:

image

ubuntujaggers:
I think all the ones we've discussed in this closed PR would be better than the current too-dark one - which should I commit?  I like the script colours but very slightly desaturated (last one above).

clobrano:
+1 for the last one

Feichtmeier:
I think the most thumbs where with that light blue. Let's do that :) already so much time spent on that icon :D

PR #1037, opened 2018-12-12
Removes firefox headerbar bleed fix because ....

Feichtmeier:
... Firefox 64 landed in Ubuntu and fixed the headerbar border-radius bleed

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1037`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1037` channel to test your snaps with those changes.

PR #1038, opened 2018-12-12
Network-server, folder-remote & Thunderbird icons

ubuntujaggers:
Hi all,

**EDIT:** Can't change the title it seems!  Can a maintainer do it?

Let me know if there are any problems:

**network-server**

image

**thunderbird**

image


Feichtmeier:
It works!
image

The network share symbolic is also being used, I think you are on the right way.
image
There is something missing for the network share folder though

Personally I don't have a problem with you making all the icons in this PR, but I don't know how @clobrano feels about it

ubuntujaggers:
Hmm, the icon I added to symbolic is actually the bottom right one in places, the ethernet port.  I called it "folder-network" because that seemed consistent with how the others were named.

Let me look again and try renaming it.

ubuntujaggers:
@Feichtmeier, see if the network glyph now appears on the folder?

clobrano:
> Personally I don't have a problem with you making all the icons in this PR, but I don't know how @clobrano feels about it

sorry, I've missed that. I'm ok if the icons all belong to the same topic. Eventually, just change PR name :)

Feichtmeier:
You need to add another folder (+layers) to this file
screenshot from 2018-12-13 11-11-59

On that folder you would need to put your new glyph
and name the layer/folder
folder-remote

But I can do that if you like now that I have your glpyh, are you finished with the glyph?

Feichtmeier:
49931790-ee9cca80-fec7-11e8-925e-e9dcf1422b6c

Feichtmeier:
I like the level of sharpness and finish :dagger: 

However, I don't know if the abstraction of the RJ45 plug is recognized as such :thinking: 
In the Humanity icon there was a cabel and everything is so super 3D that you get the icon anyways :D

ubuntujaggers:
I wonder if the problem is that the glyph was designed as dark-on-light, so this is an inversion.  Making the socket light on the side of the folder makes it harder to recognise as a socket, rather than an object of that shape.

How about if the socket is a hole in the white area?

image

Feichtmeier:
That looks nice! What do @madsrh @clobrano @godlyranchdressing think?

ubuntujaggers:
Hi all,

I've pushed Thunderbird too.

madsrh:
Nice work again @ubuntujaggers

clobrano:
Last one is more recognizable, good work ;)

Feichtmeier:
Don't forget to commit folder-remote 👼

ubuntujaggers:
Will do as soon as I'm home/with my PC this evening! :)

clobrano:
I know I'm nitpiking here, but please next time do not mix way too different icons in the same PR :)

ubuntujaggers:
Oh, apologies, I misread your post - I re-read it and understand it now.

clobrano:
No worries, it's just to keep things in order :)

ubuntujaggers:
Fixed it!

Feichtmeier:
Don't hurry - alternatively: send me the .svg's you edited and I make a new branch

Feichtmeier:
Maybe it would be nice at this point to use some UI for your commits. Like for example gitkraken or vscode :)
Then you see exactly what your unstaged changes are and what you added to your stage/index. So it is a bit more in front of your eyes what you are going to commit
You can also revert the stuff "faster"/lazier

ubuntujaggers:
That should be done now!

ubuntujaggers:
Magic! If you pin one of the remote folders, does the correct glyph appear by it on the side bar?

clobrano:
> The script sees the true size of the baseplate rectangle in the XML and helpfully makes new icon folders with names like

This looks like a bug in the script. I look at it if it can be fixed

Feichtmeier:
> Magic! If you pin one of the remote folders, does the correct glyph appear by it on the side bar?

peek 2018-12-15 10-58

Is that the glpyh you expected or the R45plug?

clobrano:
> > Magic! If you pin one of the remote folders, does the correct glyph appear by it on the side bar?
> 
> peek 2018-12-15 10-58
> 
> Is that the glpyh you expected or the R45plug?

I like the glyph showed here, but we received already complaints about the difference between glyph used in the sidebar and in the icons :smile:

Feichtmeier:
The second "Ya" is just a bookmark, I think that's okay

But yeah... RJ45 or not? I wonder if it isn't better this way for the symbolic icon only
The folder is perfect!

ubuntujaggers:
My intention was for the same glyph to appear in the sidebar, not sure why it isn't, happy to investigate if we want the consistency... @c-lobrano, really it's Inkscape's fault for snapping to non-integers :) but it would certainly be good if the script could handle that more sensibly.

Feichtmeier:
Hm, I am not sure if the R-45 glyph works in the sidebar

The network folder symbol works because it is already a whole symbol
For example when I connect to my webdav (t-online crap) cloud which I have bookmarked it looks like this:

peek 2018-12-15 15-44

clobrano:
I am the only one who think the glyph should be the same? It's an hook for the user to understand the connection between sidebar and icon

Feichtmeier:
[img](https://imgflip.com/i/2p1ntk) ;D Sorry I had to

Okay, but do you think that the RJ45 plug alone would be recognized as the network folder?
Or would you rather but the symbolic folder onto a folder? :D That's like "folder-ception"

clobrano:
You're welcome :p 

I like Adwaita approach

image

not a folder, but a server, that remembers also the proper server icon

Feichtmeier:
I am not sure about this

The remote "connection" / mount could be a server OR a network folder mounted via NFS / samba / webdav
But only the webdav is really a "server" connection

I would rather go for the network folder currently used in the sidebar, but I agree that the folder-remote should remind of the glyph

clobrano:
Sure, it's not really a server, but it's not a RJ45 connector as well :D, in any case it's just a representation

Feichtmeier:
There is one additional problem I see with the RJ45 plug: kids these days prbly don't even know what it is :laughing: 
I really like the general idea @ubuntujaggers but I agree that carlo has a point here: the symbolic icon should remind of the fullcolor one
Do you have any other idea? Or @madsrh ?

ubuntujaggers:
Happy to draw whatever - I only went with ethernet because of the screenshot, I don't actually know how these icons are used. I have a laptop and email myself files if I need them somewhere else, *blush*.

We could take the sidebar glyph from @feichtmeier 's gif and put it on the folder too, except for the fact that it's a little folder, and a folder on a folder might look odd? Should I try it?

Alternatively: the ethernet port is already contained in a rectircle. Perhaps for the sidebar version we could replace the rectircle with the little folder. However, that doesn't address concerns about the symbolism.

ubuntujaggers:
Hmm, last suggestion is harder than it sounds:

image

ubuntujaggers:
I do agree that the sidebar glyph should match the one on the folder icon  :)  So, couple of options IMO:

- Draw a new glyph.  Like I say I'm more than happy.  The folder icon does look nice, but really, any glyph we like will also work well on the folder.  The only problem is I'm scratching my head a bit for inspiration?

- Put the sidebar glyph as-is on the folder.  That would indeed be folder-ception  :)  but it would be consistent.

- Put the ethernet port on the sidebar.  This wouldn't necessarily be very intuitive, if it only works on the folder because it combines the port with the folder.  Also, I note @feichtmeier's point that ethernet ports are a bit dated (similar to using a floppy for "save").

- Make a new sidebar icon which is a miniature of the folder icon (so you have an ethernet port on a folder).  This is a little difficult because of the small number of pixels.  I had a go up there where I made the folder slightly more "closed" so I had more room inside the boundary.  I don't mind this version tbh.  Another option I tried was making the folder slightly taller, because we've got 16x16 pixels but the normal folder glyph is only 14px tall.  I feel logically that this must be "legal" because the folder in the current sidebar icon, with the line under it, is 13px tall to make room for the line.  So not all the folders are exactly the same size in the glyphs anyway.  But I tried a 16px tall folder and it just doesn't look quite right with the others.

The third thing I tried tonight was similar to the search results glyph:

image

...where the port glyph overlaps but isn't bounded by the folder glyph.  IMO that didn't work, it's important for the port to be fully on/in something so it looks like a opening rather than an object.  So, long story short: if we want a ethernet port on a tiny folder, I think it would probably have to be very similar to the mockup I already posted.  Also, same objection here as previous bullet.  Ethernet port just might be the wrong symbolism?

Feichtmeier:
@ubuntujaggers about your inspiration.... xD Maybe something "like this" (ultra crap mock-up incoming but I hope you get what I wanted to "draw"):

image

ubuntujaggers:
Yes, good idea - let's try doing the sidebar glyph but with the wire outside the folder :)

Feichtmeier:
@clobrano 

https://gitlab.gnome.org/GNOME/adwaita-icon-theme/blob/master/src/fullcolor/folders.svg
image

https://gitlab.gnome.org/GNOME/adwaita-icon-theme/blob/master/src/symbolic/gnome-stencils.svg
image

That would be the same idea like my "mock-up" + the current
image

clobrano:
Oh, ok then :D

ubuntujaggers:
First attempt at 48px:

image

??

As with Adwaita and the sidebar icon, the compromise here is that the network folder can't be as tall as the other icons, because you need to make a little room underneath it.  What do you think?

Feichtmeier:
Nice! I think that's what they also did with adwaita
And it looks good imho

Feichtmeier:
But don't let the plug flow into the icon. Or? Hm

clobrano:
Nice indeed :)

ubuntujaggers:
image

Or with the cable behind the folders:

image

clobrano:
Definitely better the latter to me

Feichtmeier:
Yes the second

ubuntujaggers:
All right, I'll push it...

ubuntujaggers:
Pushed, how does it look?  :)

Feichtmeier:
image

Awesome!

Feichtmeier:
Three thumbs - I merge this!

PR #1043, opened 2018-12-14
Make focus border for entries always the same color in dark and light

Feichtmeier:
@clobrano @madsrh @godlyranchdressing please test and say if you like it/agree

Imho that dark orange looks strange in the dark theme
That function may work in adwaita with their blue, but it looks washed out in our color theme (imho)

Makes not much sense to provide screenshots as PEEK always changes the color a tiny bit :man_shrugging:

### DIFF
--- a/gtk/src/light/gtk-3.20/_drawing.scss
+++ b/gtk/src/light/gtk-3.20/_drawing.scss
@@ -34,8 +34,7 @@
  * Entries *
  ***********/
 @function entry_focus_border($fc:$focus_color) {
-  @if $variant == 'light' { @return $fc; }
-  @else { @return if($fc==$focus_color, _border_color($focus_color), $fc); }
+  @return $fc;
 }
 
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@clobrano if you agree we could also remove all the occurences in common, but if you want to keep it for a bit "less upstream diff" I understand

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1043`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1043` channel to test your snaps with those changes.

clobrano:
Looks good, I just need to get used to it, since a color so bright in dark variant makes the border pops out a lot more

madsrh:
image

image

I didn't noticed that these weren't the same color, but now that you mention it, it's pretty easy to see the difference. 
> Imho that dark orange looks strange in the dark theme

I find the contrast when using the "full orange" a bit eye jaring at first, but it makes sense that these should be the same color 👍  As a famous software engineer once said:

> ...a color so bright in dark variant makes the border pops out a lot more

PR #1055, opened 2018-12-20
Remove headerbar and menubar side borders

### LINKED ISSUE #1049: window and headerbar border problems
image

image_2018-12-17_11-43-45

image

The headerbars, titlebars and menus have a border but the rest of the window does not have a border or a visible border-color.
It is more visible in the backdrop and even more visible in the backdrop of the darktheme but it is present always.

It creates this little glitch where 1 pixel is ahead of he other so I think the border is there but it is just transparent (didn't look it up yet).

Adwaita looks like this (just for reference):

image

image

clobrano:
Set headerbar's left and right borders width to 0px, as windows' side borders are.

close #1049 

NOTE: with this change, we will lose some distinction between the edges of the headerbar and dark backgrounds.

before
image

after
image

For reference, this is adding a border color (inkstone) to the window view

image

from my point of view, the result depends too much on the background (it's invisible in white bg, and looks bright on the wallpaper) and on the actual window view color

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1593,7 +1593,8 @@ headerbar {
   border-color: transparentize(_border_color($headerbar_bg_color), 0.7);
   border-top-color: $inkstone;
   border-style: solid;
-  border-width: 1px 1px 0 1px;
+  border-width: 0;
+  border-top-width: 1px;
   border-radius: 0;
   background-color: $headerbar_bg_color;
   color: $headerbar_fg_color;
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1055`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1055` channel to test your snaps with those changes.

Feichtmeier:
Yes this looks a lot better imho
image
image
image

clobrano:
Alright, we can improve later the distinction between overlapping windows with box-shadows. It is something I wanted to do since the beginning, but gtk2 does not support it

Issue #1025, opened 2018-12-07
[Missing icon] Backup file


vitalkanev:
MIME Type: `application/x-trash`

```
$ git describe && lsb_release -rci
18.10.7-180-g038e0c1a
Distributor ID:	Ubuntu
Release:	18.04
Codename:	bionic
```
![2018-12-07 17-53-17](https://user-images.githubusercontent.com/18398761/49663567-30f58000-fa60-11e8-840d-df11ff15dbb6.png)

Feichtmeier:
@ubuntujaggers @eaglersdeveloper :D

clobrano:
This is Awdaita

![screenshot from 2018-12-08 11-12-44](https://user-images.githubusercontent.com/2883614/49684885-545d1100-fada-11e8-99a6-e0fc04409b7f.png)

eaglersdeveloper:
https://github.com/snwh/suru-icon-theme/issues/110

meetdilip:
Do you guys like something like this instead of a plain text icon ?

https://www.iconfinder.com/icons/383184/backup_icon

ubuntujaggers:
Is the answer just to clone the normal (Suru) icon to this position, like Adwaita does?  I'm not sure I understand the symbolic logic of the one on Filefinder...?

Feichtmeier:
Hmmm maybe the same icon but with @meetdilip 's glyph

clobrano:
Sorry @meetdilip, but I don't think that icon would be nice in suru. Is there any example from most used icons set?

clobrano:
Consider that this files are usually hidden. I'd say that we can even keep it as it is

Feichtmeier:
I think he means only the shape as a glyph

Like this icon but instead of the (C) that backup icon but in "suru" style (thinner)
![image](https://user-images.githubusercontent.com/15329494/49691975-c2dbb680-fb4e-11e8-9124-f15dc8aa5a27.png)

meetdilip:
I meant the above. The file background and a time indicator glyph instead of the copyright symbol in the icon by @Feichtmeier 

@clobrano Attached what Papirus does

[application-x-trash.zip](https://github.com/ubuntu/yaru/files/2660184/application-x-trash.zip)

meetdilip:
I thought the icon should say that it is a document, from an older time. Here is a mockup

[...]

PR #1061, opened 2018-12-24
Adding x-trash.svg ("backup file" of #1025) plus .pngs of different sizes

ubuntujaggers:
This is the icon that's the subject of #1025, "Backup file."  However, I note that the icon itself is called "x-trash", so maybe @eaglersdeveloper 's idea of a translucent file icon is intuitive after all.  Here's what I've got:

image

Wasn't sure if it should be called "application-x-trash" or "x-trash".  I went for the latter so let me know if it's wrong!


clobrano:
> However, I note that the icon itself is called "x-trash", so maybe @eaglersdeveloper 's idea of a translucent file icon is intuitive after all

The name seems to the misleading. According to MIME type database, it is defined as backup file only
http://mimeapplication.net/x-trash

Curiously, Elementary Icon has a ticket open to include this icon as well and they proposed an actual trash/recycle image
https://github.com/elementary/icons/issues/703

clobrano:
Also, it looks like the name `x-trash` is fine

image

ubuntujaggers:
I recently released that there's already a Suru icon which uses a clock glyph - "image-loading", as follows:

image

IMO if we want to use a turning-back-the-clock glyph for application-x-trash then we need to re-use the existing clock rather than drawing a different one from scratch.  Here's what I've got so far, I'll submit a PR when Polari is put to bed:

application-x-trash

clobrano:
Totally agree

PR #1064, opened 2018-12-28
Fix gtk2 treeview header double lines

### LINKED ISSUE #908: Gtk2 treeview, header and scrollbar double lines
Good but half solution .  When you look at the top headerbar double-line is drown -one for header and one for the window frame. on the right top side, the last separator is just there hanging and the top line above the scrollbar is thinner . that way it looks unfinished imo.
drawisland
Another thing I think did is to remove the handle line the is way too much here, imo again :smile: 
drawisland-1

I did it by going into the main.rc file and changed the value from 9 to 2. 
 GtkPaned::handle-size = 2
  GtkHPaned::handle-size = 2
  GtkVPaned::handle-size = 2

_Originally posted by @Paz-it in https://github.com/ubuntu/yaru/pull/878#issuecomment-425882843_

clobrano:
GTK2 treeview header has its own top, right and bottom borders,
however most of the time this widget is enclosed into a frame, resulting
in double borders on top and right.
Right border cannot be removed, since it is also the separator between
several headers on the samer raw, however the double line effect is
reduced replacing the pixmap engine `function: BOX` with `function: SHADOW`
when defining `style "treeview_header_button"`.
The top border is always doubled, and could be safely removed.

see https://github.com/ubuntu/yaru/issues/908#issuecomment-450353718 for more information and screenshots

closes #908


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1064`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1064` channel to test your snaps with those changes.

Paz-it:
Tested!  - Very wel done :clap: :clap: :clap: l

clobrano:
Wait before merging this. There's an issue with gtk-common

clobrano:
all fixed, merging it

Issue #1073, opened 2019-01-01
insert-text and edit-select-all are hard to distinguish 


Feichtmeier:
![image](https://user-images.githubusercontent.com/15329494/50574430-e3fb9400-0de8-11e9-882f-a109e7160e4e.png)

![image](https://user-images.githubusercontent.com/15329494/50574445-02fa2600-0de9-11e9-8ccb-0c9ed06fdeb5.png)

![image](https://user-images.githubusercontent.com/15329494/50574446-0d1c2480-0de9-11e9-9274-cf4733676f90.png)

@ubuntujaggers could you take a look at edit-select-all ? I think adwaita got it a bit more right, maybe we can just copy this one and make it thinner? Or just copy it I don't know
![image](https://user-images.githubusercontent.com/15329494/50574455-45236780-0de9-11e9-91ca-4ba000d5e607.png)
https://gitlab.gnome.org/GNOME/adwaita-icon-theme/blob/master/src/symbolic/gnome-stencils.svg

ubuntujaggers:
I started off trying to clone the Adwaita one with a squircle boundary instead of a hard square.  16x16px is an even number which doesn't lend itself well to central dividers (there are two in the Adwaita design) because it means they have to be 2px wide (which looks too big), or not quite central.

Looking at Adwaita, they've got round this by making the whole icon slightly off-centre, which I didn't know was allowed.  So the icon is 16x16px but the actual graphic fills 13x13px (the following shows the graphic relative to the icon boundary):

![image](https://user-images.githubusercontent.com/38893390/50588500-4badef80-0e7a-11e9-8786-0e5f4cf8fae6.png)

Before I realised what Adwaita was doing I tried a different glyph to represent "everything" that didn't use central divisions:

![image](https://user-images.githubusercontent.com/38893390/50588643-e8708d00-0e7a-11e9-82a3-6cb33f78d294.png)

![image](https://user-images.githubusercontent.com/38893390/50588670-050cc500-0e7b-11e9-8414-d723d39a049a.png)

I actually quite like this, because having a stack of two things to represent "all" resembles other action icons like copy, paste, open.  However, is it intuitive (well - at least as intuitive as the Adwaita icon)?

So, two options;

- Use the above "stack" glyph
- Copy Adwaita and also have a slightly off-centre graphic that includes 1px central divisions in the glyph

I tried a third option ("Copy Adwaita and have 2px central divisions") but it didn't look good at all IMO.  I will do some mockups of the other two options.

[...]

PR #1077, opened 2019-01-02
Updated source-symbolic.svg including less ambiguous "select all" action icon

ubuntujaggers:
I went for the new glyph because I thought it went quite well with the others:

image

image

...but happy to change if others preferred the second option  :)


clobrano:
@eaglersdeveloper, there is no need to compete on a task, let's discuss the solution and then act, please

Feichtmeier:
Maybe better take this
https://github.com/ubuntu/yaru/issues/1073#issuecomment-450836296
Since the one shown in your opening post looks too similar to copy like carlo said

ubuntujaggers:
@clobrano It's certainly similar to the icon for copy - another possibility is the Adwaita design adapted to Suru shapes:

**Option 2**

image

image

...or we can swap in an undivided square to stand for "all" as per Eagler's svg:

**Option 3**

image

image

...or we can go for the simple square design but use the full 16x16px area:

**Option 4**

image

image

After the one in the PR, I think I like Option 4 best.  I think there are probably pros and cons to each...  let's decide which to use and then I'll push the change  :)

ubuntujaggers:
Good point @clobrano, let's try.

image

image

Alternatively, for this one, we could introduce a bit more padding between the perimeter and the object, to stop it looking as dense and dark:

image

image

Personally, I think either would be viable?

clobrano:
> Personally, I think either would be viable?

I agree, maybe a little preference over the former

eaglersdeveloper:
I prefer the third variant.

eaglersdeveloper:
Pleaaase

ubuntujaggers:
It certainly isn't my least favourite - what does the thread think?

clobrano:
I prefer the grey fill

eaglersdeveloper:
All icons aren't filled over. There are only outlines.
If you make the icon filled, it will be different from other icons in style.

eaglersdeveloper:
All icons aren't filled. There are only outlines.
If you make the icon filled, it will be different from other icons in style.

clobrano:
There are many icons filled as much as this one, actually

clobrano:
I started thinking that Adwaita might be the best, indeed, at least in small sizes. So, for me, undivided requires the filled bg, but Adwaita's style is very good as well.

Issue #697, opened 2018-08-09
[sound] Low frequencies barely audible with some laptop speakers


charakterziffer:
I discovered the sound files of this theme and downloaded them to hear them. With my laptop speakers and the volume I usually listen to music and podcasts some of the yaru sounds were not or barely audible: audio-volume-change, dialogue-error, dialogue-question (I hear three ticks), message-new-mail (three softer ticks).

When I boost the volume, I hear most of the sounds, but then others are clearly too loud (desktop-login, trash-empty). With headphones the result is better. So I’d say it’s not a volume or loudness problem but that my laptop speakers are weak in the lower frequency area. I know my own speakers are not that good, so could be other’s.

Maybe think about to enrich the very low sounds with higher frequencies which should be better audible even on worse speakers.

PR #1085, opened 2019-01-06
Amplified system sounds

madsrh:
I didn't change the frequency of the system sounds, but I've made them slightly louder.

Hopefully this fixes https://github.com/ubuntu/yaru/issues/697


Feichtmeier:
It works :+1: 
And I am quiet sure it is about 5%-10% louder :)

Do you think that you could fix that overdrive when you in-/decrease the system loudness "too fast" in this branch, too?
Or shall we do this in a different run?
I don't know if you remember, but I've send you a sound file once in telegram

madsrh:
> I don't know if you remember, but I've send you a sound file once in telegram

I remember - sorry 😞 

Let's merge this one and look at the other issue(s) separately

Issue #1086, opened 2019-01-07
nautilus 3.30 not styled properly


iainlane:
This should have been fixed in https://github.com/ubuntu/yaru/pull/982, but it's not working for me with yaru master.

![yarunautilus330](https://user-images.githubusercontent.com/321014/50772916-37f1f700-1287-11e9-9eb9-772bc5a6d358.png)

Part of the problem is that the path bar isn't "`linked`" (any more? I didn't see when it was, must have been when the PR was being developed though). If I use the inspector to add that class to the widget, it's better but some things still don't work properly like underlining on mouseover.

To test this, you can install `nautilus`, `ubuntu-settings` and `gnome-shell-extension-desktop-icons` from disco-proposed. Take those packages only, and be sure to disable disco-proposed straight away after installing.

cc @Feichtmeier

iainlane:
No! I'm saying that it is *broken* when I clone git and run nautilus with that yaru

Feichtmeier:
Oh... Uhm, this is strange. 

I installed nautilus 3.30 + tracker in my 18.10 installation via https://launchpad.net/~gnome3-team/+archive/ubuntu/gnome3-staging
Works very fine there. Are the packages different in disco proposed and gnome staging>cosmic  repos?

iainlane:
Yeah, one difference is that this is 3.30.4 and not 3.30.0.

Ahh, it looks like it's https://gitlab.gnome.org/GNOME/nautilus/commit/07d4876563ea7e032d4467742df2133eb3058dda :(

Feichtmeier:
Ah alrighty, I will fix that, thank you!

PR #1063, opened 2018-12-28
Remake shutdown icon

eaglersdeveloper:
Now this icon looks in the same style as other icons.

system-shutdown-symbolic system-shutdown-symbolic-new

PR #1089, opened 2019-01-08
Fix .path-bar button styling

Feichtmeier:
Closes https://github.com/ubuntu/yaru/issues/1086

peek 2019-01-08 10-21

@iainlane please test again

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1746,19 +1746,6 @@ headerbar {
         &, &:checked { border-color: lighten($inkstone, 3%); }
       }
     }
-
-    .path-bar.linked button {
-      background-color: transparent;
-      border-radius: 0;
-      &:checked, &checked:active { 
-        color: $headerbar_fg_color; 
-        font-weight: normal;
-      }
-
-      &.slider-button {
-        box-shadow: none;
-      }
-    }
   }
 
   &.selection-mode {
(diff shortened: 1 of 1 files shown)


clobrano:
Looks good to me now (I have a VM with disco and Nautilus 3.30.4)

image

Feichtmeier:
Good news it still works on 18.04 nautilus 3.26:

peek 2019-01-08 10-53

iainlane:
Yep, thanks, this looks great!

@clobrano once this is merged we could do the changelog / tag / upload IMHO

clobrano:
@iainlane, great I'll start updating the changelog asap

Feichtmeier:
Offtopic but still somehow related to nautilus:

The new shell extension styles the selected item color here:
https://gitlab.gnome.org/World/ShellExtensions/desktop-icons/blob/master/stylesheet.css#L12
Which can not be overridden by the shell theme ( I guess because the extension's stylesheet is taken after)

image

The border-radius for example is not styled in the extensions stylesheet, thus I can change it on our shell theme.

Could you either remove the styling from the fork or change the bg color to orange/transparent orange?

clobrano:
I'd really like to revert #1063 before tagging, @Feichtmeier do you think it is possible in a short time?

Feichtmeier:
I think @ubuntujaggers still needs some time to fix stuff so I think we can risk the conflict in his branch. He only needs to copy the part from the SVG into his SVG then. So yes let's do the revert now

iainlane:
There is no fork, if we want to make that themable we should try to figure out how to do that properly upstream.

clobrano:
> Do I just need to manually revert the power glyph in my copy of source-symbolic so it's the bolder one? If so, yes, that's easy, I'll make sure I remember :)

yes, that should do it, thanks!

Issue #1040, opened 2018-12-12
Refusal of small icons


eaglersdeveloper:
In Unity 8, only one application icon was used. And now are 12 icons of different sizes. What if we refuse small icons?

Firstly, less expensive for developers.
Secondly, possibility of using vector icons.
Thirdly, looks better.

On the left is 256x256 icons.
On the right is 48x48 icons.

![Снимок экрана от 2018-12-10 14-48-34|67x585](https://ubuntucommunity.s3-us-east-2.amazonaws.com/original/2X/5/5aa01bb4978715537abe342243571e2ee6fd9e52.png) ![Снимок экрана от 2018-12-10 14-48-01|66x579](https://ubuntucommunity.s3-us-east-2.amazonaws.com/original/2X/e/ef350c50a7f9a9c91ad07bf8b7fbb7f7529e22e6.png)

P.S. I posted the same thing on Ubuntu Discourse, but no one responded: https://discourse.ubuntu.com/t/call-for-participation-an-ubuntu-default-theme-lead-by-the-community/1545/2041?u=eaglers

clobrano:
Icons are used also in other places than dock and dash. Some notifications for example uses it and might makes the notification bubble too big if there isn't the proper size available

eaglersdeveloper:
Then refuse 48x48 icons?

clobrano:
I don't think this would be an improvement, the adoption of .svg depends on the support available in gtk, not in the given icon and "looks better", well it depends on the application it is using the icon, I don't see a strong documentation of this, only one specific case in which this happen.

Feichtmeier:
I think we should not make a rushed decision here. Better look at upstream Suru and GTK/GNOME here. I think there is something on the horizon for dropping that folder structure and rendering the smaller sizes automatically to have less detail, but I don't know the development status of that. Thus, I will close this here. We should stick to the convention as long as it is the standard

eaglersdeveloper:
I don't understand what you are talking about.

eaglersdeveloper:
GNOME has already refused small icons.

Feichtmeier:
No, they didn't (at least not completely)

https://gitlab.gnome.org/GNOME/adwaita-icon-theme/tree/master/Adwaita

https://gitlab.gnome.org/GNOME/adwaita-icon-theme/blob/master/Adwaita/22x22/places/folder-documents.png

eaglersdeveloper:
[...]

PR #1098, opened 2019-01-09
Use 256x256 icons for all icon sizes >= 34

Feichtmeier:
Closes https://github.com/ubuntu/yaru/issues/1040

@clobrano @madsrh @eaglersdeveloper @ubuntujaggers please test this with differnt icon sizes (change in gnome-settings app in the dock row)

### DIFF
--- a/icons/Suru/index.theme
+++ b/icons/Suru/index.theme
@@ -69,13 +69,13 @@ Type=Fixed
 
 [48x48/apps]
 Context=Applications
-Size=48
+Size=256
 Type=Fixed
 
 [256x256/apps]
 Context=Applications
 Size=256
-MinSize=64
+MinSize=48
 MaxSize=256
 Type=Scalable
 
@@ -436,14 +436,14 @@ Type=Fixed
 [48x48@2x/apps]
 Context=Applications
 Scale=2
-Size=48
+Size=256
 Type=Fixed
 
 [256x256@2x/apps]
 Context=Apps
 Scale=2
 Size=256
-MinSize=64
+MinSize=48
 MaxSize=256
 Type=Scalable
 


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1098`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1098` channel to test your snaps with those changes.

madsrh:
Hmmm, I'm not really a fan 🤔 
I miss the squircle border and the sharpness (look at Tweaks and Files). 
Unless the user changes the size to 64px - and I doubt that very many users do that - it's a step down in quality IMHO.

dock

ubuntujaggers:
Idea for an experiment: export the svgs that are used to generate the 48px icons at a suitably enormous size and see how _those_ pngs look if we use them for all icon sizes >34px.

For instance, here's the svg for the 48px version of messaging exported at 512dpi to make a 256px png:

big messaging app icon

It doesn't exactly shine at that size, but we would just need to see how it looks when it's actually on the desktop, scaled to typical sizes found in the app grid and on the launcher.

The result of this experiment would be a launcher that looks exactly the same as now (at the default size at least), and slight changes to the appearance of the icons in the app grid (and we'd have to decide if the app grid still looked good).

The point of this experiment is: **if** it looks better, when we switch to svg, we could either:

a) use the svgs that currently render the 48px icons instead of the more detailed (and less "buttony") svgs that render the 256px icons (because the beauty of an svg is, you can use any size svg and it won't pixellate or blur when scaling to larger sizes) - or

b) make new hybrid svgs that combine the more detailed pictograms/designs from the 256px version with the border details from the 48px version.

When we switch to one svg per app, there will be an inevitable slight trade off on icon quality, either in the app grid or the launcher.  The five-sizes approach was alien to me when I started working on these icons, but now I get it.  With the exact same image scaled to all sizes, you either lose the nice defined edging on the smaller icons (if you use the 256px Suru template for the svgs) or else you make the edging look a bit less sleek/sharp/more cartoonish on larger icons (if you use the 48px Suru template).

We found that when we were experimenting with the script, I think.  Once it used all the different squircles for the different sizes, the result was better.  I suspect that going with b) above might be the best compromise in the long run when we switch, but I guess we have to do some experiments?

ubuntujaggers:
> @madsrh I agree that the sh arpness is gone because PNGs are not meant to be scaled. Tried, tested failed :D

Actually, I believe it will be exactly the same when you switch to svg because svgs have similar problems to pngs when you scale _down_ to small sizes, in my experience.

The advantage of an svg is you can make it as big as you want and it will render the graphic as clearly as possible by redrawing all the curves, etc., at that size.  But whatever size it draws the graphic at, it has to use that resolution.  The quality problem with scaling an svg _down_ to 48px is, you might end up with nodes, boundaries, etc., falling between the pixels, and then it has to make decisions about how to colour individual pixels.  Stuff like anti-aliasing comes into play.

**Demo:** here is the actual svg of the 256px file manager icon zoomed to 48px inside Inkscape.  So this isn't a png, it's the real svg scaled to size in a vector graphics application:

image

...and it looks identical to the png in @madsrh screenshot (and slightly less sharp than the current dedicated 48px icon):

image

Say you have a red and blue icon.  If you have a dedicated optimised png at 48 px, you can make sure that pixel _a_ is red and pixel _b_ is a blue for a perfectly crisp boundary between the red and the blue.

If you scale an svg down to that size, when it renders, the boundary might not lie at an exact pixel.  If it ends up falling in the middle of pixel _b_, it may be that pixel _b_ has to be a whole pixel of an intermediary purple colour.  **Result:** not as sharp distinction between the red and blue when you view the svg at 48px, compared to a dedicated svg that was optimised for that size (or a png export of a dedicated svg that was optimised for that size).

So svgs scale _up_ with maximum quality, but when you scale them down to small sizes, unless it's a size you designed the svg to render optimally at - with all elements ending at exact pixel boundaries - it won't be quite as crisp as the current 48px icons in my experience  :(

Unless gnome has some new vector scaling thing that's better than Inkscape, I think the png test is actually representative of what the 256px svgs will look when scaled on the launcher, as per the comparison screenshots above.

Feichtmeier:
It's not the same for how the shell renders SVGs
For example download the icon theme la-capitaine from GitHub. He uses SVG only and as far as I can see it there is no stretching or blurriness
But that was not what I meant. I meant that we could try to export the big sizes also into the 48x48 folders. But maybe it look stretched again then because it's still a PNG.
We could still target the SVG only approach and just provide more gnome apps
Basically what you and eaglers suggested in your two topics

ubuntujaggers:
"It's not the same for how the shell renders SVGs"

I think it has to be.  If I'm drawing a dotted line in a 256px svg and I want it to look crisp at 64px as well - e.g., because I know it will often be seen at 64px in the target OS - I have to do the dots as 4px squares and align them to a 4px grid. Then at quarter size the dotted line will be black pixel, white pixel, black pixel, white pixel, positioned exactly correctly for maximum clarity with no grey. Otherwise it won't be perfect at 64px and there will be anti-aliasing. There's no way of avoiding that?

The reason we're seeing blurriness here is twofold: a) 256 doesn't scale nicely to 48px in terms of pixel boundaries (if the launcher icons were 64px by default this would be very easy because you just make sure the 256px icon is aligned to a 4px grid) and b) we're used to seeing launcher icons that have been meticulously optimised for the default launcher size of 48px. It's less the case that the new file manager is too blurry, more the case that the previous one was ultra-sharp. I think the new icons would seem perfectly good-looking to us if the previous icons hadn't raised our expectations of Suru.

ubuntujaggers:
Can we do a test with the file manager svg on the launcher, is that supported yet? I'll be genuinely delighted if I'm wrong about this :)

ubuntujaggers:
There is a really useful guide to this topic linked in the Suru guidelines, which I wanted to share earlier, but didn't have to hand on my phone: https://dcurt.is/pixel-fitting.  The author does a much better job of explaining than me  :)  so I wanted to return to this topic.

_"Most icons and logos are stored in vector image files that specify perfect mathematical representations of their shapes. Unlike JPG and PNG images, vector graphic files don’t tell the computer how to display their contents in pixels; for a vector graphic to be displayed, the computer has to perform a translation from the mathematical vectors into something that can be displayed with pixels [...] While the results are generally decent, they’re not perfect."_

The author gives comparisons of vector logos that a human has manually resized downwards (as per current Suru practice) with ones that a computer has scaled down (as per future Gnome practice).

_"The difference is subtle, but the comparison makes it very clear... For example, in the transition from 66x40 to 48x30 pixels, the human version drops the border down to exactly 2 pixels but the computer drops it to 2.8 (that’s two solid black pixels, plus one half-pixel at 80% intensity), which gives it an incorrect soft edge. The computer has no way of knowing that the border should always be sharp, so it ignorantly tries to maintain the ratio of the originally-specified image at the expense of sharpness."_

If you look at our current 48px icons there will be lots of borders that are exactly 1 or 2 pixels wide because the artist scaled the image down and then manually tweaked the position of the nodes to look as good as possible at 48px.   Whereas automatically scaling down the 256px of the same icon without tweaking would make lines of 1.9 or 2.3 pixels in size.

Best example I could think of was an icon of something like the French flag which is divided into thirds.  If you automatically scale your svg down to 40px, that doesn't divide exactly by 3, so it will have to render on screen as (from left to right): 13px of blue, 1px of very light blue, 12px of white, 1px of pink, 13px of red:

image

image

Whereas if you can supply a dedicated 40px version of the logo that you enhanced by hand, you can (as a human) think about making a little compromise that no one will notice - like (for that size only) making the thirds slightly unequal so they render at exact pixel boundaries.  Here, I slightly moved the nodes by a sub-pixel amount to produce a version that would render at 40px with exactly 14px blue, 13px white, 13px red:

image

image

I don't think it's a cause for despair, and in this case the first flag is actually fine.  It doesn't matter that the edges aren't 100% crisp at that size so it's not worth doing a separate 40px version of the svg.  You would only perceive the first flag as being a tiny bit blurry if you were used to seeing the second one.  But this will affect some svgs more than others (and at some sizes more than others).  **I fully believe it's solvable,** but suspect we will have to look at the full size svgs and (in some cases) make tiny tweaks so they scale as nicely as possible to the sizes we care about most.  For Yaru, that's 48px and whatever px (does anyone know?) the app grid uses for icons.  So, we can still have one svg per icon with a great-looking launcher, as long as we remember that it will scale down to some sizes better than others, and ensure that 48px and ??px are the best ones for our Suru svgs  :)

Feichtmeier:
Ahh thanks for the nice explanation!

Didn't gnome landed something to fix this? Could be Wayland only but there must be a reason that they make the app icons SVG now/soon

ubuntujaggers:
> there must be a reason that they make the app icons SVG now/soon

It may be because pixel-fitting icons is a solution that takes a lot of work, and it has its own problems anyway.  It's not perfect.  So maybe people now feel that the end result doesn't justify all the work.

I mean, one svg only is a perfectly fine approach, as long as you plan for it.  One part of the problem is, Yaru was designed before anyone expected to see the largest Suru graphics scaled down on the launcher.

If we'd been working with those svgs on the launcher from day one, 48px might have been rejected as the default icon size, because the big svgs render a bit more cleanly at 32px and 64px.  That's because these numbers divide exactly into 256, and the Suru house style recommends working to a 4px grid.  Two of those grid squares align exactly with one pixel when you display the icon at 32px.  So, most Suru pictograms render really cleanly at that size.  The straight edges will be very crisp and only diagonal lines and curves will require anti-aliasing:

image

image

Whereas 48px is 1/5.3333th of 256.  As we saw during the test, the full size Suru svgs have more antialiasing ( = slight blur) when they display at that size, because even straight lines end up falling between pixels:

image

image

Tbh, Messaging was a bad example to use, because the full size icon apparently looks quite nice at 48px 🤣 but you can see from the zoomed-in version that the straight boundaries genuinely aren't as crisp.  That will affect some icons more than others.  Some will look fine at 48px, and a couple (I think someone said Files and Tweaks?) will have enough extra anti-aliasing for it to become noticeable as a loss of quality.

Scaling a single svg becomes more of problem at the smallest sizes, but Gnome are moving to symbolic icons for tiny sizes, so that's not a problem.  In a sense, they'll still be having multiple icons optimised for different sizes.  It's just the number of sizes is now reduced (there will only be two per icon) and they've made a stylistic decision to go with symbols rather than full colour for the small one.  Under that system, we still get the advantage of being able to optimise one icon to be as clear as possible when it's shown at 16px.

One argument against having five svgs optimised for five sizes is, you can't do it for every possible size.  In the French flag example, even you made a perfect icon for 40px, the user might zoom in or out a bit so it renders at 42px or 38px, and then it's the same result.  In Ubuntu terms, we have some beautiful 48px icons, but the user can change the launcher size so they render at some other size.  If the user picks a size that you don't ship a dedicated icon for, you've lost the optimisations you did anyway.

So one way of looking at it is, you still get the French flag problem if you do different svgs for different sizes...  you just create a small number of "preferred sizes" which look better than the others.  And you can still achieve that with a single svg, provided you plan it a bit and think about the maths.

E.g., if you draw a 256px svg, you can align all the important boundaries to a grid where each main square of the grid is 16px squares.  Then you know that each square of the grid will render as:

- 16px exactly when the icon is shown at 256px;
- 4px exactly when it's shown at 64px;
- 2px exactly when it's shown at 32px; and
- 1px exactly when its is shown at 16px.

If you do that, your design will be clearest when it's scaled to those sizes, because your design divides neatly across the pixels.  And then you've got something very similar to the Suru situation.  You haven't made every size 100% crystal clear, but you've at least got some "preferred sizes" that look best.  Bonus: you haven't had to make multiple svgs (just one carefully planned one).

The problem with that is, the elements might seem cartoonishly bold at the largest size.  But I'm guessing you can include smaller elements of detail that aren't essential to the meaning of the graphic, which you can afford to lose between pixels at the smaller sizes.  I say "guessing" because I've never designed an icon that wasn't done using the current Suru approach of multiple icons, so it will be a learning experience for me anyway!

Anyway, I think both approaches can work as long as we bear it in mind when we work on the icons and the theme, and it may simply be that Gnome are aware of this and know that designers will start to optimise for the new system?

After typing all that: if Gnome have come up with a more intelligent way of rendering svgs, that would be superb, and I'd be very happy  :)  But I can also understand their decision if they don't have, and decided to switch to one svg + a symbol anyway.

ubuntujaggers:
Sorry for posting so much on this topic, but looks like this (all the above wall of text) is indeed what Gnome are encouraging.  They've updated the Human Interface Guidelines as follows:

_"Two styles of icon are used in GNOME 3: full-color and symbolic icons.

Full-color icons are colorful and detailed, and are optimized for larger sizes. They are defined as 128×128px SVGs, **and are sharpest when scaled up in multiples of 128 (such as 256✕256 and 512✕512).  The design of full-color icons also allows them to be rendered sharp at 64×64px and 32×32px, but is not advised to make them any smaller.**"_

So in Gnome terms, the full size svgs of Suru icons are already optimised to be sharp at all the right sizes - but our default launcher size (48px) is not one of the right sizes.  Also, we shouldn't expect other Gnome-compliant icons to be sharpest at that size when they change to svg.  The most Gnome-friendly thing to do would be to change the default size of the launcher, although I love 48px  :(

Feichtmeier:
Just woke up but first of all thanks for the eluding post. I think that I've understood the topic and the problem+solution now.
I think 32 could be a better default size than 64 though. Reading it later again with hopefully more ideas to reply :D

ubuntujaggers:
I suspect you're right because 64px looks really indulgently big (although the icons are beautiful lol).

I've been playing around with a new squircle template to try to optimise it a bit further at 32px.  This next squircle combines elements of the hi-res squircle with the smaller icon sizes (again, the colour quality has gone slightly naff on upload) and is shown at Gnome-preferred sizes:

new hybrid squircle 32px

new hybrid squircle 64px

new hybrid squircle 128px

At 32px, it means the squircle is 2px bigger (because the current hi-res svgs have a bit of padding) and it also means we get a bit more of the edging from the current 48px icons.

Here, at 32px, the current full size squircle is the top one, and the new squircle (WIP!) is the bottom one, and they're both shown on the default launcher colour:

image

IMO the bottom one looks a tiny bit better at 32px.  The edging makes it feel more defined and retains a bit more of the current launcher look.  Zoom in:

image

clobrano:
> : Gnome are now encouraging people to ship a single svg that's clearest at 32px, 64px, 128px and 256px, and not at 48px. If we stick with a 48px launcher, then that won't be conducive to new svg icons looking as clear as possible with the default settings.

I believe that this reason is valid more than enough to propose a default size change in Ubuntu. Maybe there's a gdconf setting like the one for the font size

Feichtmeier:
@madsrh I think we need to prepare that further. The first step was the big contribution from @ubuntujaggers with all the symbolic app icons (since GNOME encourages with the new svg design to not scale full color icons any further down than 32, and for places where they need to be smaller, they use symbolic icons instead)

We could now (if we really want to invest t he work *looking and jaggers*) make a new branch where we try the SVG only approach. If that turns out to be working great at the described sizes we still need to check if ubuntu is ready for that. 
Ubuntu still ships software-properties-gtk frontend outside of gnome-software (they plan to merge it, but no idea when) 
image
Ubuntu's gnome stack should be ready from cosmic on as far as I know but definitely not bionic (communitheme snap users) because gnome 3.30+ apps uses symbolic icons when the icon is too small.

I fear that it might still be too early

ubuntujaggers:
Agree that more prep needed (and more thinking) - the switch to one svg only is going to be an unexpectedly big change for us.

For a bit of context, the new Gnome svg icons would probably look something like this when they render at 48px (on the "app icon rejiggle" webpage at http://jimmac.musichall.cz/stuff/icon-revolution/ they render as 64px and are, as designed, beautifully clear at that size):

image

image

...which isn't 100% great but arguably fine if you don't take a screenshot and zoom in.  We can expect the same for any apps that redo their icon to current Gnome guidelines.

So, maybe a low cost "plan B" would be to keep the 48px launcher and ship two full colour icons of our own Suru icon set instead of just one (with the second one being optimised for 48px) if people feel that non-Suru icons are "good enough" when they render at 48px alongside our own icons.

I'm not saying that's the answer, but it could be a discussion point and would leave the current Suru icons looking 100% great on the default desktop - especially if the desktop team can switch all instances of 16px to the new symbolic icons (and we continue to work on those to make them as good and complete as possible).

But then I think it would be desirable to do Suru icons for as many popular 3rd party apps as possible, **when the licence allows it**, so our top open source apps look as good as possible on the launcher.

If we **do** transfer the current full size designs to a new Gnome-encouraged 128px svg, with the squircle details optimised to scale exactly how we prefer to the new launcher size (which would probably have to be 32px), then that's certainly doable - it would be helpful to get @eaglersdeveloper on board - I know he's keen on switching to the "one svg" approach, and I think his manipulation of the full size Suru designs is generally a bit better than mine (making sure drop shadows are 100% right, etc.).

ubuntujaggers:
@Feichtmeier, I somehow accidentally edited and wrote over your last comment when I was trying to reply to it - not sure how I did that - but I think I've put it back exactly how it was.  That's why it says "edited by me" on it.

Feichtmeier:
@ubuntujaggers there could another option still available which should be the least of a problem:

Why don't we "just" take the big SVGs in the src files and copy them onto the first one of the small sizes?
Shouldn't that be the 48 px icon?

Or will the same math problem occur when you render it and let the shell display it in the dock at 48px?

image

ubuntujaggers:
@Feichtmeier the resulting hybrid SVG would have beautiful edging at 48px but the pictogram wouldn't scale clearly to 48px.  Also, non-Suru icons designed to Gnome HIG wouldn't be at their best when people pinned them  :(

Here's my assessment of the current stock of Suru svgs in terms of the new "one SVG" approach.  Not all these points are relevant to your suggestion, they're just here in one place.

**256 PIXEL SVGS**

**Pictograms**

The pictograms from the 256px designs will mostly scale nicely to 128, 64 and 32px but not 48px.

Clarity isn't the only thing affecting how good they look.  Compared to our dedicated small icons, the drop shadows from the 256px svgs won't be as clear or emphatic at small sizes.  Currently, the smaller svgs often have sharp drop shadows of exactly 1px at chosen sizes.  A good example is the 48px version of the Messaging app:

image

...and we'll lose those if we scale down the 256px svgs, because they have more subtle drop shadows that are blurry and therefore don't snap to pixels:

image

In terms of general pictogram composition, while the shapes and positions should mostly scale clearly, we would still have to review them with 32px in mind (if we're going to show the icons at that size often) because the recommendation for Suru is to use a 4px grid, which is only half a pixel at 32px.  So, we should shrink them to 128px (the new Gnome size) and review with a 4px grid to see if any elements unnecessarily start or end at only half a grid square.  If it's possible to snap those elements to the new grid without hurting the design, they will render whole pixels (which = clearer detail) at 32px.

Lastly, when working on the 256px svgs, artists might have made little departures from the 4px grid in the largest size because there were occasions when it looked good to do so, because they didn't expect it to be scaled down by the desktop (because they knew they were also supplying dedicated icons for smaller sizes).  I'll certainly need to review my own designs when we switch to one svg per icon to see if I took any liberties with the grid that are no longer wise!

**Squircles**

These will render clearly at the same sizes as the pictograms but not necessarily at 48px.

Because they aren't as "buttony" as the smaller squircles, they might not "pop" in the same way, even when they render clearly at Gnome-approved sizes.  IMO, the pixel-width edging on our current smaller squircles is an asset because it looks good.  So, by scaling down the 256px squircles, we would lose a visual feature that some people like (this is a similar point to the drop shadows - pixel clarity isn't the only thing that determines if the icons look better or worse under the new system).

Also, because the squircle doesn't fill the whole available space in the 256px version, we'll lose precious pixels if we shrink it to smaller sizes.  For instance: when displayed at 32px, the visible squircles from the 256px icons will actually only be 30px across.

**48 PIXEL SVGS**

**Pictograms**

These will render clearly at 48px and multiples of 48px but not the Gnome preferred sizes, which are powers of 2.

Even when they render clearly at multiples of 48, they might look too bold/chunky/simplistic to be attractive at larger sizes.  This is because they're unlikely to contain borders or elements that are less than 1/48th of the area (because they were designed for 48px).  So, they don't have as much fine detail as the 256px svgs.

**Squircles**

These will render clearly at 48px and multiples of 48px, but, in my opinion, aren't suitable for any larger sizes, even when they render clearly.  The 1px edging doubles in thickness each time you double the size of the icon and looks a bit chunky/"cartoonish" even at only double size.  That's not a criticism - they were designed for 48px - but at app grid sizes they would look a bit like:

image

IMO if we go down this path, we really need a single (tweaked) svg that was optimised at 128px with scalability to small sizes in mind, because the full size svgs aren't quite as as good as the smaller ones at small sizes, and vice versa.

The squircle I'm playing around with currently looks like:

image

But there are really lots of options, I think having an experimental branch using svg icons where we can try different things would be a massive bonus.

One thing occurs to me: the 64px launcher looks great with the full-sized svgs, and zero editing or optimisation is needed.  How about, instead of making the launcher 32px by default, possibly making it 64px with autohide?

Looking at my Windows laptop (I'm doing a lot of Powerpoint at the minute, I'm normally 100% Ubuntu :P ) the desktop icons seem to be 64px, and go by default where the launcher does - in a vertical column.  And it looks fine.  So maybe the 64px launcher is only too big because the launcher takes up working space by default?  If so, maybe autohide is the answer.  Just (another!) suggestion.

PR #1103, opened 2019-01-10
Add icon for missing image

### LINKED ISSUE #1102: imagine-missing.svg
image

There is no Suru/Yaru icon for this one. Would be nice to have it :)

eaglersdeveloper:
Closes #1102
image-missing


Paz-it:
Since this icon is a kind of warning, the pictogram should remain RED. What do you think?

clobrano:
Yes, I'd like to see the colors inverted. Also because icons for file types
are mainly white

madsrh:
I agree with @clobrano. An inverted version like below would feel much "lighter" IMHO

Image of Yaktocat

ubuntujaggers:
I like this - I would like to see both versions in a realistic screenshot (white on red and then red on white).  White on red is very colourful  :)  but I note that the generic image icon is also colourful:

image

Question: when do you actually get a missing image icon, and are you likely to delete it immediately or fix it or something?  If it's something you'll want to take action on, when you see it in your file manager, it might not be a bad thing for it to be as colourful as the generic image icon but a bit more urgent-looking?

Feichtmeier:
It is shown when an application has no icon in the dock

clobrano:
Then my reference to file types icon is wrong, better a more colorful icon
like this one

Feichtmeier:
Here:
https://i.imgur.com/neiL3pm.jpg

ubuntujaggers:
If it goes on the dock should it be a squircle...?

Feichtmeier:
Ah sorry, i think eaglers could be right about this being based on the image icon shape and BG. It's also used when you add a wrong path to a GtkImage in all your new gtk apps ;0 (little joke because none of us makes gtk apps). But the first part was not a joke. Posting a screenshot later. I even think that's completely right even with that dark BG eaglets pstrd

clobrano:
> If it goes on the dock should it be a squircle...?

Since it can goes in both dock and popups, I think it's better without the squircle

Feichtmeier:
We could also create a symbolic icon for that

image

Feichtmeier:
@eaglersdeveloper could you try a colour inverted version of your icon? Red on white BG? Just to try both, I think there is the desire to try and see it

madsrh:
I tried a quick mockup with gray since the popup background is pure white 🤷‍♂️ Also in the bottom of the dock.

image

clobrano:
I am torn. Grey is probably better in notification, while red in dock :D 
I am ok with both, after all

Feichtmeier:
I would also go for both. Some reply would be great either way ;D

clobrano:
Okay, then red is probably too alarming, slight preference for the white version here

madsrh:
There's also the option of using the same colors as the old/current icon - dark background and red. Perhaps even add a white border similar to the current.

image


image

clobrano:
That would end up with a dark on dark when the dock is opaque

Feichtmeier:
We could something in between. Like inkstone :)

ubuntujaggers:
The red is my favourite Suru colour, but really I'm happy with any of these options *thumbs up*

eaglersdeveloper:
OK, wait.

eaglersdeveloper:
Which one do you like?

image-missing-red image-missing-light image-missing-dark

I like red.

Feichtmeier:
I would prefer black when I see those three beside each other
But as Carlo said the black could be hard to see in the dock
If you would not change the black to be brighter I would go for red, too

clobrano:
I'd like to see @Feichtmeier inkstone on a opaque dock. @eaglersdeveloper could you do this test?

clobrano:
Well, ok if @eaglersdeveloper does not have time to make the last change, I think we can accept the red one. Anyone against?

PR #1110, opened 2019-01-13
New Ubiquity icon

eaglersdeveloper:
Please, don't merge now.
I haven't created symbolic icon.

installer-app


Feichtmeier:
I prefer the current icon 
image

eaglersdeveloper:
An user won't find out what is this application with this icon and without the app name.

clobrano:
@eaglersdeveloper, your icon suggests more a "download" meaning than install. The current one might be not perfect, but at least suggests an operation on the disks. I agree with @Feichtmeier here

Feichtmeier:
The user won't search for this and even if he would, he will not find it because it is not installed by default anymore after you finished your installation of Ubuntu

You install Ubuntu and this icon is attached to the dock when you have the installer opened from within the live session.

eaglersdeveloper:
But what if the user is for first time launching Ubuntu?

Feichtmeier:
> But what if an user is for first time launching Ubuntu?

@eaglersdeveloper Ubitquity is the window you see when you have a new PC and insert your USB Stick then boot from the usb stick -> ubuntu installation starts and that is basically ubiquity. So if you chose "install ubuntu" you don't have a dock and thus don't even see the icon.
Whereas when you chose "try ubuntu" and THEN in that livesession you see that ubiquity logo on the desktop "Install ubuntu". When you double click it ubiquity opens again etc etc

So seeing three opinions against this I have to close this (though it is a shame, since you invested time into this good looking icon - why don't you ask before? : / )

madsrh:
Would this make it look more like you were installing Ubuntu? 😁  

image

eaglersdeveloper:
An user can press Space, select language and right away load LiveCD
https://losst.ru/ustanovka-ubuntu-18-04

eaglersdeveloper:
In addition, anything can be recorded to a disc, not only Ubuntu.

clobrano:
In any case, your icon doesn't seem to do a better job for me

PR #1114, opened 2019-01-13
.oga file extension for sounds

madsrh:
Renames the sound files to be **.oga** rather than **.ogg** as recommended in the [freedesktop sound theme specification](http://0pointer.de/public/sound-theme-spec.html).


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1114`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1114` channel to test your snaps with those changes.

clobrano:
Tested a couple of sounds (just because I am not sure how to test all of them) and it looks good :+1:

PR #1115, opened 2019-01-13
Adding light blue firefox icon - new svg and all the pngs.

ubuntujaggers:
This is my first time making my own branch and submitting a PR from it - let me know if I need to do anything differently.

image

image

Let's try it for a bit, I think it looks good in the launcher mockup, but if we don't like it in the long run we'll try gold next  :+1:


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1115`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1115` channel to test your snaps with those changes.

Issue #1122, opened 2019-01-17
Dark theme: incorrect highlight in nautilus rename popup


chrisjbillington:
There's an issue with the text highlighting in the Nautilus file-rename popup in the dark theme.

## Expected Behavior

The text should be highlighted orange like it is in the light theme:

![image](https://user-images.githubusercontent.com/1044087/51294111-7dc56280-19df-11e9-8a04-150870181abc.png)

## Actual Behavior

The text is highlighted in a colour very similar to the background colour and so is barely distinguishable from not being highlighted at all (it really is highlighted in this screenshot, if you squint you can see the text of the file extension is a slightly different shade as it is not selected):

![image](https://user-images.githubusercontent.com/1044087/51294141-a0f01200-19df-11e9-8fc9-38c8e1c5c532.png)

## Steps to Reproduce the Problem

  1. Use Yaru-dark
  2. Select a file in Nautilus and tap F2
  3. Notice the text in the popup does not look highlighted
  4. Type something or drag the mouse or tap ctrl-A to convince yourself it is in fact highlighted

## Specifications

I installed Yaru from the Arch Linux AUR. I switched to the "git" versions of the AUR package, which installs yaru directly from github. git-describe in the build directory for it gives: `r27-29-g305a4d0a` which looks like current master. I am using the yaru session this time, so it's not like #1018.

Software that presents the issue:
- Name: Nautilus/Gnome Files
- Version: 3.30.5-769665bd3

madsrh:
Again, a very thorough but report @chrisjbillington - thank you!

PR #1123, opened 2019-01-17
Don't include entry:focus and entry selection in the jetpopoverwildcard

Feichtmeier:
Closes https://github.com/ubuntu/yaru/issues/1122

peek 2019-01-17 07-20

@madsrh and/or @clobrano please test

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -2405,9 +2405,9 @@ popover.background {
 
   @if $variant=='dark' {
     separator { background-color: darken($menu_border,12%); &:backdrop { background-color: transparent; } }
-    *:not(fill):not(highlight):not(trough):not(slider):not(switch) { background-color: transparent; }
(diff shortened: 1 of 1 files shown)


clobrano:
Looks good :+1:

clobrano:
> @clobrano I think something happened to that docker container, could you try to fix it? Or could you tell me how I can fix it? Is there some CLI connection possible to that container?
> I think there are also a bunch of branches/snaps opened
> The guy from snapcraft said that the more snaps we have the worse the performance becomes. I'll clean up some branches, would be good if you could also make a cleaning :D
> (Though the error seems to be connected to some apt stuff but the builds take quiet long currently)

Error seems just a timeout of operations. I'll retry to build first

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1123`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1123` channel to test your snaps with those changes.

clobrano:
everything seems right now, merging it

Issue #1111, opened 2019-01-13
Changing volume with scroll wheel result is distorted sound


madsrh:
When you're using the shortcut keys on the keyboard to change volume, it creates a distorted sound.

[🔈](https://www.dropbox.com/s/og7schra771luak/t_voice5202110345065267720.ogg) [This]() is the issue!

This was reported by @Feichtmeier - for some reason, he is using the **bell.ogg** sound and not the **audio-volume-change.ogg** sound 🤷‍♂️ but it doesn't matter because the issue is still there.

[This ](https://www.dropbox.com/s/y26y9kd176tfl8i/audio-volume-change_mads_ogg.ogg)is what it sounds like on my box: [🔈](https://www.dropbox.com/s/y26y9kd176tfl8i/audio-volume-change_mads_ogg.ogg)
As you can hear, the issue is barely noticeable here. Can anyone else reproduce Feichtmeiers issue? @clobrano 

And [here ](https://www.dropbox.com/s/g1ps3tdfe9gs0t4/audio-volume-change-freedesktop.ogg)is what the Freedesktop audio-volume-change sounds like: [🔈](https://www.dropbox.com/s/g1ps3tdfe9gs0t4/audio-volume-change-freedesktop.ogg) 

Okay! So, my theory is that the glitches occurs because our sound is too long. This means that when the sound is looped (or rapidly repeated) the whole sound isn't played entirely. If the cut isn't done at a [zero crossing](https://en.wikipedia.org/wiki/Zero_crossing) (where the waveform is at 0) it creates clicks (glitches).

Comparing the files between Yaru and Freedesktop:
![auido](https://user-images.githubusercontent.com/3986894/51087390-f6fc5580-1752-11e9-9180-279e81e6fd3f.jpg)

Possible solutions:
1. Ignore this issue because it's a rare use case.
2. Create a shorter sound (this has proven to be difficult)
3. Use the Freedesktop sound

**EDIT:** Wow this is issue #1111 😄 Now I'm really looking forward to #1234 🤣

clobrano:
On my system the sound looks fine :/

Feichtmeier:
@clobrano hover with your mouse above the volume indicator in the gnome panel
Then use your mouse wheel quiet fast
Doesn't your sound overdrive?

madsrh:
@Feichtmeier ? 🤔

Feichtmeier:
[...]

PR #1127, opened 2019-01-19
Testing PR: audio-volume-change overdrive mitigation

Feichtmeier:
Don't merge this

Testing a brighter sound from @madsrh

Need this snap to test it on my host (VM does not really work well with sounds)

PR #1132, opened 2019-01-21
Replace audio-volume-change with a shorter sound, ...

Feichtmeier:
... resulting in not clipping when pressing volume buttons extensively

Finishing https://github.com/ubuntu/yaru/pull/1127
Closes https://github.com/ubuntu/yaru/issues/1111


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1132`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1132` channel to test your snaps with those changes.

Feichtmeier:
I bet it's not the only sound, but from the mads sound I tried it is. The reason for the clipping must be something with the length of the sound. Depending on how much time @madsrh currently has we could either keep this opened and try other short sounds or merge it and make another testing PR when mads has time for this. But at least we have a "bug"-free sound set in master :) (which we have otherwise totally)

clobrano:
Sounds good

madsrh:
TBH that sound was only to test a REALLY short file. IMHO we should go with the old one. This one doesn't fit with the other sounds.

Feichtmeier:
@madsrh hm but can't you create another sound that is short enough to not clip yet sounds like the other wooden sounds? :thinking:

Paz-it:
I can't reproduce that clipping issue in my side.

Feichtmeier:
I can reproduce it on several machines
You need to use your mouse wheel or fastly press the volume buttons

Feichtmeier:
If we are all -1 for this explicit sound then no need to take this commit into the history :) Maybe @madsrh comes up with a short alternative to the current sound some day ^^

madsrh:
Sorry for not replying. Sure, I'll look into it

PR #1135, opened 2019-01-23
New icons

mymike00:
icons for atom brackets filezilla firefox inkscape skype telegram
I made this icons time ago for myself as [suru](https://github.com/snwh/suru-icon-theme/) didn't accept 3rd part icons...
some of them might not be the best icons ever, feel free to improve them if you want to use them
I hope there won't be license problems (where I took the original images, etc...)


clobrano:
Hi @mymike00, thanks for contributing to Yaru.

The icons look good, however (also considering the current discussion on icons) for Telegram, Bracket, Atom and Skype we should has the owners first, same for Filezilla and we already have Firefox (for which we might need to contact Mozzilla anyway). Inkscape is already in development in another PR.

mymike00:
why do you need to contact every software developer to ask if you can use a custom icon?
isn't a custom icons quite the same as editing the original one with a script?

PR #1137, opened 2019-01-25
Add a build option to install compatibility stubs for communitheme theme name

jhenstridge:
This PR is intended for use by the gtk-common-themes snap.  Currently we build Yaru from source and then include files for Communitheme from the edge channel of the snap.  I want to remove the dependency on the snap from the build process while still maintaining compatibility.  This gives us:

1. a smaller gtk-common-themes snap
2. assurance that we have one version of Yaru/Communitheme in the snap

Rather than simply symlinking theme names, I've used icon theme/sound theme inheritance to point at the Yaru names.  For the GTK theme, I symlink the directories under the theme name.  This is because I can export directories via the content interface but not symlinks (the Suru symlink in the current snap causes problems).

While I could do this from the gtk-common-themes snapcraft.yaml ([that was my first test](https://gitlab.gnome.org/Community/Ubuntu/gtk-common-themes/merge_requests/8)), I think it will be easier to update as part of Yaru.  While I haven't tried it here, this approach could also be extended to building the Communitheme snap itself if you wanted to migrate those users over to the Yaru name without breaking backward compatibility.


clobrano:
Looks good, thanks!

PR #1139, opened 2019-01-27
Remove top highlight for switches in headerbar

Feichtmeier:
All elements in the headerbar are completely flat and sharp except the switches.
The darker border-top and the box-shadow make the round corners a little bit unsharp.

Before/current:
with
image

After flatness applied:
without
image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1139`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1139` channel to test your snaps with those changes.

clobrano:
I agree, nice change :+1:

PR #1142, opened 2019-01-29
Fix HB suggested- and destructive-action buttons border color

Feichtmeier:
suggested and destructive btns Border-color changes on hover and active in headerbar
wtf

This should fix it:
wtf_weg


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1142`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1142` channel to test your snaps with those changes.

PR #1145, opened 2019-01-31
Add inode-directory symlinks to folder icon

aidanharris:
Some applications such as KDE's Dolphin lookup the mimetype of files to
decide which icon to use. For folders that's inode/directory. This
mean's Yaru's custom folder icon is ignored and instead a fallback icon
used from another theme.

**Before:**

(falling back to what I believe is the Humanity icon theme?)

**After:**

(correctly using Yaru's folder icon)

### DIFF
--- a/icons/Suru/16x16/places/inode-directory.png
+++ b/icons/Suru/16x16/places/inode-directory.png
@@ -1,0 +1,1 @@
+folder.png
--- a/icons/Suru/16x16@2x/places/inode-directory.png
+++ b/icons/Suru/16x16@2x/places/inode-directory.png
@@ -1,0 +1,1 @@
+folder.png
--- a/icons/Suru/24x24/places/inode-directory.png
+++ b/icons/Suru/24x24/places/inode-directory.png
@@ -1,0 +1,1 @@
+folder.png
--- a/icons/Suru/24x24@2x/places/inode-directory.png
+++ b/icons/Suru/24x24@2x/places/inode-directory.png
@@ -1,0 +1,1 @@
+folder.png
--- a/icons/Suru/256x256/places/inode-directory.png
+++ b/icons/Suru/256x256/places/inode-directory.png
@@ -1,0 +1,1 @@
+folder.png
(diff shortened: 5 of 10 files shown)


eaglersdeveloper:
Thanks! I noticed on your screenshot that other icons, such as 7z and Images folder, also aren't have Yaru icon. I guess I can fix they.

@clobrano, @madsrh, will you merge it?

clobrano:
@aidanharris, thanks for taking time to contributing, I just wonder if this is the best approach.

Yaru has a script to automate PNG generation (`render-bitmaps.py`), so I think the best would be to create a new `icons/src/fullcolor/mimetypes/inode-directory.svg` from `icons/src/fullcolor/places/folders.svg` and then let the script do the work. 
Just pay attention that `folders.svg` is a container for all the different folders icons, while we need only one.

eaglersdeveloper:
I don't know how to portray inode directory icon 🤷‍♂️
Humanity and [even Breeze](https://github.com/KDE/breeze-icons/blob/master/icons/mimetypes/64/inode-directory.svg) (KDE) uses symlink to usual folder icon.

I for symlink.

clobrano:
What I meant was to copy the same svg from places/folders

eaglersdeveloper:
How is this different from symlink?

Feichtmeier:
I think the symlinks are the correct approach and it is done in this way throughout the whole icon theme when two apps should have the same icon.

The only reason against merging would be that we actually draw a line at GNOME Desktop, but since this is such a small change, I think we can make an exception here.

eaglersdeveloper:
The fact is that an user of standard Ubuntu easy can install any other file manager.

clobrano:
If you want to have a look for reference, I did it in branch `add-inode-directory-mimetype`.

clobrano:
I prefer this approach, however if for the others it's not, feel free to merge this PR.
Moreover, I'd prefer @aidanharris does this, since the this is his own idea.

clobrano:
Alright, sorry @aidanharris, there are some problems with this PR.
If you like to fix them it would be great, but in the meantime I will try to reproduce the same on a different branch though, so that we can speed everything up and have a working dolphin sooner

Feichtmeier:
@clobrano to get this done and honour the idea we could commit the .list file and merge it

PR #1146, opened 2019-02-01
Revert "Removes firefox headerbar bleed fix because ...."

Feichtmeier:
Reverts ubuntu/yaru#1037

The issue is back
@madsrh @clobrano decide if you want to merge this can't commit until today's evening

image


clobrano:
The revert doesn't seem to fix the issue this time

Feichtmeier:
@clobrano yeah indeed! Also, all the firefox styling does not work anymore. They changed something with the ID I guess.... let's investigate. We can close this then, no need for crappy commits in history :D

PR #1148, opened 2019-02-02
Add inode-directory mimetype

clobrano:
inode-directory is the mimetype for folders, used by some application
(e.g. Dolphin) to select the icon to use.

Here I created a new `mimetype/inode-directory.svg`, taking the
standard folder SVG from `places/folders.svg`, then I used
`render-bitmap.py` to generate the related PNGs.


eaglersdeveloper:
Firstly: this icon should be in `places`.
Secondly: what if we change standard folder icon? Symlink is in this case the perfect choice.

clobrano:
This is a mimetype, therefore it doesn't belong to places. I understand point 2, but I don't expect to change folder icon so much that it becomes a concern.

Feichtmeier:
Ahhh that's why the script didn't work on my VM
Did you try to run the script when excluding all from the "string array" but mimetypes? You would need to"manipulate" the script fur this ofc :D

clobrano:
No I didn't do that, but it works actually :D 
Nope, after all the script works with a line like below

`.../places/folder.png inode-directory.png`

in `mimetype.list`. Of course, when we move to full svg, this would be useless and we would need a dedicated SVG as well or a link between SVGs.

Anyway, either this way of symlinking in #1145 or this current PR would do

PR #1154, opened 2019-02-05
Allow generate-symlink providing a pattern

clobrano:
`generate-symlink` now accepts one argument to filter out
any `.list` entry that does not match with such argument.
With this change, it is possible to generate only specific symlinks.

If no argument is provided, the script falls back to the usual behavior,
symlinking all the entries in any `.list` file.

e.g.

  The command

    $ ./generate-simlink.sh folder-open

  generates only symlinks for `.list` entries that concern
`folder-open`, both as source and destination

  folder-open folder          (OK)
  folder-close folder-open    (OK)
  folder-one   folder-two     (NO)
  image-x      image-y        (NO)

### DIFF
--- a/icons/src/symlinks/generate-symlinks.sh
+++ b/icons/src/symlinks/generate-symlinks.sh
@@ -20,10 +20,10 @@
 
 DIR=$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )
 THEME="Suru"
+needle=$1 # Symlink only the files that matches the needle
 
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1154`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1154` channel to test your snaps with those changes.

Feichtmeier:
Works!

image

PR #1161, opened 2019-02-08
Adding a pgp mime icon

madsrh:
Add a mime icon for pgp keys:

application-pgp-keys


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1161`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1161` channel to test your snaps with those changes.

eaglersdeveloper:
What about key icon as pictogram?

madsrh:
@eaglersdeveloper I tried that too but it doesn't look good at small sizes.

I'm fine with both solutions, but I think this one IMHO works best (although it isn't actually a key 😆)

eaglersdeveloper:
Turn it over :)

madsrh:
What do you think?

image

clobrano:
I still prefer the lock. Maybe it's just my problem, but the vertical key is a bit less recognizable

Feichtmeier:
I prefer the lock, too! :dancer: 
I think the vertical key looks like a cyclope-elephant on smaller sizes but that's only my stupid brain not working correctly :D

madsrh:
Allright the lock it is 👍 

image

PR #1151, opened 2019-02-03
Improve existing emblems and add new

eaglersdeveloper:
I improved existing emblems to match one style and add new.

I made some screenshots. Take a look:

   

Closes https://github.com/ubuntu/yaru/issues/1039

PR #1166, opened 2019-02-09
Graphite for emblem-symbolic-link

madsrh:
The color was heavily discussed in https://github.com/ubuntu/yaru/pull/1151 and perhaps something was lost in translation, but I figured it would be better to create a PR so we can try it in the real and see how it feels.

Changed the color of emblem-symbolic-link to graphite (#666666) to make it visible on both light and dark backgrounds.

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1166`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1166` channel to test your snaps with those changes.

eaglersdeveloper:
It looks very bright. Maybe better `#2d2d2d`?

Feichtmeier:
I think this looks a bit less heavy than those black circles

image

madsrh:
> I think this looks a bit less heavy than those black circles and I would prefer this

+1 It looks a bit odd at first but I've been using it for some days now and it's a much better fit compared to the black IMHO

PR #1171, opened 2019-02-10
Add fullcolor SSD icon

eaglersdeveloper:
ssd-icon


madsrh:
🤔 It could be mistaken for a chip / cpu. I'm missing the "drive" part, but I don't have a better idea, so... 🤷‍♂️

clobrano:
Uhm, no for me as well. The idea of ssd or even storage is absent

Feichtmeier:
I like the idea of differentiation between the drive types in general
The question for me is: do users of all ages and skill levels have that old school HDD burned into their brain and are confused if they see a different symbol even if that one actually fits me to the real device hmmmm...

madsrh:
TBH I never liked the harddrive icon 100% 🙊 Maybe it's missing an LED or it's the perspective that's odd, but I prefer simply writing SSD on top of the icon 😃 Like [here](https://img00.deviantart.net/f1af/i/2009/098/5/9/slick_drives_remake_by_thvg.png) or [here.](https://i.stack.imgur.com/KbaZx.jpg)

eaglersdeveloper:
Google also shows the hard disks with the inscription "SSD" on the query "ssd icon".

eaglersdeveloper:
?

eaglersdeveloper:
The inscription on the 16x16 icon is poorly visible.

Feichtmeier:
@eaglersdeveloper Could you commit this one with the SSD text? I think you could just remove the SSD from the ultra small size

eaglersdeveloper:
What if I use a simple line?

clobrano:
Better without any text at all

eaglersdeveloper:
What do with symbolic icon?

eaglersdeveloper:
What do with symbolic icon?

madsrh:
@eaglersdeveloper What if we just keep the current "chip" icon? I know that it's a bit of a mixed message. 

IMHO all the drive icons needs work. We have like 7 icons in the Device folder and Humanity has +100 - sorry for the offtopic but this really needs work.

Feichtmeier:
I would prefer this: https://github.com/ubuntu/yaru/pull/1171#issuecomment-462243230
to the chip version. SSD is simply so much fitting to what it is. Everyone is talking about SSD and it is more than a simple abbreviation. It really nails the whole symbol to what it is

Feichtmeier:
@eaglersdeveloper could you please commit the version you showed us with the SSD text on it? The smallest size does not need anything. No line, no text just the baseplate. 

Please try to get this done in the next week, since you did not commit anything for 16 days, and we don't want the chip icon, I fear we'll close this soon otherwise. (UI freeze soon, we need to get things done, uncertain changes are not very helpful, though your work IS helpful)

eaglersdeveloper:
> What do with symbolic icon?

Feichtmeier:
Put the "SSD"-test on it as well, please.

eaglersdeveloper:
I don't think that the text will be visible

clobrano:
I agree, it would be probably too small. Probably what you already suggested [here](https://github.com/ubuntu/yaru/pull/1171#issuecomment-463079586) is actually better (I changed my mind)

Feichtmeier:
Yes commit it with the line then

eaglersdeveloper:
Done.

Feichtmeier:
Awesome, thank you!
Need to test it on my VM first

eaglersdeveloper:
@Feichtmeier, have you tested?

Feichtmeier:
I tried, but my VM does not realize it's host uses a SSD :thinking: 
This commit, where only 12 files are changed, looks good though (for some minutes however, I saw a commit with thousands of files changed, how did you remove the commit? :) )
File names looked good and the github preview file function looked good on each file, too.
Can we merge this without testing, @clobrano ?

clobrano:
I don't have an external ssd and the commit looks good. Let's merge it

PR #1186, opened 2019-02-14
Fix 'new email' sound

madsrh:
Apparently the [freedesktop sound-theme-spec](http://0pointer.de/public/sound-naming-spec.html) is outdated. Mail clients like Evolution and Thunderbird use _message.oga_ for mail notifications 🤷‍♂️ 
This PR simply renames the file. 

@clobrano perhaps we should add a symlink for the _message-new-email_ file for legacy?

Applications like Evolution will still play the _beep_ / _bell_ sound because that's the default setting in the notification plugin. I've filed a bug in Launchpad:  https://bugs.launchpad.net/ubuntu/+source/evolution/+bug/1815975

**EDIT:** We might want to try a completely different sound, but I'll create a seperate PR for that so we can test with a snap :tada:


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1186`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1186` channel to test your snaps with those changes.

clobrano:
With the old name, geary played the notification and thunderbird didn't. With the new name none of them played the notification :D 
It might depend on the application version? In any case, yes better use a symlink, I can add that when you are ready for merging this

madsrh:
I did some experiments myself:
**Thunderbird** 
If both files exist it plays `message.oga` first and if only one exist it will play that file. I wonder why this is different from your result 🤔 

**Evolution**
When set to "_use system sound_" Evolution just doesn't care! Even if both files are in the Yaru folder, it still plays the `message.oga` from the freedesktop folder 🤷‍♂️ 

Okay, I suggest that we keep the current filename (not the PR - sorry) and add a symlink for `message.oga`
@clobrano will you add the symlink?

clobrano:
So, do I revert the name change and use a symlink instead?

madsrh:
IMO yes

vitalkanev:
Why symbolic link is pointing to something in home directory. It would be invalid everywhere

clobrano:
I made a mistake, I'll fix it

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1186`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1186` channel to test your snaps with those changes.

clobrano:
@madsrh, @Feichtmeier have you had time to test this?

madsrh:
Not yet, sorry. I will do as soon as I get back home

Feichtmeier:
Just to understand this correctly - now all email apps, thunderbird, geary and evolution should have the same new email sound?

clobrano:
This is my understanding as well

madsrh:
> Evolution
When set to "use system sound" Evolution just doesn't care! Even if both files are in the Yaru folder, it still plays the message.oga from the freedesktop folder

I can't test atm but I still think Evolution is misbehaving

madsrh:
Can we merge this one @clobrano @Feichtmeier ?

clobrano:
If it works for you too, I guess you can

Feichtmeier:
Okay! I finally managed to install Geary, Thunderbird and Evolution on the same VM :tada: 

Thunderbird and Geary use the correct sound! And I think they did this before this PR, too.

But Evolution uses a custom sound! I found this plugin configuration which shows the reason:
image

Feichtmeier:
As discussed on telegram: merging this!

Issue #1180, opened 2019-02-13
Update Yaru with non-uniform icons and remove third-party app icons


Feichtmeier:
This issue is for organization only and communication between contributors! 
Any **off-topic** comments will be **REMOVED**

_____________________________________________________________________________________
_App Name | icon name | list | of | symlink-names | ..._

Remove Yaru icons for these preinstalled third-party apps:

- [x] Firefox | firefox
- [x] Thunderbird | thunderbird
- [x] Libreoffice Writer
- [x] Libreoffice Base
- [x] Libreoffice Calc
- [x] Libreoffice Draw
- [x] Libreoffice Impress

Remove system icons that have other issues
- [x] welcome-app

Remove Yaru icons for these non-preinstalled third-party apps:
- [x] Polari | polari | org.gnome.Polari

Update these Yaru icons with non-uniform baseplates for gnome "coreapps" which are pre-installed in the default Ubuntu installation:

- [x] AisleRiot Solitare
- [x] Alerts <-- doesn't show up in app grid but appears on launcher and currently has a Suru squircle
- [x] Amazon
- [x] Calculator
- [x] Calendar
- [x] Cheese
- [x] Document Viewer
- [x] e-Book Reader <-- has an upstream Suru icon
- [x] Files
- [x] Input Method <-- I think this one's already non-uniform, it's the keyboard?
- [x] Language Support
- [x] Mahjongg
- [x] Mines
- [x] Power Statistics
- [x] Remmina
- [x] Rhythmbox
- [x] Settings
- [x] Shotwell
- [x] Simple Scan
- [x] Software & Updates
- [x] Software Updater
- [x] Startup Applications
- [x] Startup Disk Creator
- [x] Sudoku
- [x] System Monitor
- [x] Text Editor
- [x] To Do
- [x] Transmission <-- do we want our own icon for this, I think it was one of the controversial ones?
- [x] Ubuntu Software Center
- [x] Videos
- [x] Archive Manager
- [x] Backups
- [x] Characters
- [x] Disk Usage Analyzer
- [x] Disks
- [x] Fonts
- [x] Help
- [x] Image Viewer
- [x] Logs
- [x] Passwords and Keys
- [x] Screenshot
- [x] Terminal
- [x] Live-Patch
- [x] Apport
- [x] photps-app  + symlinks
- [x] messaging-app + symlinks
- [x] ubiquity
- [x] notes-app + symlinks
- [x] clocks

ubuntujaggers:
[...]

PR #1209, opened 2019-02-24
Bring current non-uniform icon set into Yaru

ubuntujaggers:
Okay - this should be the same icon set but without any extra symlinks and without conflicts.

Unfortunately it wasn't as easy as I hoped with the symlinks, it didn't "just work".  But I ran the symlink script one-by-one and believe I've only committed relevant symlinks.  Even so it's a massive lot of commits, over a thousand, because you have ten versions of every new icon and then lots of symlinks for each changed icon.

Suggest we test this one (is it possible to roll back?) due to the sheer number of commits!!

image


ubuntujaggers:
Also: after I pushed, I deleted all my local Yaru-related folders (including usr/share/icons) and then cloned what I'd pushed to my own laptop and re-built it from there.  So I know it works (it's not like I'm sitting here with a version that works but might have unpushed changes or something).

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1209`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1209` channel to test your snaps with those changes.

Feichtmeier:
I think you missed disk usage analyzer =)

ubuntujaggers:
I'm into it!

didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1209`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1209` channel to test your snaps with those changes.

ubuntujaggers:
Missing symlink - seems to have fixed it.

Feichtmeier:
@ubuntujaggers package-app is missing and ubiquity :D Could you tackle that? :angel:

ubuntujaggers:
Oops, sorry for missed icons. But yes, I'd be happy to merge this branch if it works (rather than poking it any more!) and then just improve icons and add the few missing ones with normal non-enormous PRs if everyone's happy?

Feichtmeier:
Okay i'll merge this one, Update the issue with the missing icons and close the other pr. 

Thank you very much for this incredible one man work jaggers!!!

ubuntujaggers:
> Thank you very much for this incredible one man work jaggers!!!

No problem at all, I'm very glad we finally found a way forward with icons in Yaru  :)

- Do we want to update the screenshots on the main page of the Github?  I've got to say, when I look at the current screenshots with all the squircles, I think your squircle-free idea was definitely the right one in the end - it just looks a bit better and more polished IMO on my desktop.

- Dare we tip OMG Ubuntu off?  I think the last the wider community heard was that we were sticking squircles on absolutely everything...

madsrh:
> No problem at all, I'm very glad we finally found a way forward with icons in Yaru :)

🎉 +1

> Dare we tip OMG Ubuntu off?  I think the last the wider community heard was that we were sticking squircles on absolutely everything...

Ha ha, good point 😄 They will know on [April 18th](https://wiki.ubuntu.com/DiscoDingo/ReleaseSchedule) 😉 
I'm sorry to bring this up again, but the first question we will be asked is about 3rd party apps and non core Gnome apps. I shared [my thoughts here](https://github.com/ubuntu/yaru/issues/1180#issuecomment-464921377), but are we 100% on the same page here? I think that we need to be and include this in the OMG tip.

ubuntujaggers:
> I'm sorry to bring this up again, but the first question we will be asked is about 3rd party apps and non core Gnome apps. I shared [my thoughts here](https://github.com/ubuntu/yaru/issues/1180#issuecomment-464921377), but are we 100% on the same page here? I think that we need to be and include this in the OMG tip.

From my point of view, the good thing about the non-uniform icon set is we can mix and match.  We have the freedom to make our own Suru-ish icons for third party apps if a) upstream are happy, b) we want to, and c) I or someone else is able to draw something which looks better than the normal icon in the context of Yaru (which won't be always).  But if any of those criteria isn't met, we can just use the upstream icon like everyone else and it won't matter.  When all the criteria *are* met - as in, we've got a green light, we want it, and we've actually got a good-looking icon - I don't think there's anything stopping us?

So in terms of our stated policy on icons we could just leave the door open without committing to particular apps at this stage, other than Shotwell which already ticks the boxes?

clobrano:
I agree, point A is the most important. I don't know how we can make this clearer for upstream  and maintainers, but it would be the best if the idea comes from them first

ubuntujaggers:
@madsrh @Feichtmeier @clobrano: we're already on OMG! Ubuntu!

https://www.omgubuntu.co.uk/2019/02/ubuntu-yaru-icons-non-uniform

clobrano:
Yeah, it didn't take long 😄

PR #1215, opened 2019-02-27
Non-uniform notes-app icon and the symlinks

ubuntujaggers:
image

image

image

image

??


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1215`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1215` channel to test your snaps with those changes.

Feichtmeier:
Works and looks good (imho) thank you!
image
Waiting for +1 by @madsrh and/or @clobrano

clobrano:
I love the yellow note block, but not so much the blue lines, have you tried without them, or is it part of the original icon?

Feichtmeier:
It's part of the original icon

eaglersdeveloper:
How about to make square icons?

Feichtmeier:
@clobrano for reference this was the "original" suru icon for this:
image
jaggers changed the shape

madsrh:
> How about to make icons square?

+1 

image

Feichtmeier:
Hm but notebooks are normally not square :) I think now that we have the four baseplates we can choose the one that fits the real life object the icon should remind of best

clobrano:
Can we take this change to try replace the blue lines with handwriting like in the gnome icon? :)

Feichtmeier:
Ah you meant the fat blue lines... Prbly a good idea yes. Thought you meant the thin, transparent blue lines

madsrh:
> Can we take this change to try replace the blue lines with handwriting like in the gnome icon? :)

This is just a quick mockup. Perhpas Jaggers can do better but I'm not convinced that this is an improvement besides being closer to the upstream icon.

image

Feichtmeier:
Well, the problem is that there is not a single icon with that kind of text - and we have a lot of icons with text on it. 🤷‍♂️  Consistency police 👮 (NusiNusi) ;)

ubuntujaggers:
Could possibly do something more stylised, like lines of repeating loops, to represent handwriting while still being abstract?

Got to visit someone tonight and tomorrow but can have another stab at this on Friday...

Feichtmeier:
Yes sounds good, as long as it still fits with the rest of the icons. 

>  to represent handwriting while still being abstract?

Sounds perfect to me

clobrano:
> Could possibly do something more stylised, like lines of repeating loops, to represent handwriting while still being abstract?
> 
> Got to visit someone tonight and tomorrow but can have another stab at this on Friday...

exactly what I had in mind :+1:

eaglersdeveloper:
I think this is more appropriate for skeuomorph, we create flat icons.

clobrano:
> I think this is more appropriate for skeuomorph, we create flat icons.

Actually skeumorph and flat are different things. Also look at the Software icon, it has some 3D effect

eaglersdeveloper:
> Also look at the Software icon, it has some 3D effect

I don't like it too.

clobrano:
> > Also look at the Software icon, it has some 3D effect
> 
> I don't like it too.

I understand, but the fact we don't always do "flat icons", holds

Feichtmeier:
Hm, I am actually also not a fan of breaking the icon rules too hard. Gnome-software is an exception which seems un-dodgeable when you want a shopping bag. At least it still looks like folded paper/origami

For the next cycle, we could think of dropping some rules, like the "must to use a baseplate" or "don't leak out of the baseplate", since real life origami has not such rule ofc, but imho this is pretty much again a meta discussion in a pull request about the notes-app. 

I would prefer to merge this, since it is exactly what is the closest solution: transform the old suru squirlce notes-app into one of the new baseplates :man_shrugging:

ubuntujaggers:
image

image

image

IMO it looks okay at larger sizes but a bit messy when I try to use the same motif at smaller ones.  Could be a goer and maybe just needs more work.  Might be worth merging the thick blue lines version (which IMO is no worse than squircle one) and add it to the list of icons we want to improve?

Feichtmeier:
:D
Am I a sick guy when this reminds me of the breast hair of a comic hero? xD

I would prefer the lines. But we could try to make them less thick? Or this:

> Might be worth merging the thick blue lines version (which IMO is no worse than squircle one) and add it to the list of icons we want to improve?

clobrano:
Ok, +1 for the original lines :)

Feichtmeier:
Alright I merge this then, considering we have only 13 days left until UI freeze. Any further improvement is not forbidden! xD :D

PR #1217, opened 2019-02-28
build-helpers: trigger gtk-common-themes build on gitlab.gnome.org

### LINKED ISSUE #953: CI: travis build publishes busted gtk-common-themes snaps
I was checking out the gtk-common-themes edge channel to see how the GTK 3.18 support looked, and noticed that the snap seemed kind of weird: there was no `snap/metadata.yaml` in the package and various files were in the wrong location (e.g. Yaru's GTK 2 theme was in the wrong location).

I then realised that this snap hadn't actually been built from the gtk-common-themes snapcraft.yaml file, but instead from one constructed from half of one and half the other.  This is incredibly fragile, and means we can't tell what went into the snap.

I understand that we want to rebuild gtk-common-themes when Yaru is updated, but this should be a real build of the package (at least for stuff published to the main channels).  One option would be to have Travis trigger gtk-common-themes's own CI rather than trying to perform the build locally:

https://docs.gitlab.com/ce/ci/triggers/

This would involve something like:

1. generate a trigger token for gtk-common-themes on gitlab.gnome.org
2. Make that token available in Yaru's CI (encrypted variable, maybe?)
3. Have the Yaru CI post to the relevant gitlab REST API to trigger a rebuild.

jhenstridge:
This branch removes the CI code to build the gtk-common-themes snap, and instead posts a trigger request to gitlab.gnome.org to have it perform a build consistent with the way the snap is usually built.

This reduces the chance of building a broken snap, since Yaru's CI no longer needs to know how to build gtk-common-themes: it simply asks gitlab to build the snap using a particular revision ID of Yaru, and publish the result to a particular channel.

One other behaviour change is that rather than building the head revision of the pull request source branch, we now build `$TRAVIS_COMMIT` from the destination `ubuntu/yaru` repository.  This will be an ephemeral commit representing a merge of the pull request into Yaru master.  I think this is desirable, since it is what the communitheme build will be.

Fixes #953


jhenstridge:
Note that this is not PR on its own is not complete.  It will need some configuration changes here, probably by @didrocks.  Here's what I think is necessary:

1. Create a new trigger token for gtk-common-themes.  This can be done here:

    https://gitlab.gnome.org/Community/Ubuntu/gtk-common-themes/settings/ci_cd

2. In Yaru's Travis configuration, add a new project level variable `COMMON_THEMES_TRIGGER_TOKEN` whose contents is the trigger token generated in (1).

It also looks like the meat of this PR won't be tested, since the source branch is not from the `ubuntu/yaru` Git repository.  I'm not sure of the best way to test that.

Once this is merged, it should be possible to remove the `SNAP_COMMON_THEMES_TOKEN` project variable, since you will no longer be publishing that snap.

Feichtmeier:
Thanks for helping us @jhenstridge !

Something I'd like to add:

Our SCSS files reached a point where we can not cover 3.28+3.32 gnome shell anymore (since two changes in gnome shell 3.32 styling are dependant on a change in the .JS files, which ofc only happens in 3.32 and is not backported into 3.28 )
So I talked to @didrocks @clobrano about this and we thought of "unplugging" updates for the communitheme snap so we can continue to push changes to master for the current Ubuntu cycle without having to worry about shipping a broken snap for 18.04 users.

Could this unplugging (not yet made) be somehow relevant to this topic?

jhenstridge:
@Feichtmeier: this PR is only about how the gtk-common-themes snap is built, so doesn't really have any bearing on the communitheme snap.  Also, the gtk-common-themes snap doesn't ship the gnome-shell theme, so isn't directly affected by that problem.

As for the question of how to handle the incompatibility, I see two main options:

1. create a maintenance branch for bionic.  If you're happy to leave the theme mostly static for Bionic users, this is probably the easiest.  If you want to continue providing updates though, it could be a pain as the gnome-shell themes diverge.

2. add a build option to `meson_options.txt` to select the gnome-shell version, and expose it to scss using a `.scss` file generated with `configure_file()`.  If the version differences can be handled by a few `@if` statements, then this may be a workable solution.

In either case, you've got the question of what to publish on the snap store.  One option is to publish only one branch/configuration.  Another would be to set up tracks and publish both.

clobrano:
@jhenstridge, sorry we waited too much probably and this patch has some conflicts now

jhenstridge:
Sorry for the delay.  I've rebased the branch on current Yaru master, fixing the conflicts.

clobrano:
Thanks a lot!

About this 

> Note that this is not PR on its own is not complete. It will need some configuration changes here, probably by @didrocks. Here's what I think is necessary:

We can merge this PR without waiting for the other side of the contribution (or the other side is ready)?

jhenstridge:
> We can merge this PR without waiting for the other side of the contribution (or the other side is ready)?

Like the current `SNAP_TOKEN` usage, this requires action by someone who can edit the Travis-CI project configuration for Yaru.  The `COMMON_THEMES_TRIGGER_TOKEN` variable will need to be set to a token generated on gitlab.gnome.org.

Without the Travis-CI change to add the new secret, it won't be possible to trigger the build.

clobrano:
I see, so we will wait the other side :crossed_fingers:

Feichtmeier:
Little poke here, doesn't hurt anyone if it's open, yet maybe you guys have just forgotten about this one

@didrocks @jhenstridge

Feichtmeier:
Since we migrated to github actions for the meson build check, and the communitheme snap has not been updated since some years, could we somehow convert this into a github action @jhenstridge ?

Jupi007:
Closing this as no more relevant in the current situation.

PR #1220, opened 2019-03-01
Remove welcome-app svgs and pngs

Feichtmeier:
Ubuntu provides an unmodified circleoffriends logo for this app which is already a circle. One of our new baseplates is a circle, so it fits "out of the box" to our new design and this yaru icon is not needed.

Also, the app icon actually never worked, probably because we missed the correct icon name/symlink to the actual icon. 

Any vetos @madsrh @clobrano @ubuntujaggers ?


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1220`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1220` channel to test your snaps with those changes.

ubuntujaggers:
No objections from me  :+1:

PR #1222, opened 2019-03-01
Non-uniform version of messaging-app with symlinks

ubuntujaggers:
image

image

image

image

??


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1222`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1222` channel to test your snaps with those changes.

Feichtmeier:
considering this is not a preinstalled app, and it is exactly executed with one of the new shapes, let's merge it and if there are other ideas we could talk about it in the next cycle?

PR #1228, opened 2019-03-03
Remove 48px, 32px and 24px fullcolor icons

### LINKED ISSUE #1040: Refusal of small icons

eaglersdeveloper:
This PR removes raster application icons in 48, 32 and 24px sizes.

Closes #1040


Feichtmeier:
No @eaglersdeveloper 
Why aren't you listening to what we discuss and just ignore it and open a pull request? This icon theme is not ready for this. 
Look at the new gnome icons, they can do it because they prepared this via new guidelines and icons that look good on several sizes PLUS they do not have a dock with 48px sizes like ubuntu has

eaglersdeveloper:
But you even didn't try this icons. Why do you close PR right away without even checking it?

Firefox has only one fullcolor icon, that don't follow guidelines, but looks good in dock.

Feichtmeier:
Because it is currently not working out.

The dock has a 48px standard size when you install Ubuntu and don't change it. 
256px icons scaled to 48px icons look blurry. 
It is some math I don't remember currently, but @ubuntujaggers wrote lots and lots of explanation on how your idea COULD be made real. 

But you can't just remove the folders and say "That's it". As I said several times: Ubuntu and this icon theme is not ready for this yet. 
We have the issue where we could discuss the steps that are necessary before we could do this. I won't stop anyone from making a good plan, but we are also responsible for the content in the master, and just removing the folders is just bad (currently)

eaglersdeveloper:
> 256px icons scaled to 48px icons look blurry.

Then why did GNOME refuse small icons?

clobrano:
Because they don't have a dock whose size is 48px by default

Feichtmeier:
https://developer.gnome.org/hig/stable/icon-design.html.en

http://jimmac.musichall.cz/blog/2019-01-23-the-big-app-icon-redesign/

> The color icon is optimized for 64px (with a nominal size of 128px for historical reasons), but the simple geometric style without 1px strokes means that it also looks good larger and smaller.

image

The ubuntu launcher has a size scale for 48px (not a multitude of 2^n: 2,4,8,16,32,64,128), which is even the DEFAULT. This means that the icon is scaled down by gnome shell from either the next smaller size or the next higher size, depending on how you set up the index.theme
peek 2019-03-04 12-09

eaglersdeveloper:
GNOME also has dock in Activities. And yes, default icon size is also 48px.

clobrano:
It looks 64px to me

eaglersdeveloper:
Really 😐

However, a raster image doesn't lose quality when it is reduced.

ubuntujaggers:
It's not so much about quality, it's about pixels.  If you scale a graphic of any format to random sizes, the edges of objects, lines, etc., will sometimes fall between pixels, and that means you have antialiasing (because instead of black pixel / white pixel it has to be black pixel / grey pixel / white pixel - because the boundary between black and white falls in the middle of the second pixel).  That doesn't matter for something like a photo, but it does matter for a logo or icon with crisp edges.

Aside from being hideous :) this is drawn to the new Gnome standards for scaling:

image

It's sharp at 64px:

image

...zoomed in to show pixellation:

image

...but not sharp at 48px:

image

...zoomed in to show pixellation:

image

Some icons look fine anyway, but it's always a gamble scaling an icon down, unless it was optimised for the size you're scaling to.  We tried scaling the 256px Suru icons to 48px and some of them were noticeably less sharp than the current 48px version, especially on some wallpapers.  I think the worst offenders were file manager and notepad.  I just think we'll have to rework the icons so they look as good as possible at key sizes including 48px before we can retire these sizes.

> Firefox has only one fullcolor icon, that don't follow guidelines, but looks good in dock.

I think it's just about what you're used to.  A bit of antialiasing doesn't look terrible at all (see above) unless you're used to seeing it _without_ the antialiasing.  For instance, here's the above graphic simply scaled down to 48px, and then specifically optimised for 48px:

**Scaled down:**

image

**Optimised:**

image

You might have thought the scaled-down version looked perfectly clear and sharp till you saw the optimised version :)  Now imagine you've seen the optimised version on your launcher every day, and then we suddenly switch to the scaled-down version.  You'll notice that it's not as sharp any more, and that's the problem we've got.  The 48px Suru icons are lovingly optimised for 48px and are 100% crystal clear on today's launcher.  If we change overnight to scaled-down versions of icons that weren't specifically optimised for 48px, people will notice it isn't quite as sharp any more, with some icons being more noticeably worse than others.

eaglersdeveloper:
Thank you, @ubuntujaggers.

So we need to switch to 96px icons, right?

eaglersdeveloper:
Or we can use 128px **vector**  icon?

ubuntujaggers:
> Or we can use 128px **vector** icon?

The issue still affects vectors sadly.  A vector icon will be as clear as possible at every size.  But if you show a 128px vector at 48px, a 4px border scales to one and a half pixels.  There's no way of showing half pixels without a bit of blur.  So "as clear as possible" will be better at some sizes than others.

Gnome are doing 128px vectors but recommending two sizes where they appear sharpest (64 and 32px).

If we want to target perfect sharpness at 48px, we can make new vectors that are 96px (double size) or 192px (quadruple size) and align the elements to a suitable grid.  Then everything will be aligned perfectly to the pixels when we display at 48px  :)

The only way you could make a 128px vector that scales perfectly to 64px *and* 48px is if you used a 8x8 grid, because that divides into whole numbers of pixels at both target sizes.

image

image

image

But it's too big a grid to make nice icons  :(  The borders and lines would have to be extremely thick.

PR #1247, opened 2019-03-08
Non-uniform usage-app and symlinks

ubuntujaggers:
Missed this one, but it's in app store as a Gnome app!  Will spend this weekend making sure I haven't missed any or made any mistakes before UI freeze.

image

image

image

image

Personally, I think the line is too fine on the full size version but it clones the Suru one exactly in that regard, we can add it to the list of things to improve in future and it isn't a step back at least.

Also: doing sudo ninja install on a clone of Master made me feel like the new app symbols are a big improvement (because reverting to Master made the top bar look worse) so would be really cool if we can include the new ones.  Apologies, it was my first go at symbols and I have learned some lessons  :)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1247`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1247` channel to test your snaps with those changes.

Feichtmeier:
Well, exactly what would have been expected! +1 For merging.@madsrh @clobrano ?

madsrh:
Yes :tada:

clobrano:
Merging this

PR #1248, opened 2019-03-08
Non-uniform root-terminal-app

ubuntujaggers:
Don't know if this is even used anywhere, but better safe than sorry:

image

No symlinks for this one.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1248`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1248` channel to test your snaps with those changes.

Feichtmeier:
Yep, looks like the terminal app but in purple! I think there can't be any veto - merging :)

PR #1249, opened 2019-03-09
Non-uniform weather-app and symlinks

ubuntujaggers:
Another one I missed because I thought it was only used on Ubuntu Touch, but I somehow didn't find Gnome's weather app last time I looked:

image

image

image

image

Glad we've got this in Yaru because it's a gorgeous icon  :)

**EDIT:** I know I'm trying to reduce overuse of transparencies in the symbolics, but this is one case where it works really well, so I won't change the symbolic:

image


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1249`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1249` channel to test your snaps with those changes.

Feichtmeier:
+1
@clobrano @madsrh ?

madsrh:
Yes!

PR #1253, opened 2019-03-11
Build snaps only for bionic branch

clobrano:
As per Gnome shell changes on Disco, we are going to freeze Bionic
development. For this reason, any future change for the snap package
will be build only if pushed on the dedicated `bionic` branch.
Moreover, the majority of PRs will be for master branch, so there is no
real need to build testing snaps for each of them.

Summing it up:
- Stop building testing snap for each PR
- Build and push on edge channel only merge on `bionic` channel


didrocks:
LGTM! Please ensure that gtk-common-themes also use the correct Yaru branch if needed. Otherwise: +1

clobrano:
`gtk-common-themes` uses for the branch name the same variable we test for "bionic", so this should be safe. However, in the previous commit, the snap was built for any push or PR from the ubuntu repo, so I needed to prevent that.

clobrano:
This should be good now, merging it

PR #1255, opened 2019-03-12
Update changelog

clobrano:


debian/changelog
```diff
@@ -1,3 +1,205 @@
+yaru-theme (19.04) disco; urgency=medium
```
iainlane: This version number is used - `19.04.1`?

PR #1281, opened 2019-03-19
Fix build on target

### LINKED ISSUE #1280: Build error when installing PNG (only when rebuild)
This error occurs only in a second build (first build works fine if there is no /user/share/icons/Yaru)

    FileExistsError: [Errno 17] File exists: 'preferences-desktop-font.png' -> '/usr/share/icons/Yaru/24x24@2x/categories/org.gnome.font-viewer.png'

probably due to #1274

clobrano:
- Add missing rendering of preferences-desktop-font png
- Remove symlink for org.gnome.font-viewer-symbolic

closes #1280


clobrano:
Hi @didrocks, 

I asked your review mainly to consider if this is needed in Disco before release date or, as discussed on IRC, a problem only for developers

Thanks!

didrocks:
That break us, we shouldn't be bothered on distro thanks, but thanks for asking!

Issue #1272, opened 2019-03-15
dark variant: no background in menu inside a popover


clobrano:
## Expected Behavior

Popover backround is not transparent

## Actual Behavior
![image](https://user-images.githubusercontent.com/2883614/54445379-ef523080-4744-11e9-8ec1-e0b1d9d8e4a8.png)

## Steps to Reproduce the Problem

  1. Install a virtual machine in gnome-boxes and setup shared folder (install spice-webavd in guest)
  2. Go to "devices & share" to select to add a shared folder: the popover menu has a transparent background

## Specifications

    $ git describe r29-37-g4baa9054

PR #1282, opened 2019-03-21
Prevent menu inside popover from having transparent background

clobrano:
Exclude menu inside popover from the wildcard of transparent background
(cherry-pick e163bf9e0e85d80b74882b050ccc867574c4f4e2)

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -2430,7 +2430,7 @@ popover.background {
   }
 
   @if $variant=='dark' {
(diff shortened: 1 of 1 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge`.

You may want to download as well `gtk-common-themes` snap from `edge` channel to test your snaps with those changes.

clobrano:
This requires someone with a 18.04 installed on a real machine, since the only way I know to reproduce it is to install a VM inside it with gnome-boxes (see #1272 for steps to reproduce it)

PR #1286, opened 2019-03-23
Improve treeview focus outline for white background

mzur:
The treeview focus outline is invisible on white background. This PR changes the outline to the same style than the [global outline](https://github.com/ubuntu/yaru/blob/6a4d98747160c9d28bb0fe1249d4117d37bb472c/gtk/src/light/gtk-3.20/_common.scss#L29).

Before: 
before

After:
after

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -466,7 +466,7 @@ entry {
 
 treeview, row {
   &:focus {
-    outline-color: transparentize(white, 0.2);
+    outline-color: gtkalpha(currentColor, 0.3);
     outline-offset: -2px;
     -gtk-outline-radius: $small_radius;
   }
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Nice, ty!
The only difference should be :selected state where it should be white on orange because the global one is hard to spot on orange. Could you change that?

mzur:
Done. It's the same white than before now.

Feichtmeier:
Thank you!

I think we need to wait for this PR to be merged until disco has been released :thinking: 
Since the UI freeze for disco happened some weeks ago

Feichtmeier:
We can start merging again. Could you also make a branch out of the bionic branch and then cherry pick this commit into that bionic based branch and make a PR into the bionic branch? :) @mzur

mzur:
[Done](https://github.com/ubuntu/yaru/pull/1315).

clobrano:
looks goot to me :+1:

PR #1296, opened 2019-03-29
Add legacy document-new/open/edit icon

### LINKED ISSUE #1294: Icons for "New", "Open" and "Save" are the same in KeepassXC
I think [this issue](https://github.com/snwh/suru-icon-theme/issues/108) actually belongs here.
## Expected Behavior
In KeepassXC the icons for New, Open and Save should be different like in the adwaita theme

## Actual Behavior
The icons are the same

## Specifications

name:      communitheme
summary:   The next Ubuntu theme built by the community.
publisher: Didier Roche (didrocks)
license:   unset
description: |
  Yaru, formerly known as Communitheme, is the new Ubuntu theme built by the
  community. Yaru will become the default Ubuntu theme in Ubuntu 18.10. This
  package allows you to try out the theme on Ubuntu 18.04 LTS.
  
  To try out the theme, install this package on Ubuntu 18.04 LTS, restart
  your computer and select the "Ubuntu with communitheme snap" session from
  the login screen.
  
  More information is available at
  https://community.ubuntu.com/t/faq-ubuntu-new-theme/1930.
snap-id:      Yd6CISPIf6tEf3ZEJ0cqSoEg9rG2VkRi
tracking:     stable
refresh-date: 13 days ago, at 19:41 CET
channels:
  stable:    0.1 2019-03-13 (1768) 16MB -
  candidate: ↑                          
  beta:      ↑                          
  edge:      0.1 2019-03-22 (1775) 16MB -
installed:   0.1            (1768) 16MB -

clobrano:
Some applications might use the (I guess) legacy names for the symbolic icons document-new, document-open, document-edit. In yaru those files are named document-new-symbolic, document-open-symbolic, document-edit-symbolic (which is the new name convention) and the application (see keepassxc) might not find them.

Adding new symlinks in symbolic/actions.list for fixing it

Screenshot from 2019-03-29 17-24-54

NOTE: the difference in style is evident. Our action icons are `symbolic`, while Keepassxc ones are colorful. We are following the new convention though, so I think is up to Keepassxc to convert the other custom icons.

closes #1294

### DIFF
--- a/icons/Suru/scalable/actions/document-new.svg
+++ b/icons/Suru/scalable/actions/document-new.svg
@@ -1,0 +1,1 @@
+document-new-symbolic.svg
--- a/icons/Suru/scalable/actions/document-open.svg
+++ b/icons/Suru/scalable/actions/document-open.svg
@@ -1,0 +1,1 @@
+document-open-symbolic.svg
--- a/icons/Suru/scalable/actions/document-save.svg
+++ b/icons/Suru/scalable/actions/document-save.svg
@@ -1,0 +1,1 @@
+document-save-symbolic.svg
--- a/icons/src/symlinks/symbolic/actions.list
+++ b/icons/src/symlinks/symbolic/actions.list
@@ -39,3 +39,6 @@ selection-end-symbolic.svg selection-start-symbolic-rtl.svg
 selection-start-symbolic.svg selection-end-symbolic-rtl.svg
 system-restart-symbolic.svg system-reboot-symbolic.svg
 window-pop-out-symbolic.svg detach-symbolic.svg
+document-new-symbolic.svg document-new.svg
+document-open-symbolic.svg document-open.svg
+document-save-symbolic.svg document-save.svg


droidmonkey:
Can you please point to where adding the suffix "-symbolic" has become the new convention/standard? It makes much more sense to have that suffix on the icon theme folder (eg, Suru-symbolic) with the icon names being left at the freedesktop.org standard. This affords maximum flexibility since you can also create a Suru theme that is full-color.

Feichtmeier:
@clobrano I think that's not a good solution. When the app wants a full color icon we should give them a full colour icon. I wonder why it didn't just fall back to humanity? 🤔

clobrano:
Yeah, I wasn't too happy myself with the result, to be honest. 

@droidmonkey
> Can you please point to where adding the suffix "-symbolic" has become the new convention/standard?

I meant standard in the gtk world, but I looked better and it's indeed only a suggestion not a real standard https://github.com/GNOME/adwaita-icon-theme

clobrano:
> @clobrano I think that's not a good solution. When the app wants a full color icon we should give them a full colour icon. I wonder why it didn't just fall back to humanity? thinking

that's a good point. Yaru doesn't even have an action folder once installed

droidmonkey:
@clobrano from the adwaita readme:

> Icons follow the naming specification, but have a -symbolic suffix, so only applications specifically looking up these symbolic icons will render them. If a -symbolic icon is missing, the app will fall back to the regular name.

The real fix for this issue is to fix the fallback.

PR #1306, opened 2019-04-22
Testing PR for possible new software icons

ubuntujaggers:
See if you like these...

image

image

image

image

image

image

**EDIT:** "Testing PR" because we can still try new things in this branch (e.g., the cog visible through an opening).


eaglersdeveloper:
Current icons looks better 😐

ubuntujaggers:
Is it the removal of the "A" you don't like (and the use of the coloured app grid symbol) or the redesigned backgrounds?

eaglersdeveloper:
Both.

Feichtmeier:
Thanks for the work!

I also think that the A-style is still a better fit than these, because those grid could be glued to the "open the app grid" symbolism - various android launcher also use that grid, and ofc the shell. 
However, they look very well executed :blush:

ubuntujaggers:
Could try the "A" on these? I definitely prefer software properties background to the current globe.

We could also keep the lunch box but improve the vertical alignment. A compromise was made so the "A" would be aligned across all three. But that's no longer needed because software store has been renamed and is no longer next to the other two on the app grid.

Paz-it:
TBH +1 with @eaglersdeveloper

ubuntujaggers:
I believe that's a no! :)

Any more ideas for a generic "apps" pictogram?

ubuntujaggers:
Maybe something like

https://images.app.goo.gl/uUdsUrAvfk4Cb1rFA

?

clobrano:
+1 for the new bag and I'm also in favor to explore more the boxes paradigm, maybe with something similar to the link above :+1:

ubuntujaggers:
For info, here's how it would look with the new shopping bag and no other changes:

image

image

Could just use this till we have a better glyph to indicate apps?

Feichtmeier:
Good idea!
Or, if that isn't too much work, try the Glyph you linked from a different website with that 4 squares and 1 of that being turned over.
We have time now :)

eaglersdeveloper:
I think that the A letter is the best pictogram.
The handle of bag looks somehow not real.

Feichtmeier:
An advantage of the round handle is that the whole icon looks less hard or squarish so it prbly reminds less of that foot box omgubuntu talked about

ubuntujaggers:
I think the more simple handle might be a bit more "theme" as well? E.g., for magnifying glasses, the handle is just a simple flat rectangle.

Feichtmeier:
The more I look at them the more I like it. I only really dislike the settings icon because it has an emblem :D

Feichtmeier:
@ubuntujaggers I think the most liked idea was the old A's with the improved BGs and shapes seen here:
https://github.com/ubuntu/yaru/pull/1306#issuecomment-486209650

Could you make another PR for this? I think this one has conflicts 😃 Or would you prefer to fix the conflicts?

ubuntujaggers:
Easier to make a new one and delete this branch - I'm away for the weekend but will do this one (because the icon's already done) as soon I'm reunited with my laptop on Monday.

PR #1307, opened 2019-04-23
Fix path-bar-box with width-maximized class

### LINKED ISSUE #1304: Nautilius - odd look to headerbar
## Expected Behavior

Resized naultilus window, and expected the "Home" button to stay far left, and the search button far right.

## Actual Behavior

When you resize nautilus, it looks like it has a max width for the pathbar+search icon and so tries to center these if the window is too wide. This looks OK with Adwaita, but **horrible** with Yaru.

Yaru:
Screenshot from 2019-04-18 19-26-06

Adwaita:
Screenshot from 2019-04-18 19-29-52

With adwaita, when the path bar is centered the background of the whole widget is drawn. With Yaru, it appears as if the "Home" and "search" icons are just placed randomly. There should at least be a border drawn to connect the two.

  1. Open nautilus
  1. Resize to sufficient width/

## Specifications

Yaru, and Ubuntu session, are from an upgraded 18.10 -> 19.04 install.

clobrano:
depends on https://gitlab.gnome.org/GNOME/nautilus/issues/992

---

Latest nautilus added a `.width-maximized` class style that is applied
to nautilus-path-bar when the windows is "big enough". In such situation, the
nautilus-path-bar changes from left-aligned to centered and it starts moving
together with the `search` button. Yaru did not show any connection
between nautilus-path-bar and search icon, so the two widgets looked floating
in the headerbar with no reason.

Add a path-bar border (only when `.width-maximized` class exists) that
starts from the right end towards the search button to show that a
connection exists (very much like Adwaita does).

Note: path-bar class (i.e. filechooser) did not changed

Closes #1304

image
image
image
image

Note^2: I noticed some glitches in my VM so that sometimes the path-bar extended border is drawn unstyled. Moving the windows in backdrop and in focus again solves it (eventually, open a bug upstream if confirmed)

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1976,8 +1976,6 @@ headerbar { // headerbar border rounding
     @include button(active, $headerbar_bg_color, $headerbar_fg_color, $flat:true);
     &:hover { @include button(hover, $headerbar_bg_color, $headerbar_fg_color, $flat:true); }
   }
-  &:not(:only-child):last-child { border-radius: 0; }
-  &:only-child { border-radius: $small_radius 0 0 $small_radius; }
 }
 
 .path-bar-box.width-maximized {
@@ -2006,12 +2004,12 @@ headerbar { // headerbar border rounding
       &, &:not(:first-child):not(:last-child), &:first-child, &:last-child { border-radius: 0; }
       box-shadow: none;
 
-      &:hover { 
+      &:hover {
         box-shadow: inset 0 -2px transparentize($insensitive_fg_color, 0.5);
         background-color: transparent;
-        color: $insensitive_fg_color; 
+        color: $insensitive_fg_color;
       }
-      
+
       &:active { box-shadow: inset 0 -2px $insensitive_fg_color; }
 
       &:checked, &:checked:active {
@@ -2026,8 +2024,8 @@ headerbar { // headerbar border rounding
         color: $backdrop_insensitive_color;
         border-color: transparent;
         box-shadow: inset 0 -2px transparent;
-        &:hover { 
-          box-shadow: inset 0 -2px transparentize($insensitive_fg_color, 0.5); 
+        &:hover {
+          box-shadow: inset 0 -2px transparentize($insensitive_fg_color, 0.5);
           background-color: transparent;
           color: $backdrop_insensitive_color;
         }
@@ -4720,4 +4718,6 @@ button.emoji-section {
  * will be overwritten */
 .path-bar-box.width-maximized > .nautilus-path-bar button {
   margin: -1px;
+  &:not(:only-child):last-child { border-radius: 0; }
+  &:only-child { border-radius: $small_radius 0 0 $small_radius; }
 }
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1744,6 +1744,7 @@ headerbar {
       &:not(:last-child) { margin-right: 4px; }
     }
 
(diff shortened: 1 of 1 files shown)


Feichtmeier:
It's not possible to only axe the border radius of the last child not only child buttons when they are in the maximized width pathbar, right?

clobrano:
Yes it is possible, but very prone to this "glitch" I was talking about before, at least on VM and the same happens with Adwaita

soon after maximizing the window
image

after a cycle of backdrop-focus-again
image

clobrano:
There's a bug already https://gitlab.gnome.org/GNOME/nautilus/issues/992

Feichtmeier:
:S
That's odd. We could keep it like you did in your last commit and hope it gets fixed in this cycle. If it doesn't we could discuss a different styling

clobrano:
Unfortunately also the path bar extended is affected. We could merge this, waiting for upstream fix

Feichtmeier:
Shall we keep bionic styling as it is?

clobrano:
I'd say yes

Feichtmeier:
Let's wait for @madsrh and @ubuntujaggers - if they like it too we should merge it

madsrh:
Looks good @clobrano 👍 

You can't really tell by looking at the UI that the long rectangle is clickable/writeable IMO. It's much more obvious in Adwaita because of the background color.

Feichtmeier:
Offtopic joking:
You dudes are seriously becoming too old! :older_man: :older_man: :heart: 
I am happy that the long road of inconsistent buttons in the headerbar has an end with this :dancer:

madsrh:
> You dudes are seriously becoming too old!

🙉 

I just mean, we have a background here for a reason:
image

image

Feichtmeier:
Ah! Yeah that are entries. 
The widget in the headerbar in Nautilus is a GtkStack which includes the pathbar and and entry. With the search toggle button you switch between the stack pages.
So I would say that the pathbar, which includes buttons does not need to be styled as entries

Feichtmeier:
@clobrano
Just tested it on my machine - looks really, really good great job!
But there is a double border around the buttons and the pathbar :( it looks like this is connected to the focus issue you linked, right?

clobrano:
Yes, it should go unfocus/focus the window

Feichtmeier:
Damn :/ that border looks really shitty
If it would be only the border radius of the last button it would be tolerable but with the double border it looks too broken imho :/ can we wait until they fixed it Upstream?

clobrano:
There's no hurry indeed, since point release is way ahead. Let's see if the bug progress and I'll try to have a look at it

Feichtmeier:
@clobrano sorry for the conflict!
But good news, we only need 1px padding and it works :)
image

clobrano:
So it seems, nice catch! I'll revert the last commit and add the padding only (just saying, but the borders on linked flat buttons doesn't seem very nice)

Feichtmeier:
Nice - now we are synced with buttonboxes and stackswitchers in the headerbar, thank you!
grafik

Feichtmeier:
Merging after okay from carlo

DragoCubed:
does this mean I'll have to wait for Ubuntu 19.10? I'm a complete noob btw

clobrano:
Yes, unless Canonical schedule an upgrade for Yaru in 19.04

DragoCubed:
😨😰💀

I thought Yaru could update independently

clobrano:
In 18.04, where it's installed as snap, it updates independently

DragoCubed:
If Yaru can update independently. why should we wait for a release? It's been merged but I have to wait four months. That doesn't make sense to me.

PR #1309, opened 2019-04-24
New status icon (also emblem) for dialog-error

### LINKED ISSUE #1302: Missing status icon "error" / "dialog-error"
In the report a problem app / aport app there is some legacy kind of emblem being applied...

image

Yet I can't find that specific icon in humanity or ubuntu-mono-dark icons :thinking: 
Anyone has idea where aport sucks that icon from?

Edit 2: found it, it is a status icon (-.-) 

Screenshot from 2019-04-18 09-58-17

ubuntujaggers:
Closes #1302:

image

Not sure how to recreate the conditions to test it - as ever, happy to change further if it doesn't look good live  :)


Feichtmeier:
Thank you!!!
Could you also add a symbolic version? 
I think then it should become monochrome in the gnome shell panel
O:)

ubuntujaggers:
Oh weird, @Feichtmeier  - there's already a status --> dialog-error in symbolic, which is actually a different glyph to the Humanity one:

image

Not sure why this isn't being used then?  I was just going to add a file of this name to this group, but it's already present and not being picked up?

Feichtmeier:
I need to find an app that crashes xD
To test with you

madsrh:
Nice @ubuntujaggers 😃

Feichtmeier:
@ubuntujaggers full color icon works, and tbh it is very nice to be red in the status bar since it is a rather heavy error:

image

ubuntujaggers:
Thanks for checking!

Feichtmeier:
@ubuntujaggers and it is also working in the apport app where it is taken as an emblem ^^
image

PR #1316, opened 2019-04-25
Test PR: Toned-down folder icons (1)

ubuntujaggers:
Before:

image

After:

image

Glyph folders:

image

Thoughts?


madsrh:
I haven't tried this yet, but I'm a fan 😁 This looks much more balanced IMO!

clobrano:
Tried and :+1:, it doesn't change a bit the Ubuntu way :)

Feichtmeier:
Do we still need the backdrop opacity hack then?
And it looks nice

madsrh:
> Do we still need the backdrop opacity hack then?

IMO the backdrop is unrelated to the folder icon color.

ubuntujaggers:
Do you chaps also want to see the second suggested tones?

madsrh:
> Do you chaps also want to see the second suggested tones?

If it's easy as pie 🥧 then yes that would be great

PR #1320, opened 2019-04-26
Make category symbols in Software Store consistent with Yaru

ubuntujaggers:
This doesn't close the issue because there are more to do, but it's enough changes to ensure that every category in the Software Store has a Yaru symbol rather than a fallback one.

I'm not sure it looks brilliant because the style of the Software Store is very "chunky" (bold font etc.) and these symbols are not chunky.  But a couple of the categories were already using a Yaru symbol from "apps" or "status", so there's no way of using "chunky" symbols in Software Store without the same symbols being chunky elsewhere in the desktop.  At least this way the symbols are consistent with each other and the rest of the theme.

image


clobrano:
I think we all agreed, merging :+1:

PR #1322, opened 2019-04-26
Test PR: Toned-down folder icons (2)

ubuntujaggers:
This branch uses tones taken from the second suggestion by @madsrh.

image

image

image

BTW, if anyone wants to experiment with recolouring the folders svg, I found a quick way of doing it.  An svg is stored as markup so you can open/edit it in notepad.  There are just three colours to change, and you can use find and replace to change them as text strings.  They are:

fb7c38 - main gradient start

fda463 - main gradient end

f34f17 - the darker colour for the inside of the folder.

So once you pick your colours, it's just 3 x find/replace in notepad and then you can load it in Inkscape to see the result  :)  Leave off the last two hex characters because those are opacity (e.g. f34f17**ff**) which has its own field in the markup.


clobrano:
I prefer #1316, this one looks a bit too yellow

madsrh:
> I prefer #1316, this one looks a bit too yellow

I agree

madsrh:
Thanks for making the PR for testing @ubuntujaggers

PR #1325, opened 2019-04-29

didrocks:
gtk-common-themes snaps.


didrocks:
bionic and master CI passing. Merging (and older variable removed from Travis)

Feichtmeier:
Thank you!!

PR #1326, opened 2019-04-29

didrocks:
gtk-common-themes snaps.


didrocks:
wrong targetted branch, sorry

PR #1327, opened 2019-04-29

didrocks:
gtk-common-themes snaps.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1327`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1327` channel to test your snaps with those changes.

didrocks:
bionic and master CI passing. Merging (and older variable removed from Travis). Please rebase any bionic PR on bionic branch now to get this fix working.

PR #1330, opened 2019-05-01
Remove bordercolordifference in darkthemeswitches

Feichtmeier:
When the border color of the top border is different than the rest of the border, there are dark and washed out edges in the border top where a darker green shines through the lighter green. Probably because gtk lets the background color shine through the green. which is no problem in the light theme but in the headerbar and the darktheme. We changed this already for the headerbar some months ago, this commit does the same for the dark theme in general.

Before:
image
image

image

After:
image
image

image

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1657,10 +1657,7 @@ headerbar {
     $c: $headerbar_bg_color;
     $tc: $headerbar_fg_color;
 
-    switch {
-      @include switch(lighten($headerbar_bg_color, 25%));
-      &, &:backdrop { &, &:hover, &:checked, &:checked:hover { border-top-color: transparent; box-shadow: none; } }
-    }
+    switch { @include switch($headerbar_bg_color); }
 
     entry, %dark_entry {
       @include entry(normal, $c: $c, $tc: $tc);
--- a/gtk/src/light/gtk-3.20/_drawing.scss
+++ b/gtk/src/light/gtk-3.20/_drawing.scss
@@ -701,7 +701,7 @@
 
   slider {
     background-color: if($bg == $dark_fill, white, $headerbar_fg_color);
-    background-clip: padding-box;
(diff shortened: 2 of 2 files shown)


clobrano:
I agree on removing the top shadow, but why there is any difference between light and dark variant here? On headerbar they should be the same

Feichtmeier:
Oh you are right! Sorry, I look up my old commit and do the same for the dark theme!

Feichtmeier:
I also removed the border-top color difference, though in a stupid way by setting it to transparent:

https://github.com/ubuntu/yaru/pull/1139/commits/f04df32d6cf21c588ef98b33e1fb778a347adef6#diff-1b30fa3f60328fce0c66d39158906ac1R1652

> but why there is any difference between light and dark variant here

Because the dark background of the dark theme shines through the switch.
Actually I don't see many different ways here if we want a sharp switch in the dark theme. 
But my previous PR was a little bit strange, since the problem is the top border and not the box-shadow.

clobrano:
What I meant is that I noticed that the swtich is different in dark and light, not just for the border or shadows

Screenshot from 2019-05-02 17-21-38
Screenshot from 2019-05-02 17-21-59

specifically, the nob is smaller and less flat and I wondered why there is any difference since the css should be the same. I'll have a deeper look though

Feichtmeier:
Because I've set the border-top to transparent, which makes no sense. I push a fix

Feichtmeier:
@clobrano and/or ( @madsrh and/or @ubuntujaggers ) please test it again

The difference of the knob came from border-box / padding-box styling
If one/all/at least two of you prefer the bigger knob it would be border-box, padding-box if you prefer the smaller.

Moved the code out of common into drawing and
removed the border-color "bug" for the dark theme shown above

clobrano:
maybe a +1 for the bigger nob, but not important, now it looks good :+1:

Feichtmeier:
Bigger knob is closer to the shell switch so this might be an argument for the bigger knob

PR #1346, opened 2019-05-06
Reduce gtk and shell theme tooltip differences

Feichtmeier:
Reduce height for the gtk theme tooltips
Reduce border radius for the shell theme tooltips

Before:
image
image

After:
image
image

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1463,7 +1463,7 @@ StScrollBar {
   }
 
   .dash-label { //osd tooltip
-    border-radius: $medium_radius;
+    border-radius: 5px;
     padding: 6px 12px;
     color: $osd_fg_color;
     background-color: $osd_bg_color;
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -4271,7 +4271,7 @@ tooltip {
   decoration { background-color: transparent; }
 
   > * { // Yeah this is ugly
-    min-height: 32px - (6px - 1px) * 2; // tooltip has hard-coded 6px padding
+    min-height: 30px - (6px - 1px) * 2; // tooltip has hard-coded 6px padding
     padding: 0 8px - (6px - 1px);       // and does not consider border-width
     background-color: transparent;
     color: $porcelain;


Feichtmeier:
I don't know I think these were the wrong screenshots xD
Let me check again

clobrano:
It doesn't seem to work. In below screenshot "show other location" is the gtk tooltip, "ubuntu software" is from shell tooltip from master and "boxes" from this PR

image

Looking at the code, if this comment is still true

    // tooltip has 6 hard-coded 6px padding

I wonder if it is even possible to fix, since gtk tooltip seems to have less than 6px padding

Feichtmeier:
The minheight Def. Works. You can play around with it :D but it still doesn't match the shell height

clobrano:
Uhm, why "reduce" gtk tooltip height then if shell one is taller? :thinking: 

Unfortunately I don't remember the link, I can't send you the actual file

Feichtmeier:
No it was the other way round. 
In master:
The gtk tooltip is bigger
The shell tooltip has a higher border radius

clobrano:
This is weird. In master it looks like gtk has a 32px height, while shell has 35px

master-tooltip-gtk

master-tooltip-shell

With this patch, they're the same
image

```
carlo@lenny· workspace/yaru (master)~$ gd
diff --git a/gtk/src/light/gtk-3.20/_common.scss b/gtk/src/light/gtk-3.20/_common.scss
index 8d8cdb6a..beea7746 100644
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -4271,7 +4271,7 @@ tooltip {
   decoration { background-color: transparent; }

   > * { // Yeah this is ugly
-    min-height: 32px - (6px - 1px) * 2; // tooltip has hard-coded 6px padding
+    min-height: 35px - (6px - 1px) * 2; // tooltip has hard-coded 6px padding
     padding: 0 8px - (6px - 1px);       // and does not consider border-width
     background-color: transparent;
     color: $porcelain;
carlo@lenny· workspace/yaru (master)~$

```

Feichtmeier:
Indeed weird. Maybe I was drunken. 
Checking back later with and without your patch

Feichtmeier:
Okay let the  crazy gimp times begin! xD
I made two screenshots of my desktop when hovering above the shell and nautilus and populating tooltips in both
Screenshot from 2019-05-07 22-03-21
Screenshot from 2019-05-07 22-03-14
Then I put them into gimp as layers and just cut and moved a part of the shell layer close to the gtk tooltip and the shell tooltip is def. less high
image

clobrano:
Are you on bionic?

Feichtmeier:
Okay, Double checking that later and then revert everything except the border radius adaption thanks for the help!

clobrano:
> Okay, Double checking that later and then revert everything except the border radius adaption thanks for the help!

Okay, but have you thought about changing the value of shell `$medium_radius` instead that using a custom value only for tooltips? It should be more cohesive for the whole shell theme and being just a little variation should not affect too much the other widgets

Feichtmeier:
Oh, that might cause some side effects. That variable is used all over the place : / I would rather dodge this

Doubled checked it on master in disco... and the gtk tooltip is def. higher:

image

Feichtmeier:
And with the current commit of this branch:

image

clobrano:
Okay, I got the error, I have the `scaling factor` to 1.20, with normal scaling factor gtk tooltip is actually bigger. Eventually this is a problem with scaling factor, not this pr :+1:

clobrano:
Sure

PR #1348, opened 2019-05-08
shell login: darken login entry

### LINKED ISSUE #1342: Suggestion: Login screen password box
The white in the password box and cancel button doesn't really fit with the rest of the login screen or the topbar, the dark box as seen in Ubuntu 18.04 without Yaru fits MUCH better.

Current:
alt text

Dark boxes, like 18.04:
alt text

clobrano:
Darken login entry with a style closer to 18.04

image

closes #1342

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1972,7 +1972,8 @@ StScrollBar {
 
   StEntry {
     &:focus {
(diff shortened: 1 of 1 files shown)


Feichtmeier:
If we don't lighten up the purple bg, the contrast of that inkstone buttons and the entry to the very dark purple is too low.

clobrano:
I have problem with my VM so I cannot give a screenshot

PR #1360, opened 2019-05-13
Bionic: shelltheme gdm

Feichtmeier:
Cherry-picking and commiting "manually", to adapt the bionic branch to https://github.com/ubuntu/yaru/tree/ShellTheme_GDM

We/I need to test this on our bionic installations, I hope I did not mess up any merge conflict handling

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1370,7 +1370,7 @@ StScrollBar {
   }
 
   //search entry
-  .search-entry {
+  .search-entry, %search_entry {
     width: 320px;
     border-radius: $small_radius;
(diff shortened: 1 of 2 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1360`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1360` channel to test your snaps with those changes.

Feichtmeier:
Works, but I don't know if the 2px border is necessary for the squircle :thinking:

clobrano:
It doesn't feel bad though, I let you decide

Feichtmeier:
Okay let's keep them as similar as we can then

Paz-it:
What a nice surprise I've got! I like it very much, thank you!!!

PR #1362, opened 2019-05-14
row: no difference between pressed and pressed+selected row

clobrano:
In sidebar, selected row has a stronger pressed effect and an orange
left border which does not seem necessary.

Before
image

After
image


Feichtmeier:
Agreed! The looks like ultra pressed xD

Feichtmeier:
Could you 🍒 pick this into bionic?

clobrano:
oops, I think it went directly into bionic without the PR...

PR #1368, opened 2019-05-17
common: fix 1px gap in sidebar

### LINKED ISSUE #1363: 1px gap in gnome control center sidebar
There is a 1px gap between the row right side and the edge of the sidebar in Yaru

Yaru:
image

Adwaita:
image

Maybe I introduced this with that hdyleaflet stuff... maybe not.
It is not happening in nautilus or tweaks
@clobrano  any idea?

clobrano:
When a viewport is inside a scrolledwindow there is 1 px gap due to
separator's min-width visible only when the row is selected, since
otherwise viewport and separator have the same bg color.

To fix this, let separator's min-width be zero.

Screenshot from 2019-05-17 17-47-12

closes #1363

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -3731,7 +3731,7 @@ scrolledwindow {
 //vbox and hbox separators
 separator {
   background: transparentize($borders_color, 0.7);
-  min-width: 1px;
+  min-width: 0px;
   min-height: 1px;
 }
 


Feichtmeier:
Niceeee! Thank you!

clobrano:
Indeed deleting the line with `min-width` has the same effect, but I preferred to be explicit here and set it on purpose.

clobrano:
Bionic doesn't seem affected, could you confirm?

Feichtmeier:
Confirmed!

PR #1376, opened 2019-05-27
Make headerbar border flexible

Feichtmeier:
By introducing headerbar_border_color
and making it dependent on headerbar_bg_color
This way the border changes when the headerbar_bg changes without changing the previously designed color visibly

Inkstone:
INSTONE

lighten($headerbar_bg_color, 8%)
lightenhbbgcolor8

Difference is barely visible:
difference


clobrano:
Looks good to me :+1:

Issue #68, opened 2018-01-16
Toolitem entry highlight goes out of border to the next widget


clobrano:
![image](https://user-images.githubusercontent.com/2883614/34996552-41c05a30-fada-11e7-8570-dcc2972e02d9.png)

Not sure it is the best description, so I took a picture. This happen in Document Viewer, on top-left widget

godlyranchdressing:
Most likely a result of unlinking headerbar elements so the buttons can be flat (to fix the problem with GNOME Dictionary's linked button and entry field). I didn't think an entry would be linked to an entry, though, that's the problem there. 

I think I know how to fix it, but rogue issues like this might still pop up. Even though I'm liking the flat headerbar buttons more than I thought I would, I say we just go back to linked headerbar buttons not being flat to avoid this trouble, but we should hear what Mads and Stefan have to say first.

clobrano:
If there is no way to fix it everywhere, we should consider that gnome-dictionary is less used than document viewer

galgalesh:
Also consider that application developers specifically opted to link two elements. I think the theme should respect that decision..

I really like this btw:

![image](https://user-images.githubusercontent.com/1492981/35097576-ffb09d6c-fc50-11e7-896b-fa65b7170a21.png)

Still semi-flat but the association is more clear. Is it possible to do something similar for linked elements? (sorry if I'm asking/proposing stuff that is already discussed)

PR #1377, opened 2019-05-27
Fix hb entry border bug

Feichtmeier:
The border of linked widgets is gone when you focus the headerbar entry 

unfocused_bugged
Screenshot-20190527113506-186x152
Screenshot_20190527-152639_Samsung Internet

This removes the separation
hbentryunfocusedfixed
hbentryfocusfixed

@galgalesh was probably right here
https://github.com/ubuntu/yaru/issues/68#issuecomment-358628766

Eventually caused by https://github.com/ubuntu/yaru/pull/197

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1725,7 +1725,9 @@ headerbar {
 
     .stack-switcher.linked button,
     buttonbox.linked button,
-    buttonbox.linked button.text-button ~ button {
+    buttonbox.linked button.text-button ~ button,
+    box.linked button,
+    box.linked entry {
       border-style: solid;
       border-color: lighten($inkstone, 5%);
 
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1723,27 +1723,6 @@ headerbar {
       }
     }
 
-    .linked:not(.path-bar):not(.stack-switcher):not(buttonbox) button, .linked:not(.vertical) entry {
-      // force linked buttons and entries to be separate
-      // this needs a lot of testing
-      &:focus {
-        & + entry { border-left-color: _border_color($graphite); }
-        & + button { border-left-color: transparent; }
-
-        & + entry,
-        & + button {
-          &:disabled { border-left-color: transparent; }
-          &:backdrop { border-left-color: _border_color($headerbar_bg_color); }
-        }
-      }
-      border-width: 1px;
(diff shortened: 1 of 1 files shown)


clobrano:
The fix is good, and I agree that linked widgets should look linked.
However this is an "important" step, since the design decision to keep them separated was taken at the very beginning for the reason that flat linked widgets - not having borders - reveal to be linked only when highlighted.

For example

Nautilus now
image

and after this patch
image


/cc @madsrh, @ubuntujaggers

Feichtmeier:
I wonder, did this somehow cause some of the problem we had with border-box and other headerbar widgets?

clobrano:
It is possible. At the beginning we had many problem with this, but it seemed the only way to have completely flat headerbars

madsrh:
I'm +1 for this 👍 

[This example](https://github.com/ubuntu/yaru/issues/68#issuecomment-496194123) is a no-brainer IMO, but I wonder why we choose otherwise in the first place 🤷‍♂

clobrano:
Okay, let's see then if this is still okay for the rest of applications or needs some more fixing here and there, then merge it

Feichtmeier:
That "glade issue" comes to my mind, maybe some of the fixes you did back then are no longer needed now?
Those buttonboxes are also ".linked":
image

clobrano:
Possible ofc, have you tested glade with this branch? How does it look?

Feichtmeier:
No, sorry. 
Looked back in the issue and that's not really related sry. Was related to the old stackswitcher and buttonbox styling

Feichtmeier:
Simple-scan is also affected by this

clobrano:
Do we merge it?

Feichtmeier:
I can't imagine any negative impact - but I did not explicitly test this branch on different apps yet.
You decide 👼 😇

clobrano:
> I can't imagine any negative impact - but I did not explicitly test this branch on different apps yet.

LOL, it's your PR :D, let's test it a bit

Feichtmeier:
Tested successfully with

- [ ] nautilus
- [ ] eye of gnome
- [ ] gnome control centre
- [ ] rhythmbox
- [ ] gedit
- [ ] gnome disks
- [ ] totem
- [ ] deja dup
- [ ] simple-scan
- [ ] file roller
- [ ] gnome logs
- [ ] glade

clobrano:
Sorry, my bad, but I didn't notice the borders in all those linked flat buttons. It doesn't seem nice to me, could we revert it?

Feichtmeier:
Yes sorry, revert it! :)

PR #1384, opened 2019-05-31
Issue1380/fix hdyleaflets

### LINKED ISSUE #1380: Hdyleaflet selection mode not styled
image

We need to dig more into that hdyleaflet stuff and check how adwaita does it, since more and more gnome apps are going to have that desktop/mobile interface

clobrano:
Rework of PR #1382 

- [x] headerbar color
- [x] hdyleaflet headerbar separator
- [x] selection mode buttons
- [x] titlebuttons
- [x] standard separator
- [ ] border-radius on small window
- [x] hydleaflet glitch

closes #1380

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1654,6 +1654,31 @@ headerbar {
     }
   }
 
+  &.selection-mode {
+    separator.sidebar {
+      &, &.vertical, &.horizontal {
+        border-color: mix(white, $success_color, 30%);
+        &:backdrop { border-color: mix(darken(white, 15%), _backdrop_color($success_color), 40%); }
+      }
+    }
+
+    hdyleaflet headerbar {
+      button.titlebutton {
+        $c: $success_color;
+        &:not(.close):not(.appmenu) {
+          &:hover, &:active { @include draw_circle($c); }
+        }
+
+        &.close {
+          &,
+          &:hover:backdrop,
+          &:active { @include draw_circle($c, $close:true); }
+          &:backdrop { @include draw_circle(_backdrop_color($c), $close:true); }
+        }
+      }
+    }
+  }
+
   & {
     // style headerbar buttons and entries using the appropriate headerbar colors
     $c: $headerbar_bg_color;
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -1760,7 +1760,8 @@ headerbar {
     }
   }
 
-  &.selection-mode {
+  &.selection-mode,
+  .selection-mode & {
     $c: $success_color;
     color: $selected_fg_color;
     border-color: _border_color($c);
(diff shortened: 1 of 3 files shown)


clobrano:
Some pictures of current work

image
image
image
image

image
image
image
image

Feichtmeier:
Awesome! Thank you!

There is still a little glitch in hdyleaflet selection mode when first child and last child are both visible
grafik
:D

Feichtmeier:
One more thing: the orange circle is drawn on hover and close on the green, I would drop it completely in selection mode. Orange on the green is a hard combination :D

clobrano:
> One more thing: the orange circle is drawn on hover and close on the green, I would drop it completely in selection mode. Orange on the green is a hard combination :D

I actually keep it on purpose. I agree that having the orange always visible is too much, but on hover it seems a good balance.

Working on the other glitch

Feichtmeier:
Can't we go with transparency white circles instead? Really weird colour combination

clobrano:
Transparency looks a bit too soft for me. Does orange on hover really too much? :thinking:

Feichtmeier:
Transparentize white 0.9 looked okay to me. But I don't want eh... How was that metaphor... Let us die on this hill? XD

Feichtmeier:
It's not the important. At least the orange stays orange and thus it would make sense.

Feichtmeier:
@clobrano the issue is not only in selection mode :(
grafik

clobrano:
I see, this is why I kept the checkbox unchecked. I am about to push a fix for this, but the `border-radius` of shrinked hydleaflet is still unsolved.
I saw your PR used `first/last/only-child`, but the same didn't work here - so far - and looking at the inspector the `hydleaflet headerbar`s actually do not disappear from CSS nodes, they're just invisible, so the property `first/last/only-child` doesn't actually change

Feichtmeier:
Fucking horror!
Let's see what happens in adwaita, maybe this helps?

Feichtmeier:
Very nice with the box-shadow!
I noticed now the backdrop has become a lot sharper! Also the preview of the dash to dock doesnt have a ultra big scaled top border anymore yay
But yeah the border-radius of the first-child is crap again :(

Feichtmeier:
Also the border-radius is not axed for maxed windows
grafik
And the box shadow also stays

clobrano:
Yep, I imagined the problem with maximized windows only last night after
pushing the change, sorry. It's a bit of a mess, adwaita has nothing
explicit in master to handle hydleaflet.

clobrano:
I will work on this upcoming week, if I can't find a better solution will merge this with a know bug

clobrano:
I still could not find a solution for the border-radius of hdyleaflet when shrinked.The problem is that from a CSS nodes point of view nothing change, so we cannot adapt.
Testing solutions, I noted that basically the only way to fix it is to get rid of the top-highlight, which is the element that makes this glitch visible. The other possible solution - move the top-highlight to at higher level in "window decoration" - fixes the problem, but at this point the top highlight cannot adapt to different headerbar colors (i.e. top-highlight stays grey on green headerbar), which is not acceptable from my point of view and also way more visible than squared border-radius visible only with shrinked hdyleaflet.

Feichtmeier:
Closing this, as it is not happening anymore with new upstream refresh in master

PR #1385, opened 2019-05-31
Proposal: move down arrow on app icons so it doesn't clash with real emblems

ubuntujaggers:
This isn't my most exciting work  :)  and I know we want to do something more interesting for Ubiquity at least...  but, until we have better icons, how about simply moving the down arrow?

**USB creator**

Current:

image

Proposed:

image

image

The two emblems together won't win any design awards - **but** at least it isn't "broken"?

I don't know how to test Ubiquity  :(  but the down arrow is positioned exactly the same, relative to the USB creator.

SVGs:

image

image

Personally, if Ubiquity works in this branch, I think it's better to move the down arrow now, and maybe open an issue for "Ubiquity isn't very good" - so we can improve it when we've got something better.  At least that way the UI bug doesn't exist in master any more.  If you'd rather leave it as it is until we've got a better design then feel free to close  :)


madsrh:
👍 I'm +1 for merging this now. It's a good temporary solution and I don't have a better suggestion for Ubiquity atm. IMO it would be silly not to merge this since you've already done the PR.

clobrano:
I agree with you, it's good! Did you have tried already with an icon half
pale, to show that the USB is "in progress"?

Feichtmeier:
Good idea! 
Theoretically it's possible to have 4 emblems on an icon, for example locked, no rights, syncing, people, but not for ubiquity and the usb creator!
I agree with mads

Feichtmeier:
🍒 pick this to bionic please :)

ubuntujaggers:
@clobrano I've sat down and tried to do that design a few times because I like the concept :) but nothing usable so far. I haven't quite given up. The challenge for me is getting it to work at the smaller sizes.

PR #1392, opened 2019-06-07
Revert "Fix hb entry border bug"

Feichtmeier:
Reverts ubuntu/yaru#1377


clobrano:
Have you reverted the whole PR? I was talking only about the borders :thinking:

Feichtmeier:
Oh. Sorry :(
Whatever you want, we do it. Can't think currently.

clobrano:
Accepting this now, we can reapply the right part later

PR #1399, opened 2019-06-10
Sharpen backdrop headerbar border color

Feichtmeier:
By setting it to transparent. Previously it became blurry in the backdrop.
@clobrano please see if you like it - if not just close it and let's try other stuff!

Before:
grafik

After:
grafik


clobrano:
Looks good to me :+1:

PR #1403, opened 2019-06-11
Symbolic icon for Livepatch

ubuntujaggers:
I can't test this one myself because Livepatch isn't available for my system - I googled it, and I think you have to be on LTS or certain releases?  When I try to run Livepatch it just opens the normal software updater.

Here it is - it's called livepatch-symbolic.svg which I hope will work!!

image


Feichtmeier:
I'll cherry pick it into bionic and see if it works :)

Feichtmeier:
@ubuntujaggers hmmm sometimes symbolic icons are strange OR we just don't understand how they work xD

After testing this PR
grafik

Switching to ubuntu-mono-dark / adwaita / humanity icons:
grafik

So there IS already an icon for this :man_shrugging: 
It is called "livepatch-on.svg"
grafik
Maybe we should rather change that full color SVG on launchpad in the Ubuntu repos instead of here? Then we could leave the symbolic icon as it is in master

Issue #1311, opened 2019-04-24
Missing icons: categories


Feichtmeier:
![image](https://user-images.githubusercontent.com/15329494/56680287-498ccc80-66c7-11e9-9feb-bfa1984515fc.png)

Our categories:
![Screenshot_20191105-091836_Samsung Internet](https://user-images.githubusercontent.com/15329494/68190424-a7bb3600-ffad-11e9-8311-a50758815e77.jpg)

Adwaita's categoties, including emoji categories:
![image](https://user-images.githubusercontent.com/15329494/56680626-e5b6d380-66c7-11e9-8911-f46315aa3853.png)

All icons not viewable in gnome-software are pretty optional and kind of "bonus" if you have time @ubuntujaggers :)

(+ "-symbolic" excluded)
- [x] preferences-system
- [x] system-help
- [x] applications-utilities
- [x] applications-science
- [x] applications-engineering
- [x] preferences-other
- [x] applications-games
- [x] applications-graphics
- [x] applications-multimedia

Edit: removed emojis from the list as this is pointless to redraw

ubuntujaggers:
There are a few oddities here - the glyphs in the software store categories come from a mix of places (e.g. the communication/news symbol is taken from "status" and the productivity one is taken from "apps").  In my working branch, I've got them all showing as Yaru designs now:

![image](https://user-images.githubusercontent.com/38893390/56776118-622ddd00-67c2-11e9-8ddb-0aafa4e0b713.png)

...but I welcome thoughts on whether this represents an improvement here, because the 1px grey lines look a bit insubstantial next to the bold dark text.

Also: I was going to change the cog under "categories" to match the one we have in apps for settings, but I think the categories one actually looks better in the shell.  So I changed the apps one to match the categories one instead.

Current cog in the shell:

![image](https://user-images.githubusercontent.com/38893390/56776182-b20ca400-67c2-11e9-9d09-9e254a774109.png)

...and this is what it looks like if we make the settings-app symbol the same as the cog under categories:

![image](https://user-images.githubusercontent.com/38893390/56776195-cea8dc00-67c2-11e9-8706-30ab68bb0b2d.png)

Again, I welcome thoughts on this change!

clobrano:
New symbols in software look very nice, good work!

> ...but I welcome thoughts on whether this represents an improvement here, because the 1px grey lines look a bit insubstantial next to the bold dark text.

From my point of view, those thick colored lines look bad in any case, but we can't change them :)

[...]

PR #1414, opened 2019-06-21
Missing category icons: applications-science-symbolic and application-x-appliance-symbolic

ubuntujaggers:
I tracked down two of the missing icons in the most recent screenshot from #1311.

The "Devices and IoT" category is using something called _Mimetypes --> applications-multimedia-symbolic_ and Science is using _Categories --> applications-science-symbolic.svg_.  So, here are 1px stroke version of both icons:

image

image

I think I have to cherry-pick this, but do we merge into Master first?


Feichtmeier:
Thanks! You can either cherry pick first or merge first however you want :)

ubuntujaggers:
Let's merge because then I know exactly what I'm doing  :)  Can I press it myself?

PR #1415, opened 2019-06-21
Full colour version of drive-removable-media-symbolic

### LINKED ISSUE #1310: Missing icon: full color drive-removable-media
image

It looks like the full color version of drive-removable-media-symbolic.svg is missing :microscope:

ubuntujaggers:
Closes #1310.  Not sure it goes with the other full colour devices (well - it goes better than the weird printer - which I should really redo), but it's a full colour version of the existing symbol at least:

image

image

Can't test because I don't have a removable media  :)


Feichtmeier:
Awesome! On fire again? :dancer:

Feichtmeier:
@ubuntujaggers 
works!
image

You can test it when you use the shared folder feature of virtual box :) thank you!

ubuntujaggers:
That doesn't look too bad! Should we merge?

PR #1416, opened 2019-06-22
Add script to help sync upstream gtk folder

clobrano:


Feichtmeier:
Holy nice script!
It doesn't override our assets, right? Sorry for the noob question :)

clobrano:
The script copies all the data from `gitlab.com.../gtk/theme/Adwaita/` in whatever folder you decide to.
It is intended to update our `gtk/upstream/gtk+3.0/Adwaita` folder but only if you run

    adwaita-sync.sh --destination /path/to/yaru/gtk/upstream/gtk+3.0/Adwaita --assets

`--assets` is also optional, one can update only the scss code

Feichtmeier:
Sexy!!!

PR #1426, opened 2019-07-28
DON'T MERGE YET: Yaru icon for Transmission

ubuntujaggers:
Requires approval from @mikedld to have in Yaru.  Ready to go if we're going to use it.

image

image

image

image


Feichtmeier:
Awesome icon @ubuntujaggers!

@mikedld if you like we could alternatively put it in your transmission repo as the upstream icon so we wouldn't have a reason to put it into yaru

mikedld:
You've got a :+1: from me and @livings124.

PR #1429, opened 2019-07-28
Add missing Remmina symlink

ubuntujaggers:
I know we likely won't need this for long, but, for completeness, I added the missing symlink for Remmina.  This means Disco will use the Yaru icon for Remmina in the meantime.

Test:

image


clobrano:
It works very well, thanks!

PR #1435, opened 2019-08-04
Complete set of LibreOffice app icons

ubuntujaggers:
Have a go of these:

image

image

image

image


ubuntujaggers:
Okay, I've fixed the slight fuzziness around the 32px and 48px icons.  Latest pushed versions:

**32px**

image

**48px**

image

image

If you're struggling to see the difference, in the original screenshots, the fuzziness only affected the main body of the icon, and not the cut corner.  So if you compare the edges between the main part and the corner, you'll see the fuzziness, and then you can see that it's gone in the newest screenshots.

@Feichtmeier if you have time to test again, we can merge :+1:

Feichtmeier:
@ubuntujaggers ah I can see a difference! I think this looks better!

image

PR #1441, opened 2019-08-08
Bionic: Symbolic icons for Remmina and Transmission

ubuntujaggers:
Again, multi commit cherry pick, so hope it works :+1: 

This is the same as #1431 for Bionic.


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1441`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1441` channel to test your snaps with those changes.

PR #1453, opened 2019-08-13
Icon for Jockey

ubuntujaggers:
Haven't tested this one (other than swapping it onto a random app to see how it looks) because I think you only get it with a new install.  It's just called "jockey".

image

image

image

image


clobrano:
Awesome!

Feichtmeier:
Very nice!

You should have it in every install. The "app" is called "additional drivers"

ubuntujaggers:
Does it uninstall itself after you run out then? I'll try installing it :)

Feichtmeier:
It should be part of software-properties-gtk which contains "softwares and updates" and "additional drivers"
Maybe they removed it for disco? I think the screenshot I posted was from my new eon VM

I'll check tonight if you aren't faster

Feichtmeier:
Works !

grafik

PR #1454, opened 2019-08-13
Symbols: Updated file manager, camera, new symbol for Jockey

ubuntujaggers:
I'm doing this as one PR to avoid clashes.

**File Manager**

The symbol has been updated to use the same design as the new full colour:

image

**Camera**

I drew a camera symbol for the app, but recently found an existing one in Devices.  The existing one is a nicer design, so I retired mine:

image

**Jockey**

...and we've got a symbol for Jockey:

image


madsrh:
Nice work @ubuntujaggers 🎉 That file manager icon turned out pretty great 💪

fitojb:
This change also fixes https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1819769 <3

PR #1457, opened 2019-08-14
gnome-shell: Sync scss with upstream changes for 3.34

3v1n0:
This adds support for app-icons DnD, app folders and fixes the missing search
results.

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -619,6 +619,18 @@ StScrollBar {
   app menu inside the main app window itself rather than the top bar
 */
 
+/*************
+ * App Icons *
+ *************/
+/* Outline for low res icons */
+.lowres-icon {
+    icon-shadow: 0 1px 2px rgba(0,0,0,0.3);
+}
+
+/* Drapshadow for large icons */
+.icon-dropshadow {
+  icon-shadow: 0 1px 2px rgba(0,0,0,0.4);
+}
 
 /* OSD */
 .osd-window {
@@ -728,7 +740,8 @@ StScrollBar {
     spacing: 8px;
   }
 
-  .ws-switcher-active-up, .ws-switcher-active-down {
+  .ws-switcher-active-up, .ws-switcher-active-down,
+  .ws-switcher-active-left, .ws-switcher-active-right {
     height: 50px;
     background-color: $selected_bg_color;
     color: $selected_fg_color;
@@ -1158,13 +1171,7 @@ StScrollBar {
 
   // a little unstructured mess:
 
-  .system-switch-user-submenu-icon { // this needs removal for gnome-shell 3.33+
-    icon-size: 16px;
-    padding: 0 4px;
-  }
-
   #appMenu {
-    spinner-image: url("resource:///org/gnome/shell/theme/process-working.svg");
     spacing: 4px;
 
     .label-shadow { color: transparent; }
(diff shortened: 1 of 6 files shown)


gnome-shell/src/gnome-shell-sass/_common.scss
```diff
@@ -633,7 +645,7 @@ StScrollBar {
   .level {
     height: 0.6em;
     -barlevel-height: 0.3em;
-    -barlevel-background-color: transparentize($osd_fg_color, 0.9);
+    -barlevel-background-color: transparentize($osd_fg_color, if($variant=='light', 0.2, 0.9));
```
Feichtmeier: I think this is not needed for our theme, but I need to double check this

gnome-shell/src/gnome-shell-sass/_drawing.scss
```diff
@@ -148,7 +148,7 @@
     color: $tc;
     text-shadow: 0 1px black;
     icon-shadow: 0 1px black;
-    box-shadow: inset 0px 0px 0px 1px $selected_bg_color;
+    box-shadow: inset 0px 0px 0px 2px $selected_bg_color;
```
Feichtmeier: We decided to use a 1px focus border for button focus.
But I found an issue in the Gtk theme with 1px focus rings, so if @madsrh @clobrano and @ubuntujaggers  agree, we can make all focus borders again 2px like in the upstream theme

Feichtmeier:
@3v1n0 thanks for the PR

I need to test this to preserve our design decisions or eventually drop them after ack of @madsrh @clobrano and @ubuntujaggers 

How much time do I have to test the branch? :)

Feichtmeier:
Okay, the change to the OSD volume BG needs to be reverted : )

It's in line ... 

Screenshot from 2019-08-14 19-48-43

gnome-shell/src/gnome-shell-sass/_common.scss
clobrano: Probably, our `$osd_fg_color` is the same for light and dark variant, but it doesn't hurt and reduces the diff

Feichtmeier: It sadly hurts a little, I attached a screenshot in the comment below, the level bg becomes full white :/

gnome-shell/src/gnome-shell-sass/_drawing.scss
clobrano: It's ok for me

gnome-shell/src/gnome-shell-sass/_common.scss
clobrano: Yes, you're right. The `$variant` variable defaults to light, so instead of transparentize 0.9 we have 0.2 now, which is too bright

gnome-shell/src/gnome-shell-sass/_drawing.scss
Feichtmeier: Okay for mads and Stuart also so we keep this change and adapt the gtk theme to it

gnome-shell/src/gnome-shell-sass/_common.scss
3v1n0: Ok, I've reverted the change. I was quite unsure about it in fact :stuck_out_tongue:

3v1n0:
@Feichtmeier
Definitely... The main reason for this wasfgetting the app icons views working again, so I leave all the theming details to you guys

PR #1459, opened 2019-08-16
Switches and switches in the headerbar

Feichtmeier:
- use a darker switch border for the dark theme (exactly like upstream)
grafik
- use a local var for headerbar buttons
- adjust the headerbar switches in the light theme to the dark headerbar
grafik

- Fix selected row slider border this needs to be done upstream for the next cycle, fixing this for EON downstream


clobrano:
Not sure I can spot the difference in light variant switches, but it looks good overall :+1:

real-amano:
Hmm. What about using the dark switch for the headerbar even in the light theme?

PR #1473, opened 2019-08-26
Fix slider fine-tune

clobrano:
Fine-tune feature currently is not working properly. Instead of making
the slider thicker, it actually changes only the space around causing the
window to shrink.

eoan-fine-tune-bug
(peek doesn't seem to show the cursor, but it is over the slider handle)

The fix consists in changing the margin-(top/bottom/left/right) according
to the position of the marks (above/below/etc.).

eoan-fine-tune-fix

Note: tested only on slider with marks below for the lack of apps with
the other kind of sliders.

### DIFF
--- a/gtk/src/light/gtk-3.20/_common.scss
+++ b/gtk/src/light/gtk-3.20/_common.scss
@@ -3360,15 +3360,15 @@ scale {
             margin: -7px;
 
             @if $dir_class == 'horizontal' {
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Works for the night light slider!

Peek 2019-08-29 10-30

clobrano:
Okay, cool! I'll merge it then

PR #1476, opened 2019-08-28
Slight improvement to top highlight on LibreOffice app icons

ubuntujaggers:
IMO, there's a slight optical illusion where the top highlight on the cut corner is too dull, even though it uses the same opacity as the rest of the top highlight.  So, I've brightened it by 10%.

Now it looks like this:

image


Feichtmeier:
Three thumbs, looks indeed a tiny bit better - merging :)

PR #1478, opened 2019-08-29
ubuntu-colors: Lighten headerbar up

### LINKED ISSUE #1477: The headerbar buttons lack of contrast compared to the background
(Sorry if it's something discussed before but I couldn't find a reference to it.)

The action buttons in headerbars are not really easy to see (it's partially fault of the upstream design but Yaru has even less contrast between the bg and buttons than adwaita). The issue was discussed on the desktop IRC channel in the context of current eoan and gnome-control-center/display and some team members couldn't find the display/night mode selector at the top of the page, it's not specific to that panel though (the 'select image' in the background panel for example has the same issue)

Is there anything that we could do to make those more easy to spot for users?

clobrano:
Current headerbar color is too dark to allow a good contrast with action
buttons. To fix it, I generated a palette from the original color
(#2B2929) and taken a lighter tone #3E3C3C and some more hightlight on button borders.

Mockups collected here https://paper.dropbox.com/doc/Headerbar-button-review--Aj3giatD9eqflHWxVIclGzkCAg-LNYRgSOLjZpVsWfWbTOXP

closes #1477

### DIFF
--- a/gtk/src/light/gtk-3.20/_tweaks.scss
+++ b/gtk/src/light/gtk-3.20/_tweaks.scss
@@ -100,10 +100,8 @@ headerbar,
     $_btn_backdrop_border: darken($headerbar_border_color, 3%);
 
     @include button(normal, $_headerbar_button_bg, $headerbar_fg_color);
-    //border-color: lighten($_headerbar_button_bg, 6%);
-    //box-shadow: 0px 1px darken($_headerbar_button_bg, 10%);
-    //box-shadow: inset 0 0 0 1px lighten($_headerbar_button_bg, 6%);
-    box-shadow: inset 0 1px 0 0px lighten($_headerbar_button_bg, 8%);
+    box-shadow: inset 0 1px 0 0px lighten($_headerbar_button_bg, 8%),
+                inset 0 -1px 0 0px lighten($_headerbar_button_bg, 3%);
 
     &.flat {
       @include button(undecorated);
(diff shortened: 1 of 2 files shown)


madsrh:
At first sight I don't mind the color change, but this really clashes with the top-panel and dock IMHO.

Feichtmeier:
@madsrh any idea on a different set of colours? Thinking about giving you/carlo the upstream files and you/carlo play around with the BG/fg/base/selected colours

clobrano:
> At first sight I don't mind the color change, but this really clashes with the top-panel and dock IMHO.

Sure, we might need to keep the same color ratio with the panel, making it brighter as well

Feichtmeier:
Peek 2019-09-02 14-29

Active and checked behave very different, resulting in a jump between black and gray, also the sizes jumps

clobrano:
Thanks, I'll fix it

Feichtmeier:
Alternatively, use this branch and apply your headerbarcolor and your box-shadow changes:
https://github.com/ubuntu/yaru/tree/Lightened_Hb

Feichtmeier:
@clobrano

If you don't change the Hb BG colour depending on the theme variant, you should put the whole headerbar part into an "if light"
Because like this the headerbar is now brighter than the window in the dark theme ...

clobrano:
Don't you think it's too dark, even for the dark variant?

Feichtmeier:
Im not sure but then the whole dark theme should be brightened up. Imho a lighter Hb in the dark theme looks strange at least to me.
But I'm not sure if the drawings are that flexible because they assume that the dark theme is quiet dark

clobrano:
No, I don't think lighten up the whole theme is feasible

Feichtmeier:
Yes also then the shell needs to be brightened up even more to be in sync with gtk and it's already very bright

clobrano:
Okay

Issue #1475, opened 2019-08-27
Shell theme asset paths


Feichtmeier:
Assets are not loaded when using the ```url("resource:///org/gnome/shell/.....``` path

It works if you remove the absolute path to the resource but I think then the svg is not converted into a gresource.

This was the reason for the checkmarks not showing up before #1472 

Another example is the on screen keyboard:
With absolute path there are the adwaita assets
![grafik](https://user-images.githubusercontent.com/15329494/63784145-7d366800-c8ee-11e9-9ffd-cb4c926492e6.png)

With relative paths there is our asset (I think the green is questionable, maybe @madsrh has an idea which of our colors fits here :D ?)
![grafik](https://user-images.githubusercontent.com/15329494/63784264-b242ba80-c8ee-11e9-8797-c9900b0a0049.png)

@3v1n0 @clobrano is there a way we could use a gresource in yaru for these assets?

madsrh:
> I think the green is questionable, maybe @madsrh has an idea which of our colors fits here :D 

Questionable indeed. The green jumps out but even more the orange. Would it be possible to go with the [upstream colors](https://raw.githubusercontent.com/gnome-design-team/gnome-mockups/master/osk/osk-wires.png) but using a shade from our color palette?

I don't see any of the other options working out here 🤔

Feichtmeier:
Or use orange for the arrows and green for the enter key?

clobrano:
+1 for using upstream colors in our palette. Orange always look a bit destructive :)

madsrh:
> Or use orange for the arrows and green for the enter key?

I do really like [Arrons initial stying](https://github.com/ubuntu/yaru/pull/961), but there was an issue with green not being a great solution for ENTER because it can also be a destructive action 🤷‍♂

> +1 for using upstream colors in our palette. Orange always look a bit destructive :)

👍

PR #1484, opened 2019-09-04
FIx asset paths and blue shift keys

Feichtmeier:
- until now some assets where not taken from the yaru paths but from the upstream shell theme
- using links to the build assets fixes this
- needs eventually a correct absolute path to the yaru resource in the future
- using blue for the shift keys
- using green for enter key
- adds white enter key asset
grafik

Closes https://github.com/ubuntu/yaru/issues/1475

@3v1n0 @didrocks please review this paths - are the assets still converted into gresources like this? And if not, is this a problem?
Why did I change the paths? Because previously our asset **were not loaded at all** but some fallback to the upstream gnome theme happened
Thus this PR was needed to load out checkboxes:
https://github.com/ubuntu/yaru/pull/1472

@ubuntujaggers @madsrh @clobrano please see if you like this colours

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1067,7 +1067,7 @@ StScrollBar {
       .calendar-day-with-events {
         color: lighten($fg_color,10%);
         font-weight: bold;
-        background-image: url("resource:///org/gnome/shell/theme/calendar-today.svg");
+        background-image: url("calendar-today.svg");
       }
       .calendar-other-month-day {
         color: transparentize($fg_color ,0.5);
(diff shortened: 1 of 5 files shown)


madsrh:
I really like this! It looks good 👍

ubuntujaggers:
Agreed!

madsrh:
I don't know the upstream decision for coloring the SHIFT keys, but if it's "allowed", a fat white icon ⬆️  might look better than mixing blue and green 🤷‍♂ 

Here is (again) Aarons keyboard:
https://github.com/ubuntu/yaru/pull/961#issuecomment-437418944

Feichtmeier:
I'm not sure if those are really the shift key LOCK assets but I'll look into the bionic branch!

Feichtmeier:
Nope, all three are fatter, but both shift key lock assets are also blue:

grafik

Shall I take those assets from the bionic branch or keep the current upstream recolors? (diff doesnt really matter here, so it's purely a look/design decision)
@madsrh

madsrh:
I think I prefer the fat one but it's not a deal breaker for me.

Feichtmeier:
Peek 2019-09-06 10-33

grafik

grafik

grafik

@madsrh

Issue #1490, opened 2019-09-09
Scroll bar in the focussed Terminal window is almost invisible


vanvugt:
Scroll bar in the focussed Terminal window is almost invisible. Strangely it's quite visible for all the other (unfocussed) Terminal windows.

## Expected Behavior

The scroll bar colour contrasts with the Terminal window background.

## Actual Behavior

The scroll bar colour does not contrast with the Terminal window background.

![Screenshot from 2019-09-09 11-15-08](https://user-images.githubusercontent.com/11951986/64501316-6848d400-d2f3-11e9-9230-1c72e443683a.png)

## Steps to Reproduce the Problem

  1. Open a Terminal window.

## Specifications

Ubuntu 19.10
```
ii  yaru-theme-gnome-shell                     19.10.1                               all          Yaru GNOME Shell desktop theme from the Ubuntu Community
ii  yaru-theme-gtk                             19.10.1                               all          Yaru GTK theme from the Ubuntu Community
ii  yaru-theme-icon                            19.10.1                               all          Yaru icon theme from the Ubuntu Community
ii  yaru-theme-sound                           19.10.1                               all          Yaru sound theme from the Ubuntu Community
```

Downstream bug: https://bugs.launchpad.net/bugs/1842898

Feichtmeier:
@vanvugt that's strange it's showing up for me. Do you used X or Wayland in that picture? 
We'll fix that

vanvugt:
The bug occurs in both X and Wayland sessions, on every machine I install Ubuntu 19.10 on.

vanvugt:
Though for some reason (this white web page?) the screenshot looks worse here. Try opening the image in a new tab.

Feichtmeier:
Okay then I will just reduce the transparentization

vanvugt:
I guess the best test would be a black background?

Feichtmeier:
https://github.com/ubuntu/yaru/blob/master/gtk/src/light/gtk-3.20/_apps.scss#L179

 it looks definitely more opaque for me here, but in the picture it's not visible enough, so we could just decrease the transparentization

vanvugt:
How good is your monitor's colour reproduction? That might be a factor here. I think I've only tested this on fairly accurate monitors so far, that are close to 100% sRGB.

Feichtmeier:
No idea tbh, I only know that one of my pc's has a very bad brightness!

There was another problem I noticed while working on a fix.
[...]

PR #1491, opened 2019-09-09
Yaru app gtk theme code changes

Feichtmeier:
- remove forced scrollbar colors to make them more visible in the dafult colors but also to make them visible in all color combinations if the user doesnt take the system colors

Gnome software app css:
- remove the commented code for stars which we decided to not ship since it does not really fix the problems of hardly seeable stars in the dark theme and rather makes it look ugly
- also remove code for the infobar in the shell extension tab because it is no longer needed after the upstream refresh

Nautilus:
 Remove nautilus css that stopped working

- in nautilus 3.28 we had made selected folders in the main view to only color the label in orange and not the whole icon+image. This stopped working with nautilus 3.30, thus we can remove the code.
upstream bug: https://gitlab.gnome.org/GNOME/nautilus/issues/786

Closes https://github.com/ubuntu/yaru/issues/1490


clobrano:
Looks good to me :+1:

PR #1492, opened 2019-09-09
Lighter headerbar v8

clobrano:
Alternative to #1478, that preserves better the original Yaru dark style. Base headerbar color is #323030 instead of #424040 from #1478, plus a series of hints to keep buttons more visible

image
image


Feichtmeier:
Can we safe this code in a _headerbar.scss or would you just rely on the branch and not delete it?

Issue #1497, opened 2019-09-09
Tag labels in Nautilus are too high


madsrh:
Tag labels in Nautilus are too high
![image](https://user-images.githubusercontent.com/3986894/64571480-43407980-d364-11e9-86b1-cdc9bf0dc72a.png)

Feichtmeier:
Nautilus checks the current gtk theme
And only If the current gtk theme is Adwaita this custom CSS file and within the line which we now need to readd for us:

https://gitlab.gnome.org/GNOME/nautilus/blob/master/src/resources/css/Adwaita.css#L77

PR #1499, opened 2019-09-10
Re-add adwaita custom code for search entries

Feichtmeier:
- copy search entry custom code for nautilus that is only loaded if Adwaita is the current gtk theme https://gitlab.gnome.org/GNOME/nautilus/blob/master/src/resources/css/Adwaita.css#L77

grafik

@madsrh and/or @clobrano okay like this? Or further improvements?

Closes https://github.com/ubuntu/yaru/issues/1497


madsrh:
LGTM

clobrano:
LGTM to me as well, merging it, thanks!

PR #1507, opened 2019-09-10
Change Audio volume change

Feichtmeier:
- previously the sounds created distortion when used via the mouse-over-mouse-wheel change on the volume indicator
- this shortens the sound and overlays it with the freedesktop start and end noise, which - strangely - nullifies the distortion on mouse-wheel- change

Closes https://github.com/ubuntu/yaru/issues/1111


madsrh:
I like sound 👍 There's some noise in the beginning of the file, but it's totally unnoticiable because the sound is so short.

image

Feichtmeier:
Yes the noise is needed for the sound to get not clipped... Totally strange I know :)

PR #1519, opened 2019-09-12
Gtk and shell theme post 19.10 ui freeze fixes

Feichtmeier:
Two actual fixes (actually quiet important):

- use our titlebuttons also in the selection-mode, accidentally they were unstyled
- use a transparent active effect for clicking on a not-selected app in the dock

And some organization (not so important):

- Organize _tweaks.scss
-> remove import of ubuntu-colors because it's already imported into _colors.scss
-> remove small_radius which is no longer needed, new corresponding upstream variable is $button_radius in _common.scss
-> use a $_titlebutton_height for both the min sizes of the titlebuttons and the draw_circle mixin to avoid mistakes

- Use ubuntu-colors in colors
-> actually use the colors we imported instead of using the same hexvalue

Merge either after 19.10 or when we got UI freeze exceptions for this OR if not needed because it is not a design change but bug fixes merge when @iainlane thinks it is okay :smile_cat:

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -212,7 +212,7 @@
 
     &:active .overview-icon,
     &:checked .overview-icon {
-      // when clicking on an app
+      background-color: $base_active_color; box-shadow: none;
     }
 
     &.focused .overview-icon {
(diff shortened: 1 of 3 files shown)


Feichtmeier:

gunnarhj:
I can confirm that this change is not a problem from a docs POV. There are no screenshots which would be affected.

PR #1520, opened 2019-09-13
Dash to dock opaqueness revert

Feichtmeier:
- revert back to the fully opaque dock to apply the transparency change in dashtodock, outside of yaru


Feichtmeier:

Issue #29, opened 2018-01-10
Terminal color scheme


galgalesh:
I'd like to discuss the terminal color scheme a bit. I did some research and surprisingly, there is only one "engineered" terminal color scheme: [Solarized](http://ethanschoonover.com/solarized). When you start using Solarized, you immediately feel that it is an extremely readable color scheme. It's very friendly for colorblind people because of the combination of both color wheel and lightness relationships. As a result, the difference between red and orange for example is still clear to me.

The Unity8 color scheme looks similar to Solarized, but it is a bit less clear, even though it is a lot better than the default Gnome color scheme. The Ambiance color scheme sits in between Solarized and Unity8 in terms of clarity.

Solarized dark is shipped with Gnome Terminal, but it is disabled by default. You can try it out in `edit>profile preferences>colour`.

**Unity8**

![screenshot from 2018-01-10 11-08-30](https://user-images.githubusercontent.com/1492981/34766992-e743cd18-f5f6-11e7-8710-65cc7d453688.png)

**Solarized Dark**

![screenshot from 2018-01-10 11-08-59](https://user-images.githubusercontent.com/1492981/34767007-efd9fbaa-f5f6-11e7-992d-67f919b7312d.png)

**Ambiance**

![image](https://user-images.githubusercontent.com/1492981/34767222-d9161222-f5f7-11e7-85a8-46a4d185aecf.png)

**next to each other**

(best to zoom in :wink: )

| Unity8        | Solarized Dark           | Ambiance  |
| ------------- |:-------------:| -----:|
| ![screenshot from 2018-01-10 11-08-30](https://user-images.githubusercontent.com/1492981/34766992-e743cd18-f5f6-11e7-8710-65cc7d453688.png)     | ![screenshot from 2018-01-10 11-08-59](https://user-images.githubusercontent.com/1492981/34767007-efd9fbaa-f5f6-11e7-992d-67f919b7312d.png) | ![image](https://user-images.githubusercontent.com/1492981/34767222-d9161222-f5f7-11e7-85a8-46a4d185aecf.png) |

I've been comparing them for the past few days and my preference goes out to using Solarized Dark, because it is simply the clearest color scheme. However, the disadvantage of that approach is that the color scheme isn't "branded".

What do you guys think?

clobrano:
I'm in favor of Unity8 because of the lighter color of normal text (it's the same change I set every time I switch to Solarized)

Unity8:
[...]

PR #1521, opened 2019-09-13
Revert "Terminal additions"

clobrano:
image

### DIFF
--- a/gtk/src/light/gtk-3.20/_apps.scss
+++ b/gtk/src/light/gtk-3.20/_apps.scss
@@ -134,7 +134,7 @@ terminal-window {
 
         &:backdrop {
             background-color: lighten($terminal_bg_color, 2%);
-            color: $backdrop_fg_color;
+            color: $backdrop_selected_fg_color;
         }
     }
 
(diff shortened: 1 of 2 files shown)


madsrh:
> ...change the default variant of the terminal to dark...

IMO the is the best solution. The current white one is horrible and I really don't spent much time in the Terminal.

Alternatively we could change the color scheme [like suggested earlier](https://github.com/ubuntu/yaru/issues/29). I think the Unity 8 colors work better (_not perfect, just better_) with the light headerbar. Something like:

image

didrocks:
As stated on IRC, I still think that the aubergine terminal is a strong branding asset for ubuntu and should be kept, I will let other commenting

8none1:
Yes, I'm a big +1 on aubergine.  Blue is moving too far away from our brand IMO.

clobrano:
Do you mean aubergine in both light and dark variant? Or, as suggested previously, only in dark variant (as it is now), but setting Terminal to prefer dark variant (so that it will show aubergine by default)?

8none1:
I think that aubergine in both is the right call.

Feichtmeier:

Issue #1526, opened 2019-09-13
Add support for the folder-rename popup


3v1n0:
It has been introduced recently by GNOME shell upstream ([MR](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/675/diffs#diff-content-d27519967b4ac5df7ae7d6279de3e5ec655a72bf), will be part of 3.34.2)

Adding this should be enough:
```scss
// Rename popup

.rename-folder-popup-box {
  spacing: 6px;
  margin-left: 12px;
  margin-right: 12px;
}

```

PR #1527, opened 2019-09-13
Style rename popups

Feichtmeier:
- taken from https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/675
- closes https://github.com/ubuntu/yaru/issues/1526

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -615,6 +615,13 @@ StScrollBar {
     border-bottom-style: solid;
   }
 
+// Rename popup
+
+.rename-folder-popup-box {
+  spacing: 6px;
+  margin-left: 12px;
+  margin-right: 12px;
+}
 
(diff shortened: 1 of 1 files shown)


3v1n0:
Also update the upstream file :)

But anyways.... Probably better to keep this in hold for a bit, since it will be changed soon again very likely. See this [MR's diff](https://gitlab.gnome.org/GNOME/gnome-shell/merge_requests/720/diffs#diff-content-d27519967b4ac5df7ae7d6279de3e5ec655a72bf).

Feichtmeier:
@3n1n0 is this pr still needed?

3v1n0:
Ouch yes, but it should be adapted to https://gitlab.gnome.org/GNOME/gnome-shell/blob/3.34.1/data/theme/gnome-shell-sass/_common.scss#L613

I think we should SRU this indeed too

clobrano:
I'm deeply sorry, but it was easier to open a new PR (#1572) than rebase this one

PR #1532, opened 2019-09-16
Release Yaru 19.04 as third variant "Yaru-Ambiance"

### LINKED ISSUE #1494: Provide a mixed theme
Some thoughts:

Provide a mixed theme, based on the light theme but with inverted headerbar and headerbar button colors.

This needs probably a change to our meson build, with a new file "_headerbar.scss" which only gets imported when the mixed theme is build.

The gtk2 theme could be copied from the light theme

clobrano:
After the change of the light variant, in order to ease the transition out, release the Yaru 19.04 (sometimes called "mixed" theme) as third variant.

- [x] update build scripts to install a third them Yaru-ambiance
- [x] replace light variant tweaks with headerbar sass file
- [ ] add a dark variant (which is the same as Yaru-dark) for apps that prefers dark variant
- [x] fix different titlebutton padding between ambiance and -dark

closes #1494


Feichtmeier:
Some stuff I've noticed:

- Headerbar buttons look unproportionality strong pressed
grafik

- Titlebuttons padding changes between dark and light variant, because for the dark theme you get the titlebutton styling from _tweaks.scss and for the ambiance version from headerbar.scss were I made the mistake to remove the padding from all titlebuttons and not only min/max/close
--> I would recommend to remove all the code from _headerbar.scss that is not ambiance only code

Feichtmeier:
Could you please drop those additional box-shadows and just set 1 color for the buttons and then use the button mixins?
It looks horribly inconsistent with the rest of the buttons

clobrano:
> Could you please drop those additional box-shadows and just set 1 color for the buttons and then use the button mixins?
> It looks horribly inconsistent with the rest of the buttons

Ok

Feichtmeier:
> > Could you please drop those additional box-shadows and just set 1 color for the buttons and then use the button mixins?
> > It looks horribly inconsistent with the rest of the buttons
> 
> Ok

There is ```$_headerbar_button_bg``` at the top of the file - just use this for all the ```@include button(normal ... ) @include button(active ...) @include button(.....)``` parts, because the mixins expect one color as input no matter of the state

clobrano:
image
image

clobrano:
> entries in linked boxes are bigger than buttons

this is Image viewer, right? The app is not taking the right varian here. Applying the dark one manually it works fine

Feichtmeier:
I think the reason for the dark variant not working for videos and eye of gnome and so on... media apps... is that it searches for "gtk-dark.css" but there is none. Instead there is "gtk-ambiance-dark.css"

clobrano:
this branch started with gtk-dark only and didn't work, this is why I added later gtk-ambiance-dark, which is still not working of course.

Feichtmeier:
@clobrano maybe it's something with that gresource file names

clobrano:
Maybe. I expect gtk to look for the same name with -dark postfix. I think I replicated already all the gresource entries, but there must be something missing. @willcooke should have already talked with @didrock about that, I'll ping them tomorrow

PR #1538, opened 2019-09-17
Fixing some battery icons which have an incorrect xml class on the fill ( = orange)

### LINKED ISSUE #1531: Battery icon looks red, even when the battery is full
In Ubuntu 19.10 the battery icon in the top bar looks red (actually it's orange rgb(245,121,62)) even when the battery is full, e.g. 99%.

The orange colour makes it look like a warning that the battery is low.

ubuntujaggers:
Closes #1531.

Basically: because of how I made the icons, some of the new ones incorrectly picked up an xml class called "warning" on the inside fill.  Gnome is responding to this class by recolouring the element orange.

I've removed the xml class and pushed new versions of the affected symbols (plus an updated source-symbolic.svg).

**However** I can't test this because my install isn't using the new battery icons anyway.

@chrisjbillington @ElectricGestalt @8none1 @clobrano @Amr-Ibra : if anyone could confirm that I really have fixed it, that would be great :+1:


chrisjbillington:
Seems to have fixed the problem for me!

clobrano:
It seems good for me as well, I'll just wait for the battery to go down a little to test it fully

:+1:

8none1:
I built and installed Yaru (including the icon themes package) from this branch and I can confirm that the battery icon is no longer red/orange.
LGTM.

clobrano:
Thanks @8none1, merging it now then

PR #1543, opened 2019-09-20
Light dark default rebased

clobrano:
Same as #1542, but rebased onto master, so that Github can automatically merge

I am a bit paranoid maybe, but I prefer to create this new PR to test and do not touch #1542.

Please test that

- default Yaru is Ambiance variant
- prefer dark apps (e.g. Image viewer) always use dark variant

Thanks


madsrh:
@clobrano this looks good, but I just came across this one from Disks in backdrop:

image

clobrano:
Thanks @madsrh,

It seems to me that Disks behavior is the same in all variants and in my installation of master, for all applications.

"Format disk" is a titlebar, hence SSD, and requires a shell restart to work after building Yaru

Terminal tabs yes, it's indeed different from 19.04, but it is coherent with other tabs at least (e.g. Nautilus).

What do you think?

Feichtmeier:
> @clobrano this looks good, but I just came across this one from Disks in backdrop:
> 
> image
> 
> and...
> image
> 
> In Terminal the colors are back to normal, but tabs are still light 
> image

Maybe the separator needs also the backdrop to be styled
https://github.com/ubuntu/yaru/blob/a2737933fb627b4ff3834ad63c01368dd8d031b6/gtk/src/light/gtk-3.20/_headerbar.scss#L197

PR #1556, opened 2019-10-02
Adding new full colour app icon for Evince

### LINKED ISSUE #1489: Icon of Evince document viewer is confusing
Evince has a rather strange icon in my opinion. The use of a magnifier glass in the icon indicates that it is a search/find utility rather than a document viewer. I hope it could be improved.
grafik

Take a look at the upstream icon for inspiration:
https://gitlab.gnome.org/GNOME/evince/blob/master/data/icons/scalable/apps/org.gnome.Evince.svg

grafik

ubuntujaggers:
image

image

image

image

Closes #1489


Amr-Ibra:
When you install gnome-documents, it uses the documents-app.svg icon, which looks very similar to this icon.

clobrano:
@ubuntujaggers, could you show us the two icons together?

ubuntujaggers:
Ah, I see what you mean.  They're certainly nearly the same (the Evince icon just has the branding) but maybe we should change the icon for gnome-documents, since the Evince icon upstream does have glasses.  Installing gnome-documents now...

ubuntujaggers:
image

How about Gnome documents being the icon with the magnifying glass?

clobrano:
I'm not sure, could the magnifying glass be a little misleading? There are
actually other apps that magnify the screen

Amr-Ibra:
The original Suru design is that a magnifying glass is for a document viewer, and glasses is for a document explorer.
You can see that same design for the image viewer. It's icon also has a magnifying glass, since it's a "viewer".

Feichtmeier:
> image
> 
> How about Gnome documents being the icon with the magnifying glass?

Tbh this is totally fine and very distinguishable.
Also, gnome documents is not installed in ubuntu by default, and thus shouldn't be very high prioritized if this solution brings an improvement for the PDF viewer

Amr-Ibra:
I suggest to just keep the original Suru icons as they are, and modify the docviewer-app.svg enough to make it more "Evince" and less a generic document viewer.

Feichtmeier:
If you think of the suru baseplate: That won't happen because we already changed all icons to use one out of 4 baseplates (reasons were discussed ages ago).
Or what do you mean?

Amr-Ibra:
I mean the current two icons (docviewer-app.svg and documents-app.svg), as they are, are generic document viewer and generic document explorer. Evince uses the viewer icon and gnome-documents uses the explorer icon.
The idea for this pull request is that a user, who has both Evince and gnome-documents installed (myself included), will be getting two very similar icons for two different apps (Evince is a viewer and gnome-documents is an explorer) and both are using the "explorer" icon. This breaks the current situation.
I suggest that they would be left as they are (this pull request is already doing that) and take the docviewer-app.svg icon as a base for a new icon for Evince.

ubuntujaggers:
The reason I'm in favour of using the existing glasses icon as the base for Evince is because it harmonises so closely with upstream:

evince-new-icon

66036113-5fb77800-e504-11e9-88c2-dbd3bc5df0bd

Maybe we could even make the glasses blue and horizontal to further differentiate this from documents-app?

Amr-Ibra:
I like the idea of the glasses 👍

Feichtmeier:
Imho it looks perfect @clobrano or @madsrh any veto for merging?

madsrh:
This looks good @ubuntujaggers

ubuntujaggers:
Before you merge, give me a minute to see if rearranging the elements is better...

clobrano:
Looks great to me

ubuntujaggers:
This is the version rearranged to look more like upstream:

image

image

image

image

So either merge this PR as is, or let me know if you want me to push the new version :+1:

Amr-Ibra:
> This is the version rearranged to look more like upstream:
> 
> image


I like this arrangement too, and I have two comments:
- The blue glasses are not aligned in the centre of the icon (there is a bit more space on the right than left)
- The lines in the icon look a bit untidy, the original lines look tidier

Otherwise, the icon looks very nice.

clobrano:
I like the "misplacement", if its intended :)

ubuntujaggers:
Idea behind having the glasses on the left was to follow upstream a bit, but it's negotiable :) I can try making more of the lines full width.

ubuntujaggers:
Okay, tidied the lines up a bit - I vote for this one:

image

image

image

Happy to push :+1:

Amr-Ibra:
> Okay, tidied the lines up a bit - I vote for this one:
> 
> image

I vote for this one too.

Feichtmeier:
Okay, being different from gnome-documents is a good reason for the re-arrangement I guess. :)

ubuntujaggers:
Pushed :+1:

clobrano:
LGTM :+1: 
thanks @ubuntujaggers

PR #1562, opened 2019-10-06
Symbolic icon for new Evince design

ubuntujaggers:
image

image


clobrano:
It works fine! Thanks

PR #1568, opened 2019-10-09
When building with communitheme_compat, link the light variant

kenvandine:
gtk-common-themes builds with communitheme_compat=true. This was missing the light variant, so theming is broken on snaps when the user has communitheme-light set as their gtk theme.


clobrano:
Thanks again :)

PR #1578, opened 2019-10-19
shell: enable shell variant switch

### LINKED ISSUE #1483: Allow to switch between light and dark shell themes
According to the discussion in Discourse from https://discourse.ubuntu.com/t/yaru-next-move/12155/45?u=memeplex onwards (up to comment 55)  and also to a previous, related, Launchpad discussion in https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1832137, I would like to propose a simple way to fix two issues, one of them maybe just temporarily (in case upstream eventually decides to add a global dark mode toggle) but in any case easily reversible:

1. It's currently not possible to switch  between a light and a dark shell theme, even if both are easily generated from sources (switching `$variant` from "light" to "dark").

2. It's currently not possible to use Adwaita in an Ubuntu session nor is it possible to use Yaru in a vanilla Gnome session.

Both problems are related to the status of Yaru and Adwaita as session defaults, hence residing at `/usr/share/gnome-shell/theme/(Yaru|Adwaita)/gnome-shell` instead of `/usr/share/themes/(Yaru|Adwaita)/gnome-shell`. So if you enable the stock User Themes extension you end up seeing just one of them (according to the type of session you're in) as a `Default` entry in the shell theme selection dropdown.

Now, it's common practice for almost every other theme out there (most of which don't have the honor of being a distro or desktop default) to provide many versions (light, dark, darker, compact, etc) of both gtk and shell themes. For example, for Materia you get a plethora of themes (12!):

`/usr/share/themes/Materia(|-dark|-compact|-dark-compact|-light|-light-compact)/gtk-3.0/gtk.css`

and

`/usr/share/themes/Materia(|-dark|-compact|-dark-compact|-light|-light-compact)/gnome-shell/gnome-shell.css`

Even if it's nice to get Yaru by default, the current situation is too limiting in both aspects mentioned above:

1. People using gnome-session *in Ubuntu* have access to any packaged theme except for, of all, *Ubuntu* theme!

2. Even if upstream enabled a dark theme switch, you might prefer to provide light, dark, darker themes (there were recent discussions about this in Discourse), or even compact ones (Materia compact is very nice, indeed).

---

I suggest to generate both light and dark shell themes css in `/usr/share/gnome-shell/theme/Yaru/gnome-shell` and to provide symlinks to both variants as `/usr/share/themes/Yaru(|-dark)/gnome-shell/gnome-shell.css`. This way you get in the shell theme dropdown three options:

1. Default
2. Yaru
3. Yaru-dark

The fact that Default is an alias to Yaru might be distressing for some people (not for me). But in a gnome-session, Default = Adwaita, so you get full choice. If you don't care about this, just provide a symlink to Yaru-dark, so you get:

1. Default
2. Yaru-dark

---

Alternatively, you can deploy the real thing into `/usr/share/gnome-shell/theme/Yaru(-dark)/gnome-shell` and provide just one symlink for the default in  /usr/share/gnome-shell/theme/Yaru/gnome-shell/gnome-shell.css`. I think this solution is cleaner, since it's a standard practice to symlink to defaults or to selected options. Still, it diverges a bit more from upstream than the previous alternative.
 
---

At first sight this task looks like a mix of packaging and theme generation. But since git support symlinks I think that carefully created relative symlinks might be directly added to the repo so that they are packaged and installed automagically.

clobrano:
- generate both light (current default) and dark shell variants
- install variants in `/usr/share/themes/Yaru/gnome-shell` and `/usr/share/themes/Yaru-dark/gnome/shell`
- symlink default variant to `/usr/share/themes/`

Once installed "User themes" extension, both shell variants are
manually selectable, also in GNOME vanilla session.

closes #1483


roypen:
Works great after changing the symlinks mentioned above. Tested on a 19.10 clean install.

gnome-shell/src/dark/meson.build
```diff
@@ -0,0 +1,83 @@
...
+  '../toggle-on-dark.svg',
+  '../toggle-on-hc.svg',
+  '../toggle-on-intl.svg',
+  '../toggle-on.svg',
+  '../ws-switch-arrow-down.svg',
+  '../ws-switch-arrow-up.svg',
+]
+
+# install static data files
+install_data(data_sources, install_dir: install_dir)
```
memeplex: There is obviously a lot of duplication in this file.

My first impulse would usually be to factorize everything that is not the install_dir parameter. For example, to symlink the base build file and provide install_dir as a parameter or as part of an environment (I'm not familiarized with meson).

But in this case I believe this file could be seen as a place for little tweaks and hacks that could be required in the future to adjust the theme variant (for example, different icons), so in the end I have no strong opinion.

gnome-shell/src/light/meson.build
```diff
@@ -0,0 +1,83 @@
+# destination directory
+install_dir = join_paths(gnomeshell_theme_dir, meson.project_name())
```
memeplex: As I commented in https://github.com/ubuntu/yaru/issues/1483#issuecomment-544180979, I believe it's better here to keep symmetry with what you're doing for the dark theme, that is:

```
install_dir = join_paths(gnomeshell_alt_theme_dir, meson.project_name() + '-light', 'gnome-shell')
``` 

and then to invert the symlinking in `install-shell.sh`. This better generalizes to future variants, also. And it follows the time-tested pattern of symlinking the chosen variant between a set of configuration options.

gnome-shell/src/install-shell.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# -*- coding: UTF-8 -*-
+
+project_name="$1"
+prefix="${MESON_INSTALL_DESTDIR_PREFIX}/share"
+
+ln -sf "${prefix}/gnome-shell/theme/${project_name}" "${prefix}/themes/${project_name}/gnome-shell"
```
memeplex: I suggest to revert the direction of this symlink as explained in my comment for light/meson.build below.

clobrano:
I'll fix the symlink, which are needed to avoid more code duplication while genereting both dark and light variants

gnome-shell/src/dark/meson.build
clobrano: it is easier if the install dir is shared

dawidd6:
Will this change go live on 19.10 when merged?
Because right now having mixed white shell theme with dark gtk one is not a good experience.

Anyway, thanks for doing it!

clobrano:
Usually there's at least one update of ubuntu packages even in non LTS, so it might

gnome-shell/src/dark/meson.build
clobrano: I think I'll keep them separated tough. The `custom-target` directive now generates two different targets, one for light and one for dark, with the same name `gnome-shell.scss`, while keeping it in the same meson build is causing target duplication which resolution makes this even more complicated

ayush987goyal:
Hi @clobrano . How can I use this in the current `19.10` installation?

DarthWound:
> Will this change go live on 19.10 when merged?
> Because right now having mixed white shell theme with dark gtk one is not a good experience.

> Usually there's at least one update of ubuntu packages even in non LTS, so it might

Would be great, because even with light GTK theme I don't like the light Shell theme, popup menus "blend" too much into maximized windows, it's less readable for me. I usually prefer a light GTK mixed with a dark Shell. Of course that's just me, but I know a lot of people use the dark GTK and I'm pretty sure having the light Shell is painful for their eyes :stuck_out_tongue:

Feichtmeier:
We could surely go with a darker tone of white at some point. But then for both gtk and shell popups/popovers.

BloodyIron:
How soon before this hits mainline Ubuntu repos already? White Gnome pull-down menus with a dark theme for the rest is pain to the retinas. :/

dawidd6:
@clobrano Any word about getting these changes to 19.10 release?

clobrano:
At this point, I am not sure it will land before 20.04

BloodyIron:
That means 2 more years for the next LTS to get more dark theme support :/ what would it take to get this into 20.04?

clobrano:
What do you mean? In 20.04 will land for sure

BloodyIron:
So it will be _in_ 20.04?

clobrano:
yes sorry,  with "not before 20.04" I meant not in 19.10 :D

PR #1585, opened 2019-10-22
Ensure shell-theme link can be created when using DESTDIR

### LINKED ISSUE #1583: Build problem after directory layout change
After the changes to allow switching between dark and light shell themes, the Arch PKGBUILD I'm using broke. I figured the PKGBUILD needed to be updated to reflect the changes (It deletes directories in order to split yaru into multiple packages, sound, gtk, shell, session, icons so is vulnerable to changes in directory layout).

I thought I'd make a non-split package so that yaru can just be installed all at once and the PKGBUILD is no longer vulnerable to changes like this. But I'm hitting an issue and wondering if it is an issue with Yaru itself.

I'm building like so:

```
cd /tmp
git clone https://github.com/ubuntu/yaru
meson yaru build
ninja -C build
mkdir pkg
DESTDIR=/tmp/pkg ninja -C build install
```

And getting the following error during `ninja -C build install`:
```
<snip>
Running custom install script '/tmp/yaru/gnome-shell/src/install-shell.sh Yaru'
ln: failed to create symbolic link '/tmp/pkg/usr/share/gnome-shell/theme/Yaru': No such file or directory
FAILED: meson-install 
/usr/bin/meson install --no-rebuild
ninja: build stopped: subcommand failed.
```

Indeed, there is no such directory `usr/share/gnome-shell/theme`. There is `usr/share/themes/gnome-shell` though.

Does the script `gnome-shell/src/install-shell.sh` need to be updated to reflect the recent directory changes? Or am I misunderstanding something?

chrisjbillington:
When creating a package (such as on Arch Linux) using:

`DESTDIR="${pkgdir}" ninja -C build install`

to install all files to `${pkgdir}` for later installation onto an
actual system, ensure the symlink to the shell theme can be created.

The link target should point to a location in `MESON_INSTALL_PREFIX`,
but the link itself should be in `MESON_INSTALL_DESTDIR_PREFIX`.
Furthermore, the link's parent directory needs to be created within
`MESON_INSTALL_DESTDIR_PREFIX` if it does not exist.

Fixes #1583

### DIFF
--- a/gnome-shell/src/install-shell.sh
+++ b/gnome-shell/src/install-shell.sh
@@ -2,6 +2,8 @@
 # -*- coding: UTF-8 -*-
 
 project_name="$1"
-prefix="${MESON_INSTALL_DESTDIR_PREFIX}/share"
+destdir_prefix="${MESON_INSTALL_DESTDIR_PREFIX}/share"
+install_prefix="${MESON_INSTALL_PREFIX}/share"
 
-ln -sf "${prefix}/themes/${project_name}/gnome-shell" "${prefix}/gnome-shell/theme/${project_name}"
+mkdir -p "${destdir_prefix}/gnome-shell/theme/"
+ln -sf "${install_prefix}/themes/${project_name}/gnome-shell" "${destdir_prefix}/gnome-shell/theme/${project_name}"


clobrano:
looks good, thanks!

chrisjbillington:
:)

PR #1592, opened 2019-10-28
overwrite softlinks so install-compat can handle multiple builds

galgalesh:
`install-compat.sh` currently fails the second time you build Yaru because the softlinks already exist.

This PR fixed that by forcing `ln` to overwrite the existing softlinks.


clobrano:
Hey @galgalesh, thanks! It looks good to me, merging it


May I ask you how you noticed the failure? The buildings script seemd to work fine

galgalesh:
I found it by building the `gtk-common-themes` snap multiple times on my laptop. I hit this issue the second time I built the snap. I was trying some stuff out to [reduce the startup time of desktop snaps](https://github.com/snapcore/snapcraft/pull/2769).

clobrano:
Cool! Thanks :)

PR #1595, opened 2019-10-29
Fix headerbar focused entries

### LINKED ISSUE #1557: Mixed theme headerbar entries miss the :focus orange border
image
It only keeps the inset box shadow
https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-3.20/_headerbar.scss#L31

Without headerbar modification:
image

Eventually needs a dedicated focus styling here, because it's overwritten for not focus
Could be wrong, just the first idea.

Feichtmeier:
Closes #1557 

Screenshot from 2019-10-29 15-21-08
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -30,6 +30,7 @@ headerbar,
     &, &:backdrop {
       border-color: darken($headerbar_bg_color, 8%);
     }
+    &:focus {  @include entry(focus); }
   }
 
   button,


clobrano:
I've found the same solution, but if you notice in Yaru-dark the orange border doesn't get over the entry border

image

not sure if possible to have though

Feichtmeier:
Yeah that's because the upstream entry drawings draw a second orange line with a box-shadow made of selected_bg_color which is darker for the dark theme. It's barely visible but it's there ^.^
No idea why they do this :man_shrugging: 
Here: https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-3.20/_drawing.scss#L27
And here: https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-3.20/_drawing.scss#L30

PR #1596, opened 2019-10-30
shell: fix shell installation directories

### LINKED ISSUE #1591: corrupted yaru-theme-gnome-shell package
## Expected Behavior
package install correctly

## Actual Behavior

when install, result is:

```
root@whitewater:~# LANG=C dpkg -i yaru-theme-gnome-shell_19.10.4+201910260248+04f5461~ubuntu18.04.1_all.deb 
(Reading database ... 773161 files and directories currently installed.)
Preparing to unpack yaru-theme-gnome-shell_19.10.4+201910260248+04f5461~ubuntu18.04.1_all.deb ...
Unpacking yaru-theme-gnome-shell (19.10.4+201910260248+04f5461~ubuntu18.04.1) over (19.10.4+201910210131+909c1e5~ubuntu18.04.1) ...
Setting up yaru-theme-gnome-shell (19.10.4+201910260248+04f5461~ubuntu18.04.1) ...
update-alternatives: error: alternative path /usr/share/gnome-shell/theme/Yaru/gnome-shell.css doesn't exist
dpkg: error processing package yaru-theme-gnome-shell (--install):
 installed yaru-theme-gnome-shell package post-installation script subprocess returned error exit status 2
Errors were encountered while processing:
 yaru-theme-gnome-shell
```

install the last package

## Software that presents the issue

- Name: ubuntu
- Version: 18.04

## Yaru version

19.10.4+201910260248+04f5461

## Icon Request

N/A

clobrano:
Currently yaru gnome-shell theme is installed in the themes directory
/usr/share/themes/<themenane>/gnome-shell, which is where User-Themes
extension looks for 3rd party themes and a symlink is created under
/usr/share/gnome-shell/theme, where Gnome-shell looks for the session's
default theme.

Other than not being formally correct (the default theme should go into
the gnome-shell session's folder), this creates a dependency of the
gnome-shell-theme debian package from the gtk-theme package, that
creates the <themes>/Yaru folder first.

To fix the form and void adding dependency between packages (ideally,
gnome-shell could be installed without the gtk part), Yaru will be
installed under gnome-shell directory and the symlink will be created in
the themes directory, together with the installation of and Yaru-dark
shell, which is the alternative.

closes #1591 

- [ ] test deb package installation

### DIFF
--- a/gnome-shell/meson.build
+++ b/gnome-shell/meson.build
@@ -1,3 +1,4 @@
+gnomeshell_default_theme_dir = join_paths(get_option('datadir'), 'gnome-shell', 'theme')
 gnomeshell_alt_themes_dir = join_paths(get_option('datadir'), 'themes')
 
 subdir('src')
--- a/gnome-shell/src/install-shell.sh
+++ b/gnome-shell/src/install-shell.sh
@@ -5,5 +5,5 @@ project_name="$1"
 destdir_prefix="${MESON_INSTALL_DESTDIR_PREFIX}/share"
 install_prefix="${MESON_INSTALL_PREFIX}/share"
 
-mkdir -p "${destdir_prefix}/gnome-shell/theme/"
-ln -sf "${install_prefix}/themes/${project_name}/gnome-shell" "${destdir_prefix}/gnome-shell/theme/${project_name}"
+mkdir -p "${destdir_prefix}/themes/${project_name}"
+ln -sf "${install_prefix}/gnome-shell/theme/${project_name}" "${destdir_prefix}/themes/${project_name}/gnome-shell"
--- a/gnome-shell/src/light/meson.build
+++ b/gnome-shell/src/light/meson.build
@@ -1,5 +1,5 @@
 # destination directory
-install_dir = join_paths(gnomeshell_alt_themes_dir, meson.project_name(), 'gnome-shell')
+install_dir = join_paths(gnomeshell_default_theme_dir, meson.project_name())
 
 # generate .css files
 theme_sources = files([


chrisjbillington:
Yes you did, and my Arch package builds with these changes.

I'm a little confused about the changes though. You've reversed which way the link goes for the default theme, but since the dark shell theme and the symlink to the light shell theme are still in `/usr/share/themes/Yaru{,-dark}`, isn't there still a dependency on the gtk theme package which owns these directories?

I didn't realise until now, but Arch packages don't appear to care which package owns a directory - or perhaps the package manager is happy with both packages owning the same directory so long as its permissions are identical or something like that. Because I've been [splitting](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=yaru-git#n32) Yaru into separate sound, icon, gtk, shell theme and sessions packages and haven't encountered this issue. I guess Debian packages are different.

clobrano:
It's a mix of problems and formalities.

> I'm a little confused about the changes though. You've reversed which way the link goes for the default theme, but since the dark shell theme and the symlink to the light shell theme are still in /usr/share/themes/Yaru{,-dark}, isn't there still a dependency on the gtk theme package which owns these directories?

It isn't a problem of directory ownership. In fact installing first gtk deb and then shell one, it works fine, but I didn't want to put a dependency between those two packages.

The deb problem seemed that the `/usr/share/themes/Yaru` folder, where gnome-shell is installed in a subdir `gnome-shell`, does not exist until gtk-theme creates it.

Moreover, asking on IRC, I understood that the default theme place is under `/usr/share/gnome-shell/theme`, so Yaru is expected to be there. Secondary/3rd-party themes are expected (by UserThemes extension) to be under `/usr/share/themes/.../gnome-shell`, so being Yaru-dark and alternative, I thought it was a good place for it.

I didn't have time to test whether this actually fix deb install, however, and I am open to other ideas :)

chrisjbillington:
> It isn't a problem of directory ownership. In fact installing first gtk deb and then shell one, it works fine, but I didn't want to put a dependency between those two packages.

Right, but can the shell theme package create the directory as well? In Arch, two packages can both create the same directory and it is not considered a conflict - both packages "own" the directory, and the directory will be removed from the filesystem only once both packages are uninstalled (thousands of packages "own" `/usr/` for example).

>  I understood that the default theme place is under /usr/share/gnome-shell/theme, so Yaru is expected to be there.

Eh, I would be happy that putting a link there satisfies this expectation. But this is bikeshedding and up to you guys.

chrisjbillington:
I think I see the problem - sorry for waffling on about things that might not have been relevant. It's not about the parent directory at all, it's about the fact that the gtk theme package is shipping the shell themes.

This PR moves the default shell theme back to `/usr/share/gnome-shell` where it will be picked up by the gnome-shell package build, and shipped with the gnome-shell package. However, it still leaves the dark shell theme in `/usr/share/themes/Yaru-dark`, where it will be picked up by the gtk theme package and shipped with it. That still seems wrong. The gtk theme package should only ship gtk themes, and the shell theme package should ship all shell themes.

It seems to me that `yaru/debian/yaru-theme-gtk.install` should change to only ship the gtk themes. Instead of containing the single line `usr/share/themes/`, it should be:
```
usr/share/themes/Yaru{-light,{,-dark}/{gtk-{3.0,3.0,3.20},index.theme}}
```
(or the full list if brace expansion not meaningful in this context) to pick up only the GTK related subdirs of `/usr/share/themes`.

Then `yaru/debian/yaru-theme-gnome-shell.install` should change to
```
usr/share/gnome-shell/
usr/share/themes/Yaru{,-dark}/gnome-shell
```
(Or in full if brace expansion is not allowed)

Then the link can go in either direction, it won't matter since both the link and its target will be in the shell package.

chrisjbillington:
Testing seems to reveal that the link really does need to be where you've put it and not the other way around. So this PR fixes that and moves the default theme into the shell-theme package. To move the dark theme and the link out of the gtk theme package and into the shell theme package (so that you don't need to install the gtk theme in order to see the shell themes in gnome-tweaks), see #1597.

chrisjbillington:
You're welcome!

> update the *.install files (which you already did with #1597)
> merge this PR as it is?

Yes, I think this is right.  But I am not super well versed in packaging either (and moreso Arch than Debian-based), so I can't guarantee that I've understood correctly!

PR #1604, opened 2019-11-03
Darken progress blue in dark themes

Feichtmeier:
- $selected_bg_color darkens by 8% in the dark theme
- $progress_bg_color didn't, thus being a bit too bright in the dark themes
- darkening it also by 4% fixes this

Result is a slightly darker blue in the dark theme
image
Comparison to the light theme:
image

@clobrano and/or @madsrh and/or @ubuntujaggers please see if you like this

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -140,8 +140,8 @@ StScrollBar {
   -barlevel-height: 0.2em;
   -barlevel-background-color: transparentize($fg_color, 0.9);; //background of the trough
   -barlevel-border-color: $borders_color; //trough border color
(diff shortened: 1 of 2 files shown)


clobrano:
If the difference is the one in "pick", is barely noticeable. How are you sure this fixed it?

Feichtmeier:
Yeah we could definetly darken it more.
Currently it's darkened by 4% because I thought this is how selected_bg_color does it.
Bu maybe that's not enough for that bright blue. I try more

Feichtmeier:
@clobrano updated the PR description with new pictures: now it's darkened by 8% in shell and gtk instead of 4% - what do you think?

clobrano:
LGTM :+1:

PR #1609, opened 2019-11-05
Add symlinks for remmina - needs testing

### LINKED ISSUE #1329: Small Remmina icon issues
edit: 
working now ~~1) the full color icon is not being used in disco because it now uses the new gnome icon name convention of Icon=org.remmina.Remmina~~

2) the symbolic icon is missing and instead the new remmina icon is being used
image

Feichtmeier:
Screenshot from 2019-11-05 14-31-09

With the symlink the panel icon uses our symbolic icon... but...
We should eventually test if the "gray" means "not connected" and if the color of the panel icon changes on connection. I can't do this right now, but maybe later or @clobrano 

Closes #1329

### DIFF
--- a/icons/Suru/scalable/apps/remmina-panel.svg
+++ b/icons/Suru/scalable/apps/remmina-panel.svg
@@ -1,0 +1,1 @@
+remmina-symbolic.svg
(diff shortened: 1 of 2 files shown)


clobrano:
> We should eventually test if the "gray" means "not connected" and if the color of the panel icon changes on connection.

The taskbar icon stays gray once connected, but I don't think this is blocking.

Feichtmeier:
Hi @antenore do you have any idea why the SVG is colored gray in the systray?
It is made exactly like other symbolic icons, which take up the color defined by the shell. 
Also the remmina symbolic icon, which is the symlinks (remmina-panel) src, takes up the defined color in the shell panel. :man_shrugging:

antenore:
@Feichtmeier no, there's no reason to be grey

PR #1611, opened 2019-11-06
Fix for grey Remmina panel icon that doesn't use symlink

ubuntujaggers:
This seems to work fine:

image

...but I notice that the symbol isn't great anyway, so I'll try to draw a clearer one for both positions when I get chance :+1:

### DIFF
--- a/icons/Suru/scalable/apps/remmina-panel.svg
+++ b/icons/Suru/scalable/apps/remmina-panel.svg
@@ -1,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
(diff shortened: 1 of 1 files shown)


Feichtmeier:
My only problem with this is that you just put this into the Suru folder without a way of reproduction :D

ubuntujaggers:
I see what you mean, I don't know how we can add something to source-symbolic that'll export as we want :/ is there a property (similar to "alert" that ends up red) that would be the right colour? Can we add one?

ubuntujaggers:
Oh wait... I think I can just add the white one to source-symbolic. The whole point is that the icon is handled as a normal colour one, so there's no need for a property to tell it what colour to be. We just make it the desired colour and that will be respected (which is why the symlinked one showed as grey). I'll do that tonight 👍

PR #1622, opened 2019-11-15
Refresh from upstream gtk theme + green darkening

Feichtmeier:
- new check radio styling

image

image

- improved emoji picker
image

- a fix for not-CSD menus

--> additionally we needed to darken the green like we did with the orange and the blue to make the press effect of the checked checks less hard and also to adapt it to the other colors, which we didnt until now
image

image


Feichtmeier:
Needs one more fix from @nana-4 :)
https://gitlab.gnome.org/GNOME/gtk/merge_requests/1186
Thanks nana btw

Feichtmeier:
grafik

PR #1623, opened 2019-11-15
Bionic: Refresh from upstream gtk theme + green darkening

Feichtmeier:
- new check radio styling
- improved emoji picker
- a fix for not-CSD menus

--> additionally we needed to darken the green like we did with the orange and the blue to make the press effect of the checked checks less hard and also to adapt it to the other colors, which we didnt until now

Cherry-pick from https://github.com/ubuntu/yaru/pull/1622


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr1623`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr1623` channel to test your snaps with those changes.

PR #1633, opened 2019-11-20
Fix link backdrop color

clobrano:
Link backdrop color was orange. Using our blue instead and fixed the other configuration that was still using selected_color as base color.


Feichtmeier:
Ohhh good catch
I would not change this in _common though :(

clobrano:
Right

clobrano:
> Right

Well, kind of, since - for Adwaita - `selected_color` is the base for the link one, but not exactly the same. Let's keep this open for a while waiting for a reply from upstream

Feichtmeier:
Do you plan to make a merge request upstream?
If you do, they always want one for gtk-3-24 and one for master :|

clobrano:
I know, but I just discovered that my jhbuild in fedora doesn't work anymore

clobrano:
https://gitlab.gnome.org/GNOME/gtk/merge_requests/1201

Feichtmeier:
Nice!
Why do you use "white" instead of "$selected_fg_color" ? :)

clobrano:
Because there is no valid "fg color" counterpart for link and selected_fg_color is white.
I expect the review to bring this up as well, so to discuss a better replacement eventually.

PR #1639, opened 2019-11-26
Icons: document-export, document-import, document-open-recent, document-revert

### LINKED ISSUE #1621: Qt app, kid3-qt, uses legacy icons "document-*.png"
The Qt app, kid3-qt, uses these legacy icons:

- document-revert.png
- document-open-recent.png
- document-import.png
- document-export.png

However, they are rendered as a normal document icon under Yaru (except for "document-open-recent.png" because it is rendered as legacy "document-open.png").

Qt apps only have a problem with legacy icons starting with "document-*" and all other legacy icons are inherited from Humanity.

Is there a general solution to force Qt apps to choose the closest icon name from Yaru or Humanity?

Should this bug be issued against Qt?

For example, here are two icons in Adwaita:

document-revert.png
grafik

document-revert-rtl.png
grafik

ubuntujaggers:
document-export:

image

document-import:

image

document-open-recent:

image

document-revert:

image

image

Can't seem to take a screenshot when the dropdown menu is open  :)  but here's an extremely high quality photograph of them on the menu, including ones that aren't visible above:

image

Can't really tell in that picture, but they actually look nice on the menu!

(I note that apps using the icons on menus use the 16px versions so we have to keep those, even though it's annoying that Inkscape uses the first and therefore smallest version it finds.)

Closes #1621


clobrano:
Great work @ubuntujaggers!
I only have one doubt. Do you think it's worthwhile enforce the difference between "import" and "export" icons inverting the direction of the red arrow? I was thinking something like the "import" one, but with the arrow red and in the other direction.

ubuntujaggers:
@clobrano can you do a doodle to show?

Feichtmeier:
I think he means

```
_____
->
_____

_____
<-
_____
```

instead of

```
_____
-------->
_____

_____
<------
_____
```

:smile_cat:

clobrano:
> I think he means
> 
> ```
> _____
> ->
> _____
> 
> _____
> <-
> _____
> ```
> 
> instead of
> 
> ```
> _____
> -------->
> _____
> 
> _____
> <------
> _____
> ```

awesome design! Yes, I meant that

ubuntujaggers:
Looks a bit revert, undo, go back:

image

Maybe if it wasn't red?

madsrh:
IMHO the red (associated with danger) isn't really fitting for export - it's not that dangerous 😉  Also using red for both export from revert might be confusing. 

Would it be better if it was blue? 🤷‍♂BUT your red arrow fit well with legacy Humanity icons

image

Feichtmeier:
Now that I see it I agree that reversing the arrow is not a good idea. Indeed it looks like "reverse" / "undo"

ubuntujaggers:
Purple also works...

Feichtmeier:
I think semanticly blue makes the most sense as it's some kind of progress
Green for import fits also as it's something new

clobrano:
> Looks a bit revert, undo, go back:

Indeed, mine was a bad idea :)
I should've searched on most common export icon first. Original proposal is quite common :+1:

ubuntujaggers:
Okay, I'll turn first one blue and push 👍

ubuntujaggers:
This is the original document-export arrow in blue:

image

Change is pushed, so merge if you like :+1:

clobrano:
Thanks @ubuntujaggers!

ubuntujaggers:
I get an error message when I try to cherry pick this one for Bionic :thinking:

clobrano:
can you post it here?

ubuntujaggers:
It's all right - Frederik beat me to it as part of #1645 :+1+

PR #1652, opened 2019-12-04
Icon theme: smartphone icons for "phone"

ubuntujaggers:
I plugged my Android phone into my laptop today and noticed two things:

- The phone-symbolic icon is a feature phone with keys rather than a touchscreen smartphone - it even has an aerial!  Surely very few people are plugging feature phones into laptops?

- There's no full colour device icon, so the symbol appears on the launcher.

This PR makes phone-symbolic a smartphone and adds a full colour icon.  It'll show a conflict with the Remmina PR, but it'll be fine as long as **we merge Remmina first, and then this one** (I took the Remmina icon out of source-symbolic in this branch too).

image

image

image


Feichtmeier:
Nice catch and nice icon! :rocket: 

There are still conflicts :thinking:

ubuntujaggers:
I'll build from master and do a new PR tonight 👍

PR #1653, opened 2019-12-07
Add phone icons

ubuntujaggers:
Same as #1652 without conflicts.

image

image

image


Feichtmeier:
Nice!
I made https://github.com/ubuntu/yaru/pull/1654 to test it on my host, since it's easier to test it with the phone

Feichtmeier:
Hmmmm doesn't work for "samsung" at least
Bildschirmfoto von 2019-12-07 21-38-51

For some reason it uses the icon "multimedia-player" :thinking:

clobrano:
Probably because it shows itself as mpt player to the system

ubuntujaggers:
Weird - we could symlink it? Really the icon could stand for "pocket touchscreen device".

clobrano:
It seems to work with my phone (installed from source)

image

fcole90:
Hi, just  a comment, the icon looks very nice, but I think the reflections are a bit excessive and make it stand out compared to other icons. They seem to give it more a much more realistic look compared to other icons.

ubuntujaggers:
You might be right - I made the reflections brighter and brighter because it looked a bit drab without, but maybe the answer is a different screen colour.

ubuntujaggers:
@Feichtmeier  @clobrano @madsrh @fcole90 

Do you prefer this:

image

image

?

madsrh:
I really like these icons @ubuntujaggers 
Does some of them deliberately have different aspect ratio? The one in the dock screenshot above looks perfect, but one like `icons/Suru/256x256@2x/devices/phone.png` looks a bit more like a tablet IMO. Is that deliberate?

image

Sorry for nitpicking on the beautiful work (again)!

ubuntujaggers:
We can make the big one slimmer - I just copied the proportions of the existing keyboard icons rotated 90° - no reason not to tweak further :)

PR #1657, opened 2019-12-10
[HOLD] apps: increase difference between gnome-weather buttons

clobrano:
Currently it is hard to differentiate between checked and unchecked
buttons in gnome weather, since the (dark) colors are very similar. In a
previous Yaru version, this app was customized for this very same
reason.

Adding back button customization in apps scss file, using the same logic
Yaru dark buttons have (darker means checked, lighter means unchecked),
forcing the colors a little bit, since the bright color around the
buttons (if the weather is clear, the background is blue sky).

closes LP:1807508

Before
image

After
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -528,7 +528,7 @@ UnityDecoration {
   -UnityDecoration-title-indent: 0px;
   -UnityDecoration-title-fade: 0px;
   -UnityDecoration-title-alignment: 0;
-  
+
   // Unity "titlebars"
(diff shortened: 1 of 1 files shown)


clobrano:
Actually, let's keep this in pause until I see if this is to be fixed upstream too

madsrh:
I don't know how they color these buttons but the constrast is bad even in Adwaita.

Also "text color" for the disabled < button on blue backgrounds is hard to read. It's fine for overcast ☁️ 

image

clobrano:
This seems already fixed in Fedora 31 with both Adwaita and Yaru

PR #1658, opened 2019-12-10
icons: fix warning color on battery icon 30/40 perc

### LINKED ISSUE #1646: Battery icon is red when the laptop has 30% battery left; it should only turn red when it hits 10%
first reported here https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1854179

clobrano:
upstream icons show a warning color only at 20 percent battery or below,
while current Yaru icons shows it at 40 or below, which is not a warning
battery level.

- replace warning (orange) color from battery icons at 30 and 40 percent
  with the same color the icon at 50 percent and above have.
- replace class "error" with class "warning" in 20 percent battery icon
- fixed typo in icons/src/scalable/README.md

closes #1646


ubuntujaggers:
Actually, I'll see if it lets me push now - the change doesn't affect anything you've done and we can always put it back :+1: and this way I can delete the branch off my laptop and do another issue.

ubuntujaggers:
Okay, I've seen the them all working except the 10% alert one, and there's no reason why that one wouldn't work correctly :)

clobrano:
Great! Thanks for testing and fixing source symbolic. Can I merge?

clobrano:
Oh wait, there are conflicts

ubuntujaggers:
Yep, let's do it :) Then I will make a PR for the new symbol categories.

ubuntujaggers:
Ohh, must be source symbolic.

ubuntujaggers:
Can you revert source symbolic and I'll do one PR for all changes to that file at once? Sorry, my fault.

clobrano:
@ubuntujaggers, I reverted your commit and tried to replicate it myself, could you have a look?

ubuntujaggers:
Looks good - really the 20% rectangle should be yellow so it's obvious that it's warning (not alert) but merge it, I have to add some categories so I'll just make it yellow when I do that :+1:

PR #1670, opened 2019-12-18
Unity/sync

clobrano:
Sync Unity theme with current GTK/Shell Yaru, for what is possible.

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -512,7 +512,7 @@ $titlebar_bg_color: $headerbar_bg_color;
 $titlebar_fg_color: $headerbar_fg_color;
 
 UnityDecoration {
-  -UnityDecoration-extents: 28px 0px 0px 0px;
+  -UnityDecoration-extents: 32px 0px 0px 0px;
   -UnityDecoration-input-extents: 10px;
 
   -UnityDecoration-shadow-offset-x: 1px;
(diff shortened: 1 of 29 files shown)


clobrano:
Some screenshoot

image
image
image

For some reason, you need to install light-themes to have fully functional shadows in windows and menus. 
Also, as you can see above, some application adds a titlebar above the headerbar, while others don't.
This happens in Nautilus, but it's worse in gnome-software, where the round left corner is even visible
image

I don't think this is due to Yaru, since happens with other themes too and quite interesting it happens under GNOME too if I switch to Ubuntu session without rebooting

I think I'll try to get into contact with people involved in Unity, but this can be reviewed and merged if you agree

Feichtmeier:
Agreeing totally. Nice work!
I'm tempted to go back to unity now lol

clobrano:
doh, I forgot to change the build script to install the assets

fcole90:
How about removing the glossy effect on the Unity launcher? It makes it feel much more modern. I have seen other themes doing it, like Numix, doing it.

Example of a derivative here: https://www.deviantart.com/chrisfr06/art/Numix-Blue-GTK-Theme-Modded-without-Unity-boxes-450739876

image

clobrano:
Indeed would be better. On my backlog now

madsrh:
Looks good @fcole90 👍 Nice suggestion

clobrano:
@fcole90, please have a look at #1676

PR #1671, opened 2019-12-19
gnome-shell: Sync with 3.34.2 changes

3v1n0:
Let's keep tracking stable version for now

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -615,6 +615,13 @@ StScrollBar {
     border-bottom-style: solid;
   }
 
+// Rename popup
+.rename-folder-popup {
+  .rename-folder-popup-item {
+    spacing: 6px;
+    &:ltr, &:rtl { padding: 0, 12px; }
+  }
+}
 
 // Rename popup
 .rename-folder-popup {
@@ -755,7 +762,7 @@ StScrollBar {
 
   .ws-switcher-active-up, .ws-switcher-active-down,
   .ws-switcher-active-left, .ws-switcher-active-right {
-    height: 50px;
+    height: 52px;
     background-color: $selected_bg_color;
     color: $selected_fg_color;
     background-size: 32px;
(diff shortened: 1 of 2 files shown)


clobrano:
For reference, the main changes I noted are

- today in calendar is a full circle instead than just a border
- selected color border around circular buttons (close, shutdown, etc...), which looks a bit odd

but it looks good to me overall

Feichtmeier:
I think we might need the corresponding shell version to test this.

Do you have a deadline for this to be merged @3v1n0?

clobrano:
This is for a SRU, so I think it is expected to work with 3.32 as well

https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/1857037

gnome-shell/src/gnome-shell-sass/_common.scss
```diff
@@ -1216,12 +1230,11 @@ StScrollBar {
     &:hover, &:focus {
       background-color: $_hover_bg_color;
       color: $fg_color;
-      border: none;
-      padding: 14px;
     }
     &:active { 
       background-color: $_active_bg_color;
       color: $fg_color;
+      border-color: $selected_borders_color;
```
Feichtmeier: Ah, no this makes sense for upstream but not for us. @3v1n0 Could you remove this line again?
They have blue system menu disks so this makes sense for them

iainlane:
I have just pushed 3.34.3 to focal, so it would be good to get this PR tidied up soon.

And I'd say it's required before uploading the SRU to eoan.

gnome-shell/src/gnome-shell-sass/_common.scss
3v1n0: Ok, sorry for being late, but you know... Holidays :)

Anyways, fixed now.

3v1n0:
> And I'd say it's required before uploading the SRU to eoan.

@iainlane agree, the eoan (and actually ubuntu/master too) for eoan is then at 
https://github.com/ubuntu/yaru/tree/ubuntu/eoan

Feichtmeier:
Working - thanks!
grafik

PR #1676, opened 2019-12-21
unity: remove glossy effect on application icons in dock

clobrano:
Remove glossy effect on application icons in dock

image

suggested  https://github.com/ubuntu/yaru/pull/1670#issuecomment-568168574 in #1670

@madsrh, what do you think?


clobrano:
I decided to move step by step here. First I removed only the glossy effect and kept the colored background on opened applications. I like it, but it's also quite colorful and might be annoying.

Note: if the color is removed, also the arrows that indicate which app is open and which one is on focus must be improved

madsrh:
I like it but it looks a bit messy/cluttered :thinking: maybe it's because the icons in the dock was a bit smaller in the screenshot by fcole90. IDK

clobrano:
Respect gnome dock is a bit more messy indeed, but I think without the glossy effect is better

PR #1677, opened 2019-12-22
Icon theme: suggested improvements to consistency of devices (keyboard, server)

ubuntujaggers:
When I looked at Carlo's Unity screenshot I realised that the server (including the redesigned icon) uses a warmer grey from the Suru palette, whereas the rest of the devices use cooler tones.  So, I updated the new server to use the same gradient as (e.g.) the SSD drives:

image

I also note that the keyboard icon uses a very pale gradient and doesn't look great on the app grid, it's sort of ethereal:

image

Also, the lettering on the largest size looks lovely in Inkscape but is lost when you actually see it on the app grid.  So, I removed the letters for improved clarity and took the darker gradient from some of the other devices.  **EDIT:**  I'm now having doubts about the colour change because the mouse is pale, so I suppose there could be consistency between grey disk devices and paler peripherals - so now I've removed the lettering, kept the original gradient, but made the keys stand out more.

image

image

image

What do you think, is this alignment an improvement?  No need for us to merge if not :+1:


Feichtmeier:
I like it! 

( Maybe as a consequence of these steps you could make your own color palette out of the canonical / Ubuntu colours on their website? Currently we have basically two colours palettes )

clobrano:
It might make sense to have two palettes, one for the theme and one for icon set. up to you @ubuntujaggers

ubuntujaggers:
Rethought the keyboard a bit...

ubuntujaggers:
Does anyone object to me merging this one?  @Feichtmeier @madsrh @clobrano ?

PR #1684, opened 2019-12-31
Icon theme: fix proportions on 256px printer icon

ubuntujaggers:
This is not a "reimagining", which still might be good to do, but I think I've fixed the proportions on the 256px icon so it matches the other sizes more and looks less squat.  I think it's less ugly, so merge for now if you agree, and don't merge if not :+1:  :)

**Before:**

image

**In this PR:**

image


madsrh:
Yes! 🎉

PR #1687, opened 2020-01-02
Icon Theme: Possible new icon for Transmission

ubuntujaggers:
Just an idea, ignore if you don't like it - we have our own icon for Transmission but it's not great at all. I wonder if we'd be better doing what some other themes do (e.g., Paper) and just have a simple glyph.  The compression isn't very flattering on the screenshots so have a look at the actual files to compare.

image

image

image


clobrano:
I am not sure about this. For my point of view, since it is quite abstract, it is not immediately recognizable as Transmission. Should we involve the maintainer first?

madsrh:
> ...it is not immediately recognizable as Transmission...

I had the same thought

ubuntujaggers:
Good point, I forgot Transmission was one of the ones where we had maintainer permission. Maybe we can keep the same design as current but make it a bit "slicker"?

ubuntujaggers:
@madsrh @clobrano What if I push a version like this:

image

image

Same basic design as current in master but just a bit "sleeker"?

clobrano:
I personally prefer the current one :relaxed:

ubuntujaggers:
> I personally prefer the current one

Colours or shapes or both...?

clobrano:
Right, I said nothing useful :D

About the shapes, I see the new one a slightly better, but overall I prefer the colors and the fold effect of the current one

ubuntujaggers:
Hmm, maybe I can fuse elements of both :thinking:

ubuntujaggers:
@clobrano How about

Current:

image

Suggested:

image

clobrano:
Very nice for me :+1:
Do you remember the transmission maintainer's nick (I'm on mobile right now)?

ubuntujaggers:
I'll find it and tag him in after I've edited the first post.

clobrano:
I found the nicks in the previous PR :)

Hey @livings124, @mikedld,

sorry if I ping you here, but I found your nicks in the previous Transmission icon PR (#1426)
@ubuntujaggers worked on an iteration of the icon, which I think it's really good, what's your opinion, can we use it inside Yaru?

Thanks!

Feichtmeier:
Imho the outlines are too strong on this one

ubuntujaggers:
Ah, let's leave it for now, the latest version isn't massively different anyway :)

PR #1689, opened 2020-01-04
Adding icons and symlinks for full colour tick and cross actions

### LINKED ISSUE #1686: Missing icons: checkmark and close-X action icons
In the Ubuntu installer (ubiquity-frontend-gtk) there are green checkmarks showing up if you filled a field successfully :heavy_check_mark: 
Those are still taken from the Humanity icon theme and could use an update

Screenshot from 2020-01-02 12-44-28

It looks like two symlinks are also needed. As a bonus the X-icons could also be updated 
image

The list could be:

**Positive checkmarks:**
- [x] dialog-apply (THIS is the icon, the rest are "only" symlinks)
- [x] dialog-yes
- [x] dialog-ok
- [x] gtk-ok
- [x] gtk-yes
- [x] stock_calc-accept
- [x] stock_mark
- [x] stock_mail-filters-apply
- [x] stock_yes

**Negative cancel X icons**
- [x] window-close (THIS is the icon, the rest are "only" symlinks)
- [x] cancel
- [x] dialog-close
- [x] dialog-no
- [x] dialog-cancel
- [x] button_cancel (with _ not -)
- [x] gtk-close
- [x] stock_calc-cancel
- [x] stock_close

@ubuntujaggers and/or @madsrh please have a look in /usr/share/icons/Humanity/actions

ubuntujaggers:
As @madsrh notes, we will likely be changing the icon itself to match new colours, but this PR has all the symlinks etc. for testing.  If it works in Ubiquity we can consider whether to include the symlinking of the dropbox emblems in this PR too.  Fixes #1686 .


Feichtmeier:
Awesome Stuart!
This will be a bit tricky to test since the page with the entries comes after you partitioned some drive :D
I'll try with an usb stick

Feichtmeier:
Should we eventually exchange the icons with these two?
https://github.com/ubuntu/yaru/issues/1686#issuecomment-570265395

@madsrh @ubuntujaggers

ubuntujaggers:
> Should we eventually exchange the icons with these two?
> [#1686 (comment)](https://github.com/ubuntu/yaru/issues/1686#issuecomment-570265395)
> 
> @madsrh @ubuntujaggers

Done and pushed.

I've not touched the corresponding folder emblems yet - I think that's a different PR where we look at all the emblems (for instance: recolour all of them and make the glyphs less hairline to match these two?).

Feichtmeier:
@ubuntujaggers thanks! I don't find a way to test this except creating a new VM since those fields only come after you partioned your disk (in the installer)
Maybe you could double check if you did everything correctly from your perfection and then merge it?
This can not really break anything and we are in the dev cycle so enough time to fix

PR #1710, opened 2020-01-09
Use Aubergine as second accent color instead of blue

### LINKED ISSUE #1723: Shell slider color must be visible with hover
We should pick a color for the shell sliders that has good contrast against the background and also when hovered. A border might also help.

Current:
image

`#8e3773` with a border
image

clobrano:
As per PR title, this change is a try for replacing the blue with aubergine. The color relationships between the widget that used blue are kept the same, that is:

- checkradio, switch and progress bar share the same dark tone (aubergine in this case)
- focus ring is a lighter variant than the color for the widgets above
- **link color is still blue**, since aubergine is usually used for "visited link" and might be misunderstood  

TODO:
- [x] fix #1723
- check what gtk2 is using (probably only orange, but let's double check) -> see #1730 
- [x] shell dark: slider handle is too dark
- gtk dark: slider-has-mark... is too dark (review the assets) -> see #1728
- [from @madsrh] consider #6e3c61 as base aubergine (for checkboxes, sliders, switches) -> see #1731 
- [x] fix scrollbar through color on osd

EDIT: to avoid blocking development too much, I opened bug/discussion tickets for the remaining open points, so that we can merge this PR if you also think it's ready.

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -9,7 +9,7 @@ $fg_color: if($variant == 'light', $inkstone, $porcelain);
 $selected_fg_color: #ffffff;
 $selected_bg_color: if($variant == 'light', $orange, darken($orange, 4%));
 $selected_borders_color: if($variant== 'light', darken($selected_bg_color, 15%), darken($selected_bg_color, 30%));
-$borders_color: if($variant == 'light', darken($bg_color, 18%), darken($bg_color, 10%));
+$borders_color: if($variant == 'light', darken($bg_color, 20%), darken($bg_color, 9%));
 $borders_edge: if($variant == 'light', transparentize(white, 0.2), transparentize($fg_color, 0.93));
 $link_color: if($variant == 'light', darken($selected_bg_color, 10%), lighten($selected_bg_color, 20%));
 $link_visited_color: if($variant == 'light', darken($selected_bg_color, 20%), lighten($selected_bg_color, 10%));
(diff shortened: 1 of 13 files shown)


ubuntujaggers:
Looks great!  Are volume and brightness sliders in the top right still meant to be dark blue?  They are for me...

Feichtmeier:
Yeah I guess it's not in the shell yet

flexiondotorg:
Testing this branch, looks fab. Just need the shell to get some aubergine to complete to look :-)

clobrano:
I'll complete it with the shell part :+1:

clobrano:
image

Feichtmeier:
Could you set the 
```$switch_border_color: darken($switch_bg_color, 35%)```
Before you merge?
It improves the crispness of the switches in the dark gtk theme :D

Feichtmeier:
@clobrano I think the not checked sliders also need some tiny color corrections. Background, border and the knob looks brighter for the light theme in gtk I think
But maybe this whole nit picking can be done after this PR, even though I do not think we are in a hurry at all

ubuntujaggers:
Loving the purple sliders :+1:

clobrano:
After having synchronized the slider colors between gtk and shell, and more specifically the color of the through, I noticed this side effect

image

now the through's color is the same as the background of the osd. Mind if I lighten up the osd to match the shell popovers?

Feichtmeier:
I would rather use osd_fg_color or a slightly darkened white for the active trough
The purple doesn't have a good contrast on the dark and if you brighten up more, it's magenta

clobrano:
I'll try it. It might clash with the overdrive, but let's see

clobrano:
to avoid blocking development too much, I opened bug/discussion tickets for the remaining open points, so that we can merge this PR if you also think it's ready.

PR #1716, opened 2020-01-10
Select branch for autorebase

didrocks:
Switch gtk to gtk-3-24 one.
GNOME Shell tracks master.


didrocks:
Let's merge it. We now are able to track and keep a stable PR version without deleting and recreating a branch and PR every time.
Bug fixes in actions V2 are required for this to work.

PR #1724, opened 2020-01-11
shell: white lock screen cancel button

clobrano:


Feichtmeier:
Why not $bg_color?

I think this way it's white in both dark and light variants.

clobrano:
I can try what you propose, but lock screen does not change on dark/light variant

Feichtmeier:
I agree that it should be one design for both variants! No idea what colour though. Maybe rather dark by default? The bright could too bright

The login/Lockscreen Changes with the variant changed in Gnome-Shell.scss

clobrano:
I understood that the outcome of the design fest was to make it white. I'll try some variants next week so that the whole team is available to have a look.

Feichtmeier:
Okay if it's white eventually take the new bg_color colour f7f7f7 from vanilla

clobrano:
> Okay if it's white eventually take the new bg_color colour f7f7f7 from vanilla

I used `$porcelain` which is `#F7F7F7`

Feichtmeier:
Could you please insert a screenshot. Back then we tried this and I'm pretty sure everyone agreed that the bright white was the most visible thing on the whole screen with this dark purple bg

clobrano:
> Could you please insert a screenshot. 

Sure

Feichtmeier:
image

What about using the $destructive_color for the cancel button? @madsrh @clobrano 

Or if red is too strong and canceling the password is not destructive enough - some sort of lightened black?
image

clobrano:
Screenshot from 2020-01-14 14-27-42

This is current PR.
I am sorry if I said differently before (I don't remember now), but I like this.

@madsrh, @ubuntujaggers, @flexiodontorg what do you think?

Eventually we can have it under test 👷 for a while before taking a definitive decision

madsrh:
I tried a red button in the my first [mockup ](https://ubuntucommunity.s3-us-east-2.amazonaws.com/original/1X/632398c0bea53875fbfeb807527fc48d894da20e.jpg)and somewhere there's a comment (maybe from MPT) saying that it isn't a destructive action. Anyway, looking at the screenshots the light button is sticking out like a sore thumb, but when I tried the PR it felt really natural for that button to be light.

ubuntujaggers:
Looks good to me @clobrano :) :+1:

Feichtmeier:
I'm out of thumbs then - feel free to merge :)

PR #1733, opened 2020-01-14
Crisp dark switch border

### LINKED ISSUE #1732: Slider colour is somewhat low contrast in dark theme
Grey folders are a bit low contrast in the dark theme. So are the purple sliders.

Regardless of contrast, I preferred orange and green for folders and sliders/checkboxes. Particularly with checkboxes, sliders, radio-buttons, green meaning on was a breath of fresh air when Yaru first started doing it. It's so obvious because of the association of green with things being enabled/active.

Colour of folders is a personal preference probably going back to the actual colour of Manila folders.

image

Feichtmeier:
- by using a darkened $borders_color instead of the purple base

image

Closes #1732

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -65,7 +65,7 @@ $progress_border_color: if($variant=='light', $aubergine, darken($aubergine, 4%)
 $checkradio_bg_color: if($variant=='light', $aubergine, darken($aubergine, 4%));
 $checkradio_fg_color: #ffffff;
 $switch_bg_color: if($variant=='light', $aubergine, darken($aubergine, 4%));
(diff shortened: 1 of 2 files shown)


clobrano:
Slight variation, but it's surely an improvement

on the left the current master, on the right this PR for anyone that cannot build both fro comparison

image

Feichtmeier:
Is this "fixed" for you @chrisjbillington ? Or should we darken it slightly more? (the border)

chrisjbillington:
It's an improvement for the problem you identified, but it wasn't what I was talking about - I liked the very high contrast of green sliders (and the fact that green = on is a pretty universal association), and I liked having more colours in the theme in general to make a stronger distinction between different elements. So I was thinking about the actual fill colour and not the border.

I understand this is a personal preference though.

Issue #1703, opened 2020-01-08
Fixes and tweaks for the theme


madsrh:
- [X] disabled checkmarks & radio buttons gray 
- [X] change checkmark to Vanilla style 
- [x] Filemenu background should be gray -> #1734 
- [x] hover color for blue checkmark & radio buttons should be more visible
- [x] Remove aubergine from the GDM Cancel button (Shell)
- [x] Yaru 1.0 style tabs
- [x] Menu ~~borders~~ contrast to the area it is drawn on in the dark theme are not strong enough
- [x] 0.5% or 1% lighter buttons (for the headerbar buttons)
- [x] adjust the scale assets to the new colors (also the new focus color ;) )
~~Implement new design for switches~~
Edit by @Feichtmeier : after the design call with mpt we agreed on keeping the switch design as it is
- [x] no borders for checkmarks in file menus and context menus
- [x] grabhandels in the dark theme should be brighter (upstream: https://gitlab.gnome.org/GNOME/gnome-control-center/issues/858, see #1847 for more info)
- [x] Combobox menu’s edges are both 1 pixel too wide -> see #1849 
- [x] Draw a new hand cursor for https://github.com/ubuntu/yaru/issues/855
- [x] NOT-FIXABLE; When a radio menu is opened, the text of the current item should be in the same place.

Feichtmeier:
Eventually the check/radio hover is already fixed
What do you think @madsrh :
Normal:
![image](https://user-images.githubusercontent.com/15329494/72514956-07a04980-3847-11ea-9964-6f4893d8c116.png)
Hover:
![image](https://user-images.githubusercontent.com/15329494/72514997-1b4bb000-3847-11ea-824b-e16508d86f39.png)
![Peek 2020-01-16 10-04](https://user-images.githubusercontent.com/15329494/72515311-957c3480-3847-11ea-9245-4309c5b23835.gif)

Amr-Ibra:
Is this issue on the radar? https://github.com/ubuntu/yaru/issues/1423

madsrh:
@Amr-Ibra yes! Everything is 😄 @ubuntujaggers is currently revisiting the cursor so perhaps he could take a look at this too?

clobrano:
This list comes from the latest design fest, is only a subset of all the open issues and it is by no mean a complete list of goals for 20.04

Feichtmeier:
This issue is mostly about the design fest,... edit: oh nvm look whos there

[...]

PR #1734, opened 2020-01-14
gtk: grey selected filemenu item

clobrano:
Use popover hover color for menuitems inside the menu bar

see #1703 

image
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -128,6 +128,7 @@ menubar,
   > menuitem {
     &:hover {
       color: $fg_color;
+      background-color: $popover_hover_color;
     }
   }
 }


Feichtmeier:
Looking at your tab PR... Would MPT still be okay if we make the hovered menu items look like active tabs?

clobrano:
This means making the menu grey?
I remember the idea was dropped in London though

Feichtmeier:
No making the menu items look like your new tabs from your pr I meant

clobrano:
However, to have the selected menu like the new tab, the rest of the menu should be gray as the unselected tabs are

image

Feichtmeier:
Alright - sorry to diverge the discussion - if this is what MPT wanted then feel free to merge :+1:

madsrh:
@clobrano are you sure that the orange line is needed? 🤔

clobrano:
I think it helps differentiate the menu from the item selected in the drop down

madsrh:
> I think it helps differentiate the menu from the item selected in the drop down

Okay, I think it's a bit like having belt and braces but NP. Perhaps it's because the gray is too dark (I know it's the same as in the menus). Especially comparing it to the Nautilus sidebar.

clobrano:
Let's have it for a little while, then we can rethink about it

PR #1735, opened 2020-01-14
Gtk upstream updates

Feichtmeier:
This takes the changes from https://github.com/ubuntu/yaru/commit/5008061f20347c41f8329c6018d80e671d4babae in gtk/upstream and applies them on gtk/src/gtk-3.20/_common.scss only - it does not touch _drawing.scss (yet) and ofc not our _colors.scss

It contains updates to _common.scss for newer versions of scss

@didrocks do you agree with this workflow - or - did I understand you correctly? :)


didrocks:
:+1: on the workflow

PR #1736, opened 2020-01-14
gtk: notebook's tab is an extension of the notebook

clobrano:
This change re-apply previous Yaru style for tab in a way that they
look like an extension of the notebook they belong to.

NOTE: here "overshoot" is used also as undershoot.
The undershoot is used to style the effect of the notebook's content
sliding below the tab and was removed from upstream

Also, unselected tabs do not look disabled anymore having the same text color and non-bold font

Old below, new above
image

**Old below, new above**
image


Feichtmeier:
Awesome!!!
Could you put this whole section into tweaks?

clobrano:
Let me check

clobrano:
I had a quick look. The downside of moving it to tweaks is that default variant does not use tweaks and uses headerbar.scss instead, that is I should duplicate this same change in both files (and default does not use tweaks to avoid styling the same thing three times and risking conflicts)

Feichtmeier:
Oh this is bad?
The default variant should really include _tweaks.scss as it contains tweaks to many things Oo why didn't we include it

friendlyapple:
The new tabs are cool :) However, is it normal that the background (behind the tabs) is grayer on the first image?
72361738-47a1e800-36f2-11ea-8639-0efcc4dca290

On the second image, the background is lighter:
72361814-63a58980-36f2-11ea-96dc-80220b1f437f

It should rather be (like the previous Yaru style) ?
68747470733a2f2f692e696d6775722e636f6d2f49617538575a6f2e706e67

Feichtmeier:
@clobrano alright please put these changes in _tweaks.scss now - master is ready

madsrh:
Awesome work @clobrano 🎊 

* In the dark theme the background isn't working, so how about doing something like this for the notebook header: 

`background-color: darken($bg_color, 2%);`


Highter than 2% makes the dark theme tab blend with the bg. Lower than 2% can't be seen in the light theme - which I don't mind TBH 👼



* Can we use the dark theme tabs for Terminal again:

image

> However, is it normal that the background (behind the tabs) is grayer on the first image?

* Good point @friendlyapple. I don't care much for that dark background either.

clobrano:
> The new tabs are cool :) However, is it normal that the background (behind the tabs) is grayer on the first image?

It is because the colors in Nautilus have always been a bit paler, but indeed I can try the same color in normal tabs

clobrano:
@madsrh, @friendlyapple, while moving the changes form common to tweaks I also changed bg color in tab's header to make it as in Nautilus.

image


As side effect, dark variant has improved in contrast (just a little)

image
image


do you think this fixes the problem?

gtk/src/default/gtk-3.20/_apps.scss
```diff
@@ -64,7 +64,7 @@ $ambiance: null !default;
     }
 
     notebook > header {
-      background: if($variant==light, $porcelain, $bg_color);
+      background: if($variant=='light', $porcelain, $bg_color);
```
Feichtmeier: I think bg_color is already porcelain but I did not put the variable in the reorganize commit *selfslap*
So you could prly just write $bg_color and adjust the $bg_color to use $porcelain just to be a bit cleaner

clobrano:
image

I think I can merge :D

ubuntujaggers:
Looks great 👍

friendlyapple:
It's awesome!

real-amano:
> Awesome work @clobrano confetti_ball
> 
>     * In the dark theme the background isn't working in this PR, so how about doing something like this for the notebook header:
> 
> 
> `background-color: darken($bg_color, 2%);`
> 
> Highter than 2% makes the dark theme tab blend with the bg. Lower than 2% can't be seen in the light theme - which I don't mind TBH angel
> 
>     * Can we use the dark theme tabs for Terminal again:
> 
> 
> image
> 
> > However, is it normal that the background (behind the tabs) is grayer on the first image?
> 
>     * Good point @friendlyapple. I don't care much for that dark background either.

Would it be possible to just use the terminal background for the tab. Thus be aubergine by default, white if you prefer a white background and so on. That probably has implications for the font readability.

Feichtmeier:
We should just make the terminal default to the dark version in gsettings

Is this possible yet for 20.04 @didrocks ?

didrocks:
yes, it should be. Can you find the right key and I'm happy to do the change :)

clobrano:
this is from dconf-editor

image

PR #1750, opened 2020-01-16
Green color improvements

Feichtmeier:
- vanilla green looks great but it is a little bit too dark for gtk and gnome shell
- Lighten up all greens by 5% (shell and gtk)

(Left old, right upper side new [5% lighter])
grafik

- adapt the green border for mixed theme pressed buttons in the headerbar to look in the other themes: same color as the bg
(Pressed button: )
Bildschirmfoto-20200116063621-384x131

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -17,7 +17,7 @@ $top_hilight: $borders_edge;
 
 $warning_color: $yellow;
 $error_color: $red;
-$success_color: if($variant == 'light', $green, darken($green, 10%));
+$success_color: if($variant == 'light', lighten($green, 5%), darken($green, 5%));
 $destructive_color: if($variant == 'light', $red, darken($red, 10%));
 
 $osd_fg_color: #eeeeec;
@@ -58,8 +58,8 @@ $dash-alpha-value: 0.6;
 $dash-opaque-alpha-value: 0.0;
 
 //special cased widget colors
-$suggested_bg_color: if($variant=='light', $green, darken($green, 5%));
-$suggested_border_color: if($variant=='light', darken($suggested_bg_color, 15%), darken($suggested_bg_color, 30%));
+$suggested_bg_color: if($variant == 'light', lighten($green, 5%), darken($green, 5%));
+$suggested_border_color: if($variant=='light', darken($suggested_bg_color, 5%), darken($suggested_bg_color, 10%));
 $progress_bg_color: if($variant=='light', $aubergine, darken($aubergine, 8%));
 $progress_border_color: if($variant=='light', $aubergine, darken($aubergine, 8%));
 $checkradio_bg_color: if($variant=='light', $aubergine, darken($aubergine, 8%));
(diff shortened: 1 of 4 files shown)


madsrh:
Hmmm, I really don't have a problem with the current green 💚 

I find the current green -  especially for the button - to have a much better contrast (sorry it's the buzzword of 2020), but I think you and @clobrano both prefer this lighter green, right? 
If so, could we maybe find something in between? 🙏 

image

Feichtmeier:
The picture of the button in the PR description is a PRESSED  or :active/:checked button

There are situations where the very dark green has actually a worse contrast.

I did not change the hex value from vanilla, I used the scss "lighten( )" function
https://codepen.io/baudoin/pen/HdliD
You can insert #0e8420 in that page and see if what you like
I tried 5% in this PR

clobrano:
I like current green as well, but it seems a bit too dark one pressed, doesn't it?

green

Feichtmeier:
> I like current green as well, but it seems a bit too dark one pressed, doesn't it?
> 
> green

Yes this is one part of the PR, fixing the mixed theme suggested buttons active/pressed border

madsrh:
> I like current green as well, but it seems a bit too dark one pressed, doesn't it?

I agree that the pressed state is too dark. How about just 2% lighter?

Alternatively, we could only change the pressed state but I'm guessing that's a bad solution.

Feichtmeier:
Only changing the pressed state is a no-go as it's in drawing and I think the rest of the buttons looks really good pressed.

Let's just lighten it up by 5% this current green is really too dark, except - like in vanilla - drawn on a completely white surface

Here is a reality-check photo made with my phone from my crap thinkpad screen (you should try at yours mads ;) )

IMG_0317
IMG_0318

clobrano:
The problem is actually only on the active state, being too dark for green button, so I suggest editing drawing scss in order to let all the buttons be a little less dark only in the active state

for example I modified line 277

https://github.com/ubuntu/yaru/blob/0d2d19ac026eef95b5acddd662d067881b97d182/gtk/src/default/gtk-3.20/_drawing.scss#L277

in this way

```
$button_fill: if($variant == 'light', image(darken($c, 10%)), image(darken($c, 5%))) !global;
```

Feichtmeier:
Nooooo . they look perfect except when they are this dark green. Look, let's just ask what MPT says about lightening up the green. I am pretty damn sure he would understan dthe problem

Feichtmeier:
The green is generally too dark, not only when pressed.
Please don't change drawing any further. We already at the absolute limit and should get some of the changes into upstream, some of them agree on making it more flat anyways

clobrano:
Let's just avoid this "Nooo" walls. I don't think it's good for the discussion.

5% lighter green as base it's probably too bright in normal state, but it's ok for pressed, so I'm good

madsrh:
I would love for the default state to be something less than 5%, but this is not a _make or break_ issue for me 😃

clobrano:
> I would love for the default state to be something less than 5%, but this is not a _make or break_ issue for me

I agree. Ok for 5% for me

PR #1753, opened 2020-01-16
tweaks: add missing undershoot style

clobrano:
In previous notebook's tab redesign, undershoot style was left in
common.scss, so that - after upstream sync - it is now gone.

Add again the style in tweaks.scss


Feichtmeier:
Great!
Maybe we can sit down some day (virtually) and think about what assignment we really need int tweaks and what now.
Switches, Notebooks and scrolledwindows are now doubled fully

PR #1754, opened 2020-01-16
common: reduce entry side padding

### LINKED ISSUE #1705: Remove space infront of caret in entrys
The text in a text field — and therefore the caret in an empty text field — is inset so much (8 pixels) that it’s unclear whether the field is actually empty, or whether it contains a space character.

image

clobrano:
8px of side padding on entries can be mistaken by an empty space. Reduce
the side padding by half (4px).

image

closes #1705

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -276,8 +276,8 @@ spinner {
 entry {
   %entry_basic, & {
     min-height: $_entry_height;
-    padding-left: 8px;
-    padding-right: 8px;
+    padding-left: 6px;
+    padding-right: 6px;
     border: 1px solid;
     border-radius: $button_radius;
     transition: all 200ms $ease-out-quad;


clobrano:
@madsrh, WDYT, is it enough?

Feichtmeier:
It's too much imo
Could we make it somehow evenly padded to top and right?
Also... :) ... can we get this into upstream indead? I introduced this mistake for sure

madsrh:
> It's too much imo

I agree with @Feichtmeier 
This looks odd. Currently we have (roughly) the same padding above, below and before the text. This makes sense IMO. Could you try with only 2px?




> can we get this into upstream instead?

That would be great

> I introduced this mistake for sure

What do you mean?

madsrh:
Just to be clear: you did a great job @clobrano. This bug comes from the MPT review

clobrano:
> Just to be clear: you did a great job @clobrano. This bug comes from the MPT review.

No worries 😉 



> This looks odd. Currently we have (roughly) the same padding above, below and before the text. This makes sense IMO. Could you try with only 2px?

I didn't get it. To have even padding around the text, 8px side padding is probably the solution. If 4px side padding is already too much, 2px side padding is really close to the entry border.

> Also... :) ... can we get this into upstream instead? I introduced this mistake for sure

We can have it upstream **also** 😄

madsrh:
> I didn't get it. To have even padding around the text, 8px side padding is probably the solution. If 4px side padding is already too much, 2px side padding is really close to the entry border

Lost in translation 😄 I meant only reducing it by 2px

clobrano:
> Lost in translation 😄 I meant only reducing it by 2px

ahh 😄 

> We can have it upstream **also** 😄

I meant, once we agree on a change here, I will propose it upstream too

clobrano:
It's probably a bit odd as well

image

is it just me, or do you notice the top padding is a bit different from bottom? I am looking at  the uppercase letters.

madsrh:
I'm fine with [your suggestion](https://github.com/ubuntu/yaru/pull/1754#issuecomment-575139890) above, but I would also be fine with 8px too 😄  @Feichtmeier ?

Feichtmeier:
I can basically copy paste your comment: I'm fine with the current entries, but I think this is a tiny improvement which is visual if you know where to look at.

I personally don't think that this is the time to change common for those paddings.

If you do this upstream, I would try to make it somehow inheriting the $_entry_height or whatever weird variable I had come up with at these days @.@ sorry for the sloppy implementation
$_entry_height changes depending on $default or $everythingelse variants

clobrano:
okay, let's see if it lives enough upstream

PR #1756, opened 2020-01-16
Folder-based File Manager icon

ubuntujaggers:
As per @Feichtmeier 's request.  This first design is identical to Suru, except for a different shaped baseplate.

I wonder if it looks a bit too faded, with not enough contrast - especially on the app grid (which is an inherited problem).

image

image


ubuntujaggers:
One option would be to darken the shadow on the folder a bit:

image

Feichtmeier:
I like this darker shadow!
What do the others say? @clobrano @madsrh ?

ubuntujaggers:
I've pushed the darker shadow :+1:

Feichtmeier:
Just tried the branch and imho this looks way better and feels more yaru/ubuntu/suru/origami
@clobrano are you neutral on this or -1?
image

clobrano:
checking...

clobrano:
Not sure, probably neutral.

@flexiondotorg what do you think? 

/cc @seb128 because last time the Nautilus change surprise the desktop team :smile:

seb128:
Looks good to me, I still didn't manage to get used to the current blue one and often visually 'browse' the launcher from top to bottom a few time to find nautilus but the new icon matches better a files/folders navigator in my opinion :-)

clobrano:
I was also thinking whether it would be good to increase the size of the folder inside the squircle

madsrh:
@ubuntujaggers would it be possible to add more contrast? I have a really bad screen and here it looks like this:

image

Feichtmeier:
Don't get me wrong but the icon still needs to be somehow compliant to the suru guidelines. If the contrast is really bad, we should rather think about a different color for the baseplate

ubuntujaggers:
How about this:

image

image

image

??

madsrh:
VERY nice @ubuntujaggers 🎉

Feichtmeier:
Def. very nice - holy smoly!
@clobrano ?

ubuntujaggers:
Pushed the dark grey and updated the symbol (y)  Ready to merge when everyone's happy.

clobrano:
Contrast is definitely better 😊
Is there a reason for having the folder so smaller than the baseplate?

Feichtmeier:
@clobrano 

https://snwh.org/suru/guidelines

grafik

I can't say it 100% but I think @ubuntujaggers just used the guidelines

ubuntujaggers:
> 
> 
> Contrast is definitely better 😊
> Is there a reason for having the folder so smaller than the baseplate?

I literally just took and recoloured the folder from the Suru icon and changed the baseplate around it:

image

madsrh:
What do you think @clobrano Should we merge this?

coxackie:
Sorry for bothering with a merged PR, but anyone else feels that the non-transparent background on this icon makes it feel out of place with respect to other icons? If placed on the dock, it feels like it is focused all the time, as other icons have back transparency... I like the folder design, but it the grey-brownish background necessary? One could make the actual folder icon larger and get rid of the whole background.

Feichtmeier:
The sizes come from the guidelines linked above. If you change it it looks inconsistent to the rest.

friendlyapple:
I agree with @coxackie that the background is too dark.
On the other hand, with this type of icon, I understand that the background is important to differentiate the icon from a simple folder.

clobrano:
I have to agree, only after trying it I think the previous background was better. Maybe something in between?

madsrh:
Should we try rolling back to commit 85221b1?

Feichtmeier:
No, stuart already used the big symbolic SVG for other stuff
Let's just reopen the issue and discuss it there and then make new PRs if needed

ubuntujaggers:
@coxackie

A larger folder without a background was the first thing I had for the non-uniform icon set - I liked it ☺️ but we got feedback that it just looked like a link to a folder rather than the file manager app.

Maybe we could revert to that with the Windows-style holder added to the bottom?

coxackie:
I understand that you want to follow the guidelines, and I respect that. If only all the other icons would follow them too, there would not be an inconsistent look. Right now, the folder icon does look small (in comparison to other non-suru icons). But I think what stands out most is the darker background - it makes the icon feel "focused" in some sense. Just my two cents here - you are the designers.

Feichtmeier:
Please use the reopened issue for further discussions

PR #1759, opened 2020-01-17
headerbar: increase headerbar backdrop contrast

### LINKED ISSUE #1740: Titlebar/headerbar of inactive windows
In Yaru theme (not light, not dark) the titlebars/headerbars of inactive windows are the same color (or almost) of active windows; It should be more like in Yaru-dark that is significantly different.

clobrano:
closes #1740

Before
image

After
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -14,7 +14,7 @@ headerbar,
   }
 
   &:backdrop {
-    border-color: mix(darken($inkstone, 11%), darken($inkstone, 1%), 80%);
+    border-color: $inkstone;
     background-color: $backdrop_headerbar_bg_color;
     box-shadow: inset 0 1px $headerbar_border_color;
   }
--- a/gtk/src/default/gtk-3.20/_colors.scss
+++ b/gtk/src/default/gtk-3.20/_colors.scss
@@ -85,7 +85,7 @@ $headerbar_fg_color: $porcelain;
 $headerbar_border_color: lighten($headerbar_bg_color, 8%);
 $headerbar_text_color: $silk;
 $headerbar_insensitive_color: mix($headerbar_fg_color, $headerbar_bg_color, 50%);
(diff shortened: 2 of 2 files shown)


madsrh:
This increases headerbar backdrop contrast for sure ✔️ 👍 but that means that the black button borders are even more pronounced now. Can we lighten up (or remove) the borders in backdrop or is that a separate PR/issue?

IMHO removing the border would be the "prettiest" solution but that might cause some XU problems IDK.

Before:
image

After:
image

clobrano:
I agree that the borders are too dark in backdrop, I thought of doing a different PR for that however

gtk/src/default/gtk-3.20/_headerbar.scss
```diff
@@ -14,7 +14,7 @@ headerbar,
   }
 
   &:backdrop {
-    border-color: mix(darken($inkstone, 11%), darken($inkstone, 1%), 80%);
+    border-color: $inkstone;
```
Feichtmeier: How about making use of the https://github.com/ubuntu/yaru/blob/684cf87538ac41ea5ac45da309bf0f130daad7b2/gtk/src/default/gtk-3.20/_colors.scss#L85 variable instead? So if you change the headerbar color the border adapts to it
For the backdrop you could add a backdrop variable

clobrano: The goal was to adapt the BG color only and I removed the overly complicated expression for border color with inkstone, which is basically the se, but I agree

clobrano:
> I agree that the borders are too dark in backdrop, I thought of doing a different PR for that however

alright, since I have to change some border colors anyway, I changed my mind.

@madsrh, what do you think about this lighter borders for buttons in backdrop?

right now
image

using hb color for borders
image

2% lighter hb color (it's barely visible for me)
image

madsrh:
Much improved, thank you - this was a torn in my eye 😄 🎉 So I'm happy with both.

clobrano:
very well, merging it then

PR #1764, opened 2020-01-19
tweaks: notebook tab hover text color

### LINKED ISSUE #1762: Notebook tab text is wrong on hover
Notebook tab text looks disabled on hover.

This is a bit odd because the Adwaita tab hover works fine.

Peek 2020-01-17 07-20

No video but in the light theme the text is also grayed.

**EDIT:** 

One fix would be to just keep the same $fg_color
border_dark_
border_light

OR...

add a subtle background highlight like the buttons has (I vote for this one 😉 )
darktabs_bg
lighttabs_bg

I was almost ready to make a PR with the code above, when I discovered that backdrop was broken 😭

clobrano:
On hover, notebook tab's text is grey and so it might be mistaken for
disabled.

This change sets for tab's text on hover the same color that unchecked
tab's text have.
Moreover it sets a slightly brighter background color, still less than
checked tabs, to provide more feedback to the user.

closes #1762

image
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -311,7 +311,7 @@ notebook {
           border: 1px solid transparent;
           border-radius: 4px 4px 0 0;
 
-          &:hover {
+          &:hover:not(:backdrop) {
             background-color: transparentize($base_color, 0.5);
             box-shadow: none;
             border-color: $borders_color;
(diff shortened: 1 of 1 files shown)


madsrh:
This looks great 👍 

We should probably make tabs insensitive/passiv in backdrop like the rest of the UI.

madsrh:
Lightning fast :zap: 
Can we merge this one?

clobrano:
green light for me🚦

PR #1765, opened 2020-01-19
tweaks: no borders for checkmarks in file menus and context menus

clobrano:
see #1703 

image
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -2981,7 +2981,7 @@ radio {
   }
 
   &:backdrop { transition: $backdrop_transition; }
-  
+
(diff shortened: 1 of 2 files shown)


madsrh:
LGTM

Feichtmeier:
Looks nice!
Do you think Mpt would think that the ring of checks also should be removed together with this?

madsrh:
> Do you think Mpt would think that the ring of checks also should be removed together with this?

I was considering the same

clobrano:
That's why I screenshotted the radios too :D

clobrano:
These are the radios without border

image

madsrh:
Yes, let's  do that. You can test this in Solitaire.

clobrano:
image

PR #1766, opened 2020-01-19
Aubergine-symlink-emblems

### LINKED ISSUE #1742: Symlink folder badge has the same color of the folder
As in below picture, the symlink badge color is too similar to the new folder color and it's hard to see the difference

image

Since the new folders are in an experimental phase, there is no need to work on this immediatly.

madsrh:
A simple recolouring of the emblem icon (no border added). I picked the aubergine from the folder icon, so it's slightly different than the official Canonical aubergine.

Closes #1742


clobrano:
Point 1, good work Mads, everything is working great! :)
Point 2, aubergine emblems work fine on file icons

image

but I wonder if they have enough contrast agains grey folders. Maybe the white border could improve it?

image

madsrh:
> but I wonder if they have enough contrast agains grey folders. Maybe the white border could improve it?

We can try that also. Would you like a different PR or just update this one? 
I kind of prefer that it's so subtle but still much improved over the gray, because there's so freaking many symlinks 😄IF we prefer the border, we should ofc update all the emblems.

Feichtmeier:
The white outlines looked a bit like something we never did before. I'm not sure they really fit into the theme

clobrano:
> > but I wonder if they have enough contrast agains grey folders. Maybe the white border could improve it?
> 
> We can try that also. Would you like a different PR or just update this one?
> I kind of prefer this where it's so subtle but still much improved over the gray, because there's so freaking many symlinks smileIF we prefer the border, we should ofc update all the emblems.

That's ok for me, we can just merge this one

PR #1770, opened 2020-01-21
Ubiquity icon based on Circle of Friends forming from pixels

ubuntujaggers:
image

Made this PR so someone who knows how to test Ubiquity, can do  :)

I think the other sizes need more work.


Feichtmeier:
Thanks!!!
I think it looks very good
 I'll test it tomorrow

madsrh:
Nice work ubuntujaggers 👍 but are you really reinstalling Ubuntu so often that you've put Ubiquity in the dock 😂😜

Feichtmeier:
@madsrh

When you want to install Ubuntu you either choose "install Ubuntu" (!!! In which case you don't see either the icon, a desktop or the dock at all !!!) and sooooome people choose "try Ubuntu"... For those people there is that shortcut on the dock which starts the installer again after they tried Ubuntu enough. When they click the icon the installer launches again. This time as an app on top of a regular gnome session so it gets placed on the dock like all apps you start :)

(If you know this and I just didn't get the joke sorry for this wall of text)

Feichtmeier:
Screenshot from 2020-01-22 09-51-01

I think it looks perfect :+1: Good job!

ubuntujaggers:
I've pushed an improved version at the different sizes:

image

ubuntujaggers:
Let me see if I can make it "cleaner"...

madsrh:
@Feichtmeier Are they (still) blue? BTW Yes, that looks great!

image

Feichtmeier:
Yes we need to fix this

ubuntujaggers:
I don't think I can make it cleaner without rethinking, it's an effect of the "backplate" having tiny pieces.

If it's only seen on the launcher and wallpaper, it doesn't matter.  It will only affect people who change the wallpaper in the live trial.

But if this icon shows up in a dialogue box or something it won't be great.

Feichtmeier:
> I don't think I can make it cleaner without rethinking, it's an effect of the "backplate" having tiny pieces.
> 
> If it's only seen on the launcher and wallpaper, it doesn't matter. It will only affect people who change the wallpaper in the live trial.
> 
> But if this icon shows up in a dialogue box or something it won't be great.

I don't think anyone will change the wallpaper in the live session to be white :)
Imho this is looking very nice.

ubuntujaggers:
So it should look like this on the live disk desktop (mockup):

image

Pretty smart I think :+1:

PR #1782, opened 2020-01-23
tweaks: use lighter/darker border for checked check- and radio-boxes

clobrano:
Use darker border in light variant and lighter border in dark variant to
increase the contrast with the background

Right=current, Left=new
image

image

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -675,4 +675,21 @@ radio {
       }
     }
   }
+
+  & {
+    // for checked
+    @each $t in (':checked'), (':indeterminate') {
+      &#{$t} {
+        @each $state, $t in ("", "normal"),
+                            (":hover", "hover"),
+                            (":active", "active") {
+          &#{$state} {
+            border-color: if($variant=='light', darken($checkradio_bg_color, 10%), lighten($checkradio_bg_color, 5%));
+          }
+        }
+      }
+    }
+  }
 }
+
+


madsrh:
I didn't test yet, but from the screenshot it looks good 👍 

For the unchecked boxes (that you didn't touch here) I really dislike the dark borders in the dark theme and would love to return to the light borders (like we had in Yaru 1.0 🤔), but that's a different issue 😄

clobrano:
It makes sense, I'll propose a change here as well

Feichtmeier:
Hm, this looks very out of place

We don't use light borders anywhere in the theme yet

I understand why you want it, but I think it does not fit

clobrano:
Really out of place? :eyes:

madsrh:
> out of place?

I think he means using a light border for the checked boxes but a dark border for the unchecked.

clobrano:
That's to increase the contrast with the background only in the border, which is the immediate point of contact and keeping the current aubergine tones, which I think are good. As said in chat, the light variant might not need it, but looking at the dark variant the checkboxes/radio on the right looks bigger than the others just because the border is a little more visible

Feichtmeier:
Sounds a bit harsh. Let's say it this way: stylistically unexpected :)

I would rather try to lighten up the ```$checkradio_bg_color``` in ```_colors.scss``` for ```$variant=='dark'``` and keep the colour for the light theme

Same for the unchecked border
Currently they look like the other buttons
With a white border they look... Different

Maybe we could rather talk about the overall contrast of the dark theme with the current borders instead of changing specific widgets.
It would be nice if changing the variables in colours would be enough to spin all the wheels in common automatically. This would reduce the complexity a lot. Otherwise we need to check all the different places in _tweaks and see what we did there
(((Imho (my) change on the switchea and the change of the tabs in tweaks was already too much)))

Feichtmeier:
> Really out of place? 
> 
> EDIT:
> @Feichtmeier which emoji is the right one? 
> image

Both :) I agree that the contrast is not enough for the dark theme, but I don't favour the solution and I don't agree for the light theme:)

clobrano:
This is the comparison with:
Left: current but with bg 4% brighter
Right: this PR with borders only 4% brighter than the BG

image

My point of view is that, while left is probably okay, the borders are a bit lost into the dark background, while the right one keeps the same tone overall, but is much crisper.

The new border would be just slightly brighter, it's not something evident, but still make a good impression for me.

Feichtmeier:
Could you try to use the same box Shadow as for the normal buttons in drawing instead of a bright border? (Just an idea)

clobrano:
For the reference, this is the result (left: master,  right: master + box-shadow)
image

I think there isn't much difference, because the box-shadow color changes from dark 95% trasparent to 90%
I could reduce the transparency, but the effect would be to make the checkbox/radio popup from the background. Below it's an exaggeration with transparency al 70% only, just to give the idea
image

Also, the effect is only at the bottom of the checkbox/radio, not all around and it is applied to both light and dark and both checked and unchecked, which was not the goal of this PR

Feichtmeier:
Yeah, looks weird.

Could you try with $switch_border_color ?
Sorry for all that"ideas". I'm just not a fan of that bright border as it makes the asset look as it lies in a collapsed area and the highlight to be higher

Feichtmeier:
But this one is already looking good imho
image

clobrano:
It's good, but It doesn't address the problem at all for me

Feichtmeier:
Just for reference, this would be with switch_border_color
image

It works, but it makes them look smaller :S So... I guess it doesnt work

PR #1788, opened 2020-01-23
focus ring: review focus ring for some widgets

clobrano:
The focus ring (outline) of the following widgets have been reviewd
taking care of their peculiarities

- button
    it required a change in drawing, where the outline-color was
    somehow hardcoded
- switch
- check/radio button
- row (selected/unselected)
    it required - in selected state - a different outline. When
    selected, the row's background is a solid orange, where the
    standard aubergine outline is not visible enough. For this
    reason bg_color has been choosen. Moreover, since bg_color
    is the same as the row's backgroud window, a dashed line has
    been preferred, so that it is clear that it is an outline and
    not a border or a smaller row.

image
image
image
image
image
image
image
image

### DIFF
--- a/gtk/src/default/gtk-3.20/_colors.scss
+++ b/gtk/src/default/gtk-3.20/_colors.scss
@@ -78,7 +78,7 @@ $checkradio_bg_color: if($variant=='light', $aubergine, darken($aubergine, 8%));
 $checkradio_fg_color: #ffffff;
 $switch_bg_color: if($variant=='light', $aubergine, darken($aubergine, 8%));
 $switch_border_color: if($variant=='light', darken($aubergine, 15%), darken($borders_color, 5%));
-$focus_border_color: lighten($aubergine, 14%);
+$focus_border_color: transparentize(lighten($aubergine, 14%), 0.3);
 // Headerbar bg colors for the "mixed" theme
 $headerbar_bg_color: #323030;
 $headerbar_fg_color: $porcelain;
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Perfect :+1:

Feichtmeier:
@clobrano as discussed an idea with no diff in drawing

```css
// Focus "ring" - upstream is also interested in this, so better get this into upstream at some time
*, * button, headerbar button {
  outline-color: transparentize($focus_border_color, 0.7);
  outline-style: solid;
  outline-offset: -1px;
  outline-width: 2px;
  -gtk-outline-radius: $button-radius;
}

switch slider {
  outline-color: $focus_border_color;
  outline-style: solid;
  outline-offset: -1px;
  outline-width: 2px;
  -gtk-outline-radius: 100%;
}
row:selected { outline-color: transparentize($selected_fg_color, 0.5); outline-offset: -4px; }
```

clobrano:
Does this work in your environment? In mine doesn't seem to have effect.

image

Feichtmeier:
Sorry you are right, for the mixed theme headerbar it doesn't work
I guess because it loads headerbar.scss after tweaks and thus the button outline is overridden again? :thinking:

clobrano:
It might be, I'll invert include position in `gtk.css` for a while and check whether it has any side effect

Feichtmeier:
Okay never mind - sorry for the delay was a bit dizzy IRL 

The few lines in drawing are the cleanest for sure. Otherwise we need to style all button style classes explicitly again which is not really cleaner (like we did for the inverted headerbar :/ )

PR #1789, opened 2020-01-23
TESTING PR: More contrast for context menus

madsrh:
More contrast for context menus and popovers 🙈 

This might be a bit too much but IMO the backgrounds in master still blend a bit together 🤷‍♂ 

Left is master and right is `lighten(bg_color, 2%)`

image

image

image

image


Feichtmeier:
LGTM, changes in colours are mostly the best and safest changes :)
And it still has contrast on both bgcolor and basecolor
(Just woke up because my daughter ran screaming into our bedroom... Nightmare)

Feichtmeier:
Ah wait... About the dark popovers. I'm not sure. Do you see that the buttons are now the same colour again? :)

madsrh:
I see your point. Also I've noticed that the checked radiobuttons, especially on hover, blend.

I added popovers to align the two (menu and popover), but we could just change $menu_color only

Feichtmeier:
Yeah better keep popovers as they are, they are also matched in the shell

madsrh:
Like this @Feichtmeier ?

PR #1796, opened 2020-01-25
Small code change

madsrh:
I thought there was a simple way of doing this:

```
$_alpha: if($variant=='light', 0.08, 0.16);
$_separator_color: rgba(0, 0, 0, $_alpha);
background-image: image($_separator_color);
```

It turns out my solution looks just as complicated, but at least it uses the `$border_color ` variable.

`background-image: image(if($variant=='light', transparentize($borders_color, 0.6), darken($borders_color, 1%)));`


madsrh:
This might be even better:

`      border-color: if($variant=='light', lighten($borders_color, 8.3%), darken($borders_color, 1%));`

Issue #1791, opened 2020-01-24
Improvements to the floating bar


friendlyapple:
I think the floating is too visible.

I suggest that we change the background and the borders and reduce the font size.

**Current floating bar:**
![1](https://user-images.githubusercontent.com/59767069/73070022-951b1380-3eaf-11ea-91af-9928a845eb81.png)

**Suggested floating bar:**
The difference is more visible when using nautilus.

![2](https://user-images.githubusercontent.com/59767069/73070046-a2d09900-3eaf-11ea-9e47-6d0b000f2093.png)

**Used colors:**
![3](https://user-images.githubusercontent.com/59767069/73070051-a6fcb680-3eaf-11ea-95c2-579567e70984.png)

**SCSS:**
yaru/gtk/src/default/gtk-3.20/_apps.scss

```css
paned box.floating-bar {
      font-size: 95%; /* or equivalent em unit or 14px */
      background: #fbfbfb; /* TODO: replace with an existing variable */
      border-style: solid;
      border-color: #d9d9d9; /* TODO: replace with an existing variable */
      border-width: 1px 0 0 1px;
      border-radius: $button_radius 0 0 0;
}
```

The change is slight but it makes the interface cleaner.

Feichtmeier:
Nice idea, and and safe spot to change :) Are you interested in a PR?

friendlyapple:
I think it will be faster if you did it :) (if it doesn't bother you)
Especially that it is necessary to find the variables for the color of the border and the background (I don't know their name).

meetdilip:
No accessibility issues ? Someone with glasses will have a better chance of reading it when it is properly highlighted.

friendlyapple:
Accessibility is still OK. The Floating is always clearly visible and legible (the font is only reduced by 5%).

clobrano:
> No accessibility issues ? Someone with glasses will have a better chance of reading it when it is properly highlighted.

If background and font colors are the same as in the sidebar there won't be any accessibility problem (or all the sidebar will have it). However, reducing font size by a small % might be actually a problem in small screens (12'' for example)

friendlyapple:
On a small screen I don't know. But I think that 5% less on the font size should not affect its readability. Moreover, since it's an indication/help, I find it normal that the font size is smaller than the label of the icons.

madsrh:
Putting screenshot (with comments) and code in the issue, is just so cool @friendlyapple 🥇 

[...]

PR #1798, opened 2020-01-25
Improvements to the floating bar #1791

madsrh:
Fix (without the font-size change) for https://github.com/ubuntu/yaru/issues/1791

image


madsrh:
@Feichtmeier I just noticed that I've used `background-color` and not `background `but I'm not sure if it makes any difference here.

clobrano:
> @Feichtmeier I just noticed that I've used `background-color` and not `background `but I'm not sure if it makes any difference here.

`background` and `background-color` have the same effect

PR #1804, opened 2020-01-26
Add icon for mimetype ROM files (consoles)

madsrh:
No fancy icon design here. I simply took the controller icon 🎮  from `source-symbolic.svg` to use for the following roms:

**Nintendo:**
 NES ROM (application/x-nes-rom)
 Nintendo DS ROM (application/x-nintendo-ds-rom)
 Nintendo64 ROM (application/x-n64-rom)
 Game Boy ROM (application/x-gameboy-rom)
 Game Boy Advance ROM (application/x-gba-rom)

**Sega:**
 Sega Master System/Game Gear ROM (application/x-sms-rom)
 Genesis ROM (application/x-genesis-rom)

I've added the symlinks in the `mimetype.list` but I couldn't make the script run without fail 😩  This means that the symlinks aren't applied.

image


Feichtmeier:
Okay .gb works :) But your GEDIT trick didn't. So I got one from a super legal source...
image

.nds files don't work (guess because of the symlinks?)

Let me check to make the symlink
Yup worked with carlos fancy script change
image
navigate to
yaru/icons/src/symlinks
and run
./generate-symlinks.sh -m application-x-gameboy-rom

madsrh:
👍 Nice

> `./generate-symlinks.sh -m application-x-gameboy-rom`

What's the `-m` ?

Feichtmeier:
I think "match" :)

Feichtmeier:
Works!
image

PR #1811, opened 2020-01-26
New notebook tab's style fixes

### LINKED ISSUE #1802: Gtk3: notebook issues
1) When a tab is selected, there shouldn't be a change on hover. Here a bottom border appears.
Peek 2020-01-25 20-05

2) In the backdrop they have the wrong bg color
image
I guess they need the backdrop_base_color

3) Padding changes when clicked
Peek 2020-01-26 15-34

Edited by @Feichtmeier

clobrano:
- tweaks: notebook tab checked backdrop use same color as .view
- tweaks: notebook tab checked in backdrop no bottom border
- tweaks: notebook's tab unchecked font weight
- tweaks: fix notebook's tab checked-hover state

closes #1802

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -26,7 +26,7 @@ $ambiance: null !default;
     @at-root %dim_icons_in_backdrop,
     & *scrolledwindow:backdrop { opacity: 0.9; }
 
-    paned box.floating-bar {  
+    paned box.floating-bar {
       background-color: mix($bg_color, $base_color, 50%);
       border-style: solid;
       border-color: if($variant=='light', lighten($borders_color, 8.3%), darken($borders_color, 1%));
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -544,7 +544,7 @@ notebook {
       outline-offset: -5px;
 
       color: $fg_color;
-      font-weight: normal;
+      font-weight: 500;
 
       border-width: 1px;         // for reorderable tabs
       border-color: transparent; //
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -369,7 +369,9 @@ notebook {
             border-color: $borders_color;
           }
 
-          &:backdrop { box-shadow: none; }
+          &:backdrop {
+            box-shadow: none;
+          }
 
           &:checked {
             background-color: $base_color;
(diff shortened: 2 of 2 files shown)


Feichtmeier:
Backdrop fixed :+1: 
Hover border gone :+1: 
Font-weight change sadly still visible :crying_cat_face: 
Peek 2020-01-26 22-46

Feichtmeier:
nice font jump is fixed :+1: 

Super nitpicking alert: do you think we should keep this backdrop color? 
Peek 2020-01-27 00-13

friendlyapple:
When a tab is inactive, maybe we should make the text clearer? Currently, it looks like all tabs are active.

Feichtmeier:
I would rather use a lighter colour for the inactive tabs, fg_color is fine for the tabs

clobrano:
> When a tab is inactive, maybe we should make the text clearer? Currently, it looks like all tabs are active.

At the design fest in London has been pointed out that having all the inactive tabs text a bit transparent made it look disabled, this is the reason of the change. I tried to keep a different in font-weight, but the effect of size-change when selecting them is annoying.

friendlyapple:
If we can't change the thickness of the font and we can't gray out an inactive tab, there is an ergonomics problem. We must make a choice.

I like the style of Yaru in Ubuntu 19.04. We quickly identify which tab is selected (I think).
68747470733a2f2f692e696d6775722e636f6d2f49617538575a6f2e706e67

With the gray color, we don't have the impression that the tab is disabled. Moreover, Adwaita also uses gray for inactive tabs.

Feichtmeier:
I think $insensitive_fg_color was perfectly fine - but I think this was where it all started at the design fest - it was said that the inactive tabs looked... inactive. Well yes! That's what they are :) Anyways, I think in the current version in this branch one can still identify which tab is active

clobrano:
In the very latest commit, unchecked tabs text is actually a little (10%) more transparent

friendlyapple:
With 10% transparency, the difference is not noticeable :/

test

Moreover, why put everything in bold? Wouldn't normal thickness or 500 be enough?

clobrano:
That's actually a good idea. I kept the bold, but there's no reason anymore

Feichtmeier:
@clobrano I meant the background color :)

Active: $bg_color
Backdrop: $backdrop_dark_fill -> I guess you should use $backdrop_bg_color here :)

https://github.com/ubuntu/yaru/blob/116eb714ef5b87a88e8396f6e0f46154a3dd81b8/gtk/src/default/gtk-3.20/_tweaks.scss#L352

Nitpicking alert: you can just say $bg_color here: https://github.com/ubuntu/yaru/blob/116eb714ef5b87a88e8396f6e0f46154a3dd81b8/gtk/src/default/gtk-3.20/_tweaks.scss#L348

Feichtmeier:
image

Works :tada:

PR #1812, opened 2020-01-26
tweaks: use $bg_color for sidebar

### LINKED ISSUE #1803: Nautilus sidebar color
PR #1798 made me realize that Nautilus siderbar and Gnome-settings (and gedit, and tweaks) background colors are not aligned as they used to be.

Mads correctly used `background-color: mix($bg_color, $base_color, 50%);` for the floating bar, because it's the same color the sidebar has, but it used to be `$bg_color` (and they're almost the same color (#FBFBFB vs #F7F7F7 which is porcelain) as the main window background color of the mentioned main Gnome applications. The difference is below

Nautilus in master
image

Nautilus with sidebar using `$bg_color`
image

and here the matching settings
image

also gedit has the same colors
image

It is not a big problem, but at least explain to me why the sidebar seemed different in the last months.
Personally, nautilus sidebar color is not different enough to create a separation from the main windows. The two colors are so similar that one might ask why not making them the same.

My motivation to revert sidebar to `$bg_color` is to avoid having yet another color in the palette and to create some matching with the other main applications with a sidebar (gnome-settings, gedit and tweaks).

clobrano:
Nautilus sidebar color is not different enough to create a separation
from the main window. The two colors are so similar that one might ask
why not making them the same.

The motivation to revert sidebar to $bg_color is to avoid having yet
another color in the palette and to create some matching with the other
main applications with a sidebar (gnome-settings, gedit and tweaks).

closes #1803


gtk/src/default/gtk-3.20/_common.scss
```diff
@@ -4016,7 +4016,7 @@ filechooserbutton:drop(active) {
  ***********/
 .sidebar {
   border-style: none;
-  background-color: mix($bg_color, $base_color, 50%);
+  background-color: $bg_color;
```
Feichtmeier: Can't we put this in _tweaks?

clobrano: I would have bet on it. Okay

Feichtmeier: Well, either we stick to the plan or we make a new one ¯\_(ツ)_/¯
Currently it's still possible to have a simply "Yaruwaita" with just using our colors. If we start fiddling with common ... no one knows where it ends if one starts it :) Don't take it personally

clobrano: I don't, no worries :)

Feichtmeier:
Looks good! 
Only thing I've found that could be considered as an issue is that the .search entry for tweaks and settings has a different bg color now
image

clobrano:
I double checked, this comes from Adwaita

image

clobrano:
Also, I run the nautilus css sync script and there are no updates

PR #1820, opened 2020-01-27
Hover for treeview

madsrh:
Add hover for treeview similar to what we have in the sidebar and context-menus.

**Current:** (only the checkboxes change on hover)
yeau

**Suggestion:**
ex


clobrano:
Yes, very well done :+1:

clobrano:
Also, thanks a lot for removing those white spaces :)

PR #1823, opened 2020-01-28
common: fix backdrop link color

clobrano:
Currently, backdrop link color is orange, instead that some blue.
This has been alredy fixed upstream

https://gitlab.gnome.org/GNOME/gtk/merge_requests/1201

but it does not seem available yet, so this change manually cherry-picks it in Yaru.

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -1182,28 +1182,29 @@ list row button.image-button:not(.flat) {
   &:visited {
     color: $link_visited_color;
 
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Did you also port it back to the gtk-3-24 branch?

clobrano:
No, I didn't, that might probably be the reason

Feichtmeier:
Okay, put it in tweaks until it's in upstream please.

Feichtmeier:
https://github.com/ubuntu/yaru/wiki/Yaru-theme-suite-workflow-and-guidelines

Feichtmeier:
We can do it the other way round:

when it lands in upstream, there is no diff anymore :)

Issue #1423, opened 2019-07-08
Cursor misalignment


jhakonen:
The normal cursor's point isn't at mouse's coordinates. It seems to be a few pixels too high and a bit to the left. See attached screenshot:
![misaligned-normal-cursor](https://user-images.githubusercontent.com/1950698/60827985-09af8e80-a1ba-11e9-9e90-de0f573f4714.png)

In the picture I have changed settings from Universal Access --> Zoom:
* Set zoom to 2.00
* Enabled cross hair and made it as thin as possible 

It seems that other cursor types have similar issues as well.

Using Ubuntu 19.04 with the Yaru theme that came with it.

In the mean time I've switched to Hackneyed cursor theme (https://gitlab.com/Enthymeme/hackneyed-x11-cursors) as it seems to be better aligned. Here's for a comparison what the position probably should look like:
![aligned-hackneyed-cursor](https://user-images.githubusercontent.com/1950698/60828753-e71e7500-a1bb-11e9-94fc-d09cb29fa9ad.png)

```
$ snap info communitheme
name:      communitheme
summary:   The next Ubuntu theme built by the community.
publisher: Didier Roche (didrocks)
license:   GPL-3.0
description: |
  Yaru, formerly known as Communitheme, is the new Ubuntu theme built by the
  community. Yaru will become the default Ubuntu theme in Ubuntu 18.10. This
  package allows you to try out the theme on Ubuntu 18.04 LTS.
  
  To try out the theme, install this package on Ubuntu 18.04 LTS, restart
  your computer and select the "Ubuntu with communitheme snap" session from
  the login screen.
  
  More information is available at
  https://community.ubuntu.com/t/faq-ubuntu-new-theme/1930.
snap-id: Yd6CISPIf6tEf3ZEJ0cqSoEg9rG2VkRi
channels:
  stable:    0.1 2019-03-13 (1768) 16MB -
  candidate: ↑                          
  beta:      ↑                          
  edge:      0.1 2019-07-06 (1823) 17MB -
```

fitojb:
Indeed, if one puts the mouse pointer all the way to the screen’s left edge, one can notice the missing border.

Amr-Ibra:
I wish the cursor could be sharpened more.

rustushki:
[...]

PR #1825, opened 2020-01-29
Changing hand cursors to white hands

ubuntujaggers:
image

I can't screenshot this because cursors aren't included in screenshots :)  but I've got it working, _except_ for when I drag a window by the top bar.  I can't work out why though.

I'm making a draft PR so we can test it, to see if:

- We're happy with the design, and
- Others experience the same black hand when dragging a window (it might be a caching problem at my end that would eventually resolve itself)

Dunno if anyone can see a rogue black hand that I've left in the build, but I've got rid of all the ones I can see myself.


madsrh:
Now that looks like a hand :tada: 

> Others experience the same black hand when dragging a window

I see it too when dragging a window. I think `all-scroll` and `move` are somehow link, but I had the same problem when looking in to this (and I couldn't even draw a hand). I'll investigate again when I have some time.

Did you try the `test.html` I included? :smiley:

ubuntujaggers:
> Did you try the `test.html` I included?

I'm trying it now  :)  but weirdly it shows the black hands, even for ones that work as white hands elsewhere  :\  I will keep investigating!

ubuntujaggers:
I've worked out the problem - there are some cursor files that aren't created by the script, which already exist in the relevant folder.  The cursor files with no extension or .cur.  So when you run the script it doesn't replace them with anything and they stay as black hands.

I'll see if I can update the scripts, or will make them manually :+1:

clobrano:
If you have problem updating the script, I can have a look into it, just create a bug as a reminder :wink:

ubuntujaggers:
Okay, try now, I think I've fixed it  :)

I'm still getting the old hands in the browser test, but at least the hands are white in the desktop :man_shrugging: and maybe it's just a case of snaps not picking up the latest theme?  Certainly dragging windows now works and dragging icons works.

Feichtmeier:
As always, awesome work!

However I'm not sure if this actually fixes mpts initial problem with the ✋ not being recognised as such. Isn't it the same icon but with inverted colours now? :)

Also: do we really want to be a mixed cursor theme? (Mixed in terms of some cursors are white with black outline and others are black with white outline)

madsrh:
> Isn't it the same icon but with inverted colours now? :)

It's a bit different. It fixes the issue because when you look at the icon, it easy to see that it's a hand. That's not the case with the current.
image


> do we really want to be a mixed cursor theme? (Mixed in terms of some cursors are white with black outline and others are black with white outline)

Looking at [another OS that uses a black pointer](https://daviddarnes.github.io/mac-cursors/), I think this works fine. The pointer and the hand are two different things IMO - like the spinner (we don't want a black spinner with a white outline).

Feichtmeier:
Oh, your are right. Sorry, it's indeed different.
Alright then +1

ubuntujaggers:
I'll change the hotspots and push tonight to fix the other issue 👍

madsrh:
> I'll change the hotspots and push tonight to fix the other issue

@ubuntujaggers I don't know if it was this issue you were referring to but perhaps you could look at this one at the same time?

https://github.com/ubuntu/yaru/issues/1423

ubuntujaggers:
That _is_ the issue I mean, but this is going to take some diagnosis, because moving the hotspots (the "click point") and rebuilding the cursors hasn't changed anything  :(  so I'll keep investigating.

madsrh:
> That is the issue I mean, but this is going to take some diagnosis, because moving the hotspots (the "click point") and rebuilding the cursors hasn't changed anything :( so I'll keep investigating.

I never tried this, but what about leaving the hotspot and moving the icon? (sorry if that's stupid)

ubuntujaggers:
@madsrh idea worked, so I pushed an improvement to the main arrow pointers, give it a try now:

image

madsrh:
Could you check if this is working for you @jhakonen ?

 Excellent work @ubuntujaggers  🎉

jhakonen:
The arrow pointer is now much better indeed. :+1: 
But, there's still issues with at least the border dragging cursors.

Here's the cursor on window's left border, just a pixel shy before it becomes a normal cursor:
cursor-horizontal-left

And here's the right border:
cursor-horizontal-right

I think the cursor's hotspot should be one pixel more to the right and one down?

ubuntujaggers:
Oh yes, lol - there's potentially lots more arrows to do.  I'll put it on the to-do list for ASAP :+1:   There's also the arrow with swirly ball that's likely not in the preferred place.

Can you do me a favour, if you get chance, and mark on your screenshot where you think the hotspot should actually be for those two?

ubuntujaggers:
Also we can't move the graphic around the hotspot for the arrow-with-swirl icon because the graphic uses all the space in the grab area.  We have to work out how to move the hotspot relative to the canvas.  But I'm happy to do that as another PR when we know how.

madsrh:
Sounds like a good solution. So let's merge this one now?

ubuntujaggers:
> Sounds like a good solution. So let's merge this one now?

I can still do the < > arrows but I don't quite get where we want the hotspot  :)

jhakonen:
Well, if you look at the < > cursor, the arrow that is over the window is in the 2nd picture wholly within the window, while in the 1st picture, the arrow is 2(?) pixels over window's edge. If both cursors would be moved a pixel to right, they would go over the window's edge by same amount.

In addition the arrows' points are not vertically on top of the crosshair.

Here's Pintaed pictures with the cursors moved a bit:
left-border-2
right-border-2

Were these any help?

ubuntujaggers:
Okay, not sure yet how to fix this...

- I can't move these ones to put the hotspot over the points because then part of the graphic will be outside the canvas...  I have to work out how to move the hotspot, rather than keeping it in place and moving the graphic around.

- The source files for these two look exactly symmetrical/central so not sure why it would behave differently on the two sides of the window.

If people are keen on getting the white hands in (@madsh :) ) then I say merge this and we can open subsequent PRs for improvements to hotspots, as and when we know how to do them?

madsrh:
Another way to test is with the `gtk-3-demo` and sadly here there is sadly still two missing icons (wait and grab). I'm not sure if this is just a missing link.

image

fitojb:
White isn’t the only color of human hands, did you know? The fact that macOS uses a white hand cursor and black cursors for the rest of shapes isn’t at all a reason to introduce this (thinly-veiled racist) inconsistency to Ubuntu’s cursors.

Feichtmeier:
Are you serious?

madsrh:
Please let's not make this into a racist discussion! 

As mentioned, I don't think we have inconsistency here. To be **100% consistent** we would also need invert the zoom icon to be black with a white outline and the busy spinner should have a black background - we don't want that IMHO.

I see the grab hand a bit like the hands of a magician. But sure, that could be Kenrick McDonald 🤷‍♂
image

fitojb:
> racist discussion

*racism discussion. Fixed it for you.

> To be 100% consistent we would also need invert the zoom icon to be black with a white outline and the busy spinner should have a black background - we don't want that IMHO.

I’ve never questioned “100% consistency” nor I challenged those? macOS also has a colored beachball, and it’s not a problem.

> I see the grab hand a bit like the hands of a magician

Nice rationalization, but still troubling since those were historically used as cover to avoid... Bah, why am I wasting my time trying to teach Caucasians issues of people of color—you will obviously never appreciate it. I will just say Microsoft’s cursors have never conveyed gloved hands but are neutral, and BTW Kenrick McDonald doesn’t use white gloves.

The whole thread doesn’t explain why *hands* have to differ from the other pointer shapes in color. To fix the issue of an unrecognizable hand shape, the cursor color didn’t have to change from black to white. That change was gratuitous.

madsrh:
You are creating a problem that doesn't exist! This fixes the issue because the old cursor didn't look like a hand, this does.

> The whole thread doesn’t explain why hands have to differ from the other pointer shapes in color.

It looks better? 🤷‍♂You might want to join the _GDM / search entry_ discussion too, because I also prefer white version there, for no reason at all.

fitojb:
> You are creating a problem that doesn't exist!

I acknowledge that it doesn’t exist **for you**.

image

> It looks better?

Sure, but then make an all-white cursor theme. Again, there is *no reason* for *this one* cursor shape to be *inconsistent* in color. It doesn’t fit.

clobrano:
This discussion seems based on a false assumption that we went for a white hand because of racism, which is - frankly obviously - false

PR #1828, opened 2020-01-29
Improving contrast on Transmission icon

ubuntujaggers:
Transmission really annoys me every time I see it...  this is a very minor change but I think it makes it better, merge if you agree :+1: 

Before:

image

After:

image

image

image

image


madsrh:
Could you elaborate @friendlyapple ? 😃

friendlyapple:
@madsrh Of course! 😄 I prefer the old icon because I find the borders more consistent with the style of the icon.

PR #1830, opened 2020-01-30
Remove orange icon overlay

madsrh:
Hopefully this closes https://github.com/ubuntu/yaru/issues/43

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -27,9 +27,9 @@ $ambiance: null !default;
     & *scrolledwindow:backdrop { opacity: 0.9; }
 
     paned box.floating-bar {
-      background: $bg_color;
+      background-color: $bg_color;
       border-style: solid;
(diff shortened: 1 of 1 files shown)


friendlyapple:
Good work! I don't know if you can correct this at the same time: I noticed that these icons may have a focus ring. Would it be possible to delete it ? Example (look around the orange area):

frame_027_delay-0 1s

clobrano:
Great Mads, it works flawlessly!
However, you probably make this  change on top of your previous branch, instead of starting from master and it gives you a conflict.

The new change is only the one under
```
scrolledwindow widget.view:selected {
```
?

Feichtmeier:
1) Carlo is correct the branch has conflicts :)

2) @friendlyapple indeed there is something weird with the edges. Outline colour could be set to transparent. I think we load _tweaks after _apps so the dashed line we once assigned for these icons may be overwritten. Could be wrong. Need to check

madsrh:
I'll make a clean PR 👼

PR #1831, opened 2020-01-31
Remove orange on icons

madsrh:
**THIS** PR should be better and fix the orange spill that @friendlyapple mentioned.

image

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -26,7 +26,7 @@ $ambiance: null !default;
     @at-root %dim_icons_in_backdrop,
     & *scrolledwindow:backdrop { opacity: 0.9; }
 
-    paned box.floating-bar {  
+    paned box.floating-bar {
       background-color: $bg_color;
       border-style: solid;
       border-color: if($variant=='light', lighten($borders_color, 8.3%), darken($borders_color, 1%));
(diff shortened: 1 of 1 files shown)


Feichtmeier:
@madsrh I am still not sure - at least on the picture it still looks like the edges are cut off except the corners :thinking:

madsrh:
Sorry, I'm stupid. The picture is the one posted by @friendlyapple (https://github.com/ubuntu/yaru/pull/1830#issuecomment-580624553) to illustrate what the problem was - very misleading 🤦‍♂️ 

Is there anything else you are concerned about?

Feichtmeier:
Holy smiffelz you fixed it really
image
image

BUT :( !

Currently you assigned it outside of 
.nautilus-window { }

This means it is assigned for all widgets in scrolled windows, not only nautilus :(

((( And nitpicking: don't use $orange and $ash there. Try to use colors from _colors, like $selected_bg_color )))

madsrh:
Excellent advice BUT I can't make it work inside` nautilus-window ` 😭 Would you mind just pushing the correct changes? 👼 

>  And nitpicking: don't use $orange and $ash there. Try to use colors from _colors, like $selected_bg_color

`$orange` should obviously have been `$selected_bg_color` 👍 but I don't think there's a way to replace the `$ash`.

Feichtmeier:
Hm didn't work for me :(. I'll check tomorrow if you didn't find a solution until then :D

clobrano:
It's weird, under `.nautilus-window` it seems that only the label color can be changed, while icon background is forced to `@theme_selected_color` when the windows has focus and to some grey when in backdrop

Feichtmeier:
@madsrh 

Just tried it again myself. This seems to work:

.nautilus-window widget.view { background-color: rgba(255, 255, 255, 0.1); }

Peek 2020-02-11 21-59

madsrh:
:man_facepalming: Moved to a new PR  #1900 #SorryGitNewbieHere

Issue #1743, opened 2020-01-15
Selected disabled checkmarks in listview


madsrh:
For the selected disabled checkmarks and text in listview the checkmark shouldn't be orange and the text is too subtle.

Current:

![image](https://user-images.githubusercontent.com/3986894/72424742-c2b4de00-3786-11ea-8827-7b2387adf2c1.png)

I don't really have a good idea for this one 😞 Maybe something like this?

![image](https://user-images.githubusercontent.com/3986894/72424953-1b847680-3787-11ea-8cce-a8ec09a63d75.png)

Feichtmeier:
Thanks :+1: 

Fix should be placed in _drawing.scss

PR #1833, opened 2020-01-31
Color fix for disabled checkmarks in treeview

madsrh:
Made a PR with the oneliner from @Feichtmeier in issue #1743 

image


Feichtmeier:
Ah, No sorry. This needs to be put into tweaks. It's in my to do list, promise :)
You also inserted lots of diff with the spaces but np I'll change it promise promise!

PR #1843, opened 2020-02-02
Adding full colour app icon for bluetooth

### LINKED ISSUE #1841: Icons: blueetooth icon has the old suru squircle baseplate
image

ubuntujaggers:
After further investigation, looks like there were icons for apps --> bluetooth in the theme, but no corresponding svg in src, which is why we missed this (only the full colour category is in src).  Fixed that in this PR.

image

Closes #1841


Feichtmeier:
Damn you're fast! Thank you ! :clinking_glasses:

madsrh:
Where is this used?

Perhaps it would make sense to use this shape rather than a perfect square?

image

ubuntujaggers:
It would certainly match the existing symbol better - give me a sec :+1:

ubuntujaggers:
Also I don't think it's used anywhere in more, in recent version the symbol is shown  :(  but I just pushed an oblong version to match the symbol.

image

Feichtmeier:
I think it is only used when no other icon is available
here 
image
((( this was some crap tablet with android 4.1 on it )))

Keyboard:
image

Mouse
image

PR #1683, opened 2019-12-30
Miscellaneous mimetype icons

madsrh:
PLEASE check thoroughly! ⚠️ 

I need help figuring out the symlinks. This PR contains only the .svg files. Is the font size too big? (compared to the audio files)

Closes https://github.com/ubuntu/yaru/issues/1163 
Closes https://github.com/ubuntu/yaru/issues/1374

The firmware icon very rough, but perhaps @ubuntujaggers can give it a makeover.

image

PR #1853, opened 2020-02-03
Added icon for mimetype inode-symlink

madsrh:
This icon is for broken symlinks. To test just move a symlink to a different location.

One of the icons that got ditched in https://github.com/ubuntu/yaru/pull/1683

image


clobrano:
Is this what you expect? It looks a bit crowded

image

madsrh:
>  It looks a bit crowded

Not really the icons fault. Do you have an idea for something that would work better when it is covered in emblems?

clobrano:
Yes, you're right it's not the icon's fault and there's no way it can "fix
it", I was just surprised that there are multiple ways to show the broken
link (both emblem and icon)

PR #1871, opened 2020-02-08
Add icon for mimetype geo

madsrh:
Icon for geo json files and symlink to Google earth files.

image


madsrh:
? 

image

clobrano:
I like the third one, but I'm curious, why not the first set?

madsrh:
> why not the first set?

I prefer that one too. Remember the example of all the colourful backgrounds I posted here: https://github.com/ubuntu/yaru/issues/1885#issuecomment-583906239

They all have one thing in common and that is that the background is a solid or gradient - nothing more complex than that. It doesn't say that anywhere in the guidelines but it makes good sense and would eliminate the concern Frederik had about files being mistaken for apps

Feichtmeier:
I am fine with either the current one or the third one :+1:

PR #1872, opened 2020-02-09
Add missing device icon for media-floppy

madsrh:
VERY legacy but my VM has a floppy drive 😄 
I've used the $linkblue for the lines.

floppy

image


madsrh:
I've just now realized that metal piece is flipped from the symbolic version 🤦‍♂

ubuntujaggers:
@madsrh 

I did a tiny bit of pixel-fitting on the smaller sizes and took the opportunity to flip the metal piece, see what you think :+1:

madsrh:
Thanks you @ubuntujaggers for all your work!

PR #1876, opened 2020-02-09
Add icon for mimetype icc profile

madsrh:
Here's an icon for [ICC ](https://en.wikipedia.org/wiki/ICC_profile)profiles. Again, just reusing what we already have. Other themes use a circle or a document with a drop 💧 . Could this work? 👇 

image

Just in case you would like the text added (like the sound icons) there's a hidden layer that can be turned on.
image


madsrh:
?

image

clobrano:
This is neat :D

Feichtmeier:
It looks very good indeed! :+1: 

Just one tiny question: is the shadow not a little bit too strong?

madsrh:
> Just one tiny question: is the shadow not a little bit too strong

They are. I didn't push it yet so I'll fix it according to the guidelines before posting (also it uses the mask 🙈)

Feichtmeier:
I am not sure about that request to remove a feature that modern SVGs provide. I would not target this currently. I think it is to be discussed

madsrh:
I've pushed the [drop](https://github.com/ubuntu/yaru/pull/1876#issuecomment-584398045) design. Feel free to merge if you :+1: 

NOTE: This uses the cut effect that gusbemacbe mentioned was problematic for downstream. We could merge it now and fix it later (because it's not a problem here).

friendlyapple:
Good job @madsrh! I just noticed that there is a bug in:
icons/src/fullcolor/mimetypes/application-vnd.iccprofile.svg

img

madsrh:
Hmmm, odd 🤔 
It looks fine here and in master. Could you check again if you did something wrong?


image

friendlyapple:
Yes the problem is on the master branch. By clicking on the link below, do you see the problem?
https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/mimetypes/application-vnd.iccprofile.svg

My vector graphics editor ([Sketch](https://www.sketch.com)) also cannot display the smaller icons.

clobrano:
I think it's just a problem with the viewer. I downloaded the svg and it's visible with both gimp and inkscape, but github cannot render the small icons.

friendlyapple:
Really strange...

PR #1877, opened 2020-02-09
Added device icon for media-flash

madsrh:
Okay this is a bit of a longshot because I can't test this icon. I tried to adapt [the Suru++ icon](https://github.com/Bonandry/suru-plus-ubuntu/blob/master/Suru++-Ubuntu/devices/scalable/media-flash.svg), but I couldn't reuse the standard Yaru template borders, so I added a stroke. 

@ubuntujaggers could you have a look?

image


ubuntujaggers:
Will check the edging tonight 👍

ubuntujaggers:
Do you have your heart set on the little cutaway on the right?  I'm in the process of refactoring into Suru-style edging:

image

madsrh:
> Do you have your heart set on the little cutaway on the right? 

No 😄 I'm sure it isn't even visible with the small sizes.

Are you able to test it?

ubuntujaggers:
No, lol.  Shall I push?

madsrh:
> No, lol. Shall I push?

Yes, please 😃 

What do you think about the testing @Feichtmeier ?

clobrano:
LGTM :+1: 
image

Feichtmeier:
> > No, lol. Shall I push?
> 
> Yes, please 
> 
> What do you think about the testing (or lack of) @Feichtmeier ?

Not sure - does it have the same highlights as the other device icons?

madsrh:
Awesome, you had a SD card reader @clobrano 😄 but the PR you tested was not the tweaked version because @ubuntujaggers didn't push his changes yet.

Glad to see it worked 👍

ubuntujaggers:
image

Give it a whirl...

clobrano:
I'll have a go tomorrow :+1:

madsrh:
So much sharper! Thank you!

clobrano:
Here is the new icon
image

image

Again, my only concern is that the gray gradient seems stronger respect the other device icon, but It's not a major one, good work :+1:

ubuntujaggers:
I think this is an optical illusion... both the background gradient and the secondary grey are taken from the drive icons. However, because there's a big area of darker grey rather than a fine glyph, maybe the overall impression is a "darker" icon?

clobrano:
Oh, it might be indeed, sorry then :D

ubuntujaggers:
Optical illusions are still a factor - we can try tweaking either shade and see if it restores visual balance 👍

clobrano:
Well, only if you think it's a problem as well, otherwise feel free to
merge it

Feichtmeier:
@madsrh @ubuntujaggers are you sure that the shadows and highlights are correct? I'm not sure :thinking:

ubuntujaggers:
The baseplate is the same width, but more of the colour is bleeding into the next pixel:

image

Looks like this is due to a slight error in 48px - try now?

Also the highlight was 30% opacity instead of 40% opacity - try now  :)

ubuntujaggers:
@clobrano 

I think the juxtaposition with the server icon is problematic in the screenshot - the server has folds which make it brighter than average.

If you set it by the hard drive, it looks right to me:

image image

...so maybe the problem is the server being too bright?

Feichtmeier:
@ubuntujaggers 
Okay then it was only the glitch + some optical illusion :)

About the highlights: didn't we have highlights here in other icons?
image

ubuntujaggers:
Oh, I see what you mean - there doesn't seem to be a hard and fast rule in Suru, some pictogram elements have edging and some are completely flat, but I like highlights and am always happy to try adding some  :)

clobrano:
> ...so maybe the problem is the server being too bright?

Yes, I think you are right. HD and sdcard are quite aligned and server looks a bit brighter

ubuntujaggers:
Try this:

image

Feichtmeier:
Imo this looks good :thums_up:

PR #1879, opened 2020-02-09
Added icon for mimetype application-pgp-signature

madsrh:
I'm not certain that this is the best icon. It's for pgp signature. Again, feel free to close this PR. What do you think?

image

image


Feichtmeier:
Looks good and I think inline with the others :+1: 
Is this for a signed or unsigned?
The X looks kind of "something bad happend".

madsrh:
> The X looks kind of "something bad happend".

You're right. I think I'll close this one a have a second attempt. Also perhaps keep the three in line - pgp key, pgp signature, php encrypted). Although a quick fix would be to symlink the three.



I have a few more icons that I have the same mehhhh feeling about, so I'll refine this before posting.

Issue #1727, opened 2020-01-12
Colour selection for window snapping


popey:
## Expected Behavior

With the latest build of Yaru on 20.04 (built from source as per the excellent Contributing document), I noticed the window snapping guide is bright orange.

Given during the recent desktop yaru design sprint we committed to reducing the overall amount pf orange on the desktop. I wanted to start a discussion whether we think the orange snap is still appropriate, and if not, what would it look like when changed?

![Screenshot from 2020-01-12 10-35-20](https://user-images.githubusercontent.com/1841272/72218606-81c58b00-3534-11ea-8f0e-0df1cd64e50f.png)

clobrano:
I had a look at the code and, luckily, it's entirely in our powers

![image](https://user-images.githubusercontent.com/2883614/72220803-7c743a80-354c-11ea-8b0d-9b5085c2c06e.png)

madsrh:
If we can change border and background independently, how about white solid border and background only 20-25% transparent white? I was thinking that this would be sort of related to the highlight on hover 🤷‍♂ 

![image](https://user-images.githubusercontent.com/3986894/72225276-cc222880-3583-11ea-9654-03686ef1a4b0.png)

clobrano:
I still don't know how you can be so precise with a mock-up :D 

This is on the real thing

![image](https://user-images.githubusercontent.com/2883614/72225961-66ce3780-3583-11ea-8563-2f1a090e7b8f.png)

clobrano:
So, @popey, @madsrh, any preference? White is probably more neutral as orange alternative, but I don't have a strong opinion here

madsrh:
I prefer the white one (if it fits with the theme). Can you make a test PR?

I tried to make a PR for this myself for testing, but I didn't succeed. So I'm keen to see how you did it 😃

popey:
I really like the white/porcelain, that's much better than 🍊 or 🍆. 
How does it look with each of the three theme options?

clobrano:
Being a gnome-shell thing, it's the same on all three options

popey:
I meant can you show a mockup of it in all three?

clobrano:
Oh sorry, sure I can do that

clobrano:
Uhm, I am afraid this white is actually a problem with white backgrounds

![image](https://user-images.githubusercontent.com/2883614/74110607-4d260d00-4b8e-11ea-96a6-eda88bc25b33.png)

clobrano:
[...]

PR #1883, opened 2020-02-09
shell/gtk: use silver color for mouse selection

clobrano:
Replace current orange with a silver-ish color for mouse selection

- gtk rubberband (e.g. Nautilus folder selection with mouse)
- shell rubberband (e.g. Desktop Icon selection and area screenshot)
- shell tile-window-preview (e.g. snapping windows on left/right side of the screen)

Co-authored-by: MadsRH <madsrh@gmail.com>

Closes: #1727

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -1247,8 +1247,9 @@ StScrollBar {
 
 // Rubberband for select-area screenshots
 .select-area-rubberband {
-  background-color: transparentize($selected_bg_color,0.7);
-  border: 1px solid $selected_bg_color;
+  background-color: transparentize($ash, 0.8);
+  border: 1px solid transparentize($silk, 0.2);
+
 }
 
 // Pointer location
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -812,7 +812,7 @@ StScrollBar {
 /* Tiled window previews */
 .tile-preview {
   background-color: transparentize($ash, 0.8);
-  border: 1px solid $silk;
+  border: 1px solid transparentize($silk, 0.2);
 }
 
   .tile-preview-left.on-primary {
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -811,8 +811,8 @@ StScrollBar {
 
 /* Tiled window previews */
 .tile-preview {
-  background-color: transparentize($ash,0.87);
-  border: 1px solid darken($silk, 10%);
+  background-color: transparentize($ash, 0.8);
+  border: 1px solid $silk;
 }
 
   .tile-preview-left.on-primary {
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -811,8 +811,8 @@ StScrollBar {
 
 /* Tiled window previews */
 .tile-preview {
-  background-color: transparentize($selected_bg_color,0.5);
-  border: 1px solid $selected_bg_color;
+  background-color: transparentize($porcelain, 0.8);
+  border: 1px solid $porcelain;
 }
 
   .tile-preview-left.on-primary {
--- a/gnome-shell/src/gnome-shell-sass/_common.scss
+++ b/gnome-shell/src/gnome-shell-sass/_common.scss
@@ -794,14 +794,14 @@ StScrollBar {
 
   StEntry { @extend %bubble-entry; }
   .button {
-    &, &:hover, &:focus, &:active, &:disabled { 
+    &, &:hover, &:focus, &:active, &:disabled {
       box-shadow: none;
       border-color: $_bubble_borders_color;
     }
     background-color: $bg_color;
     color: $fg_color;
     &:hover { background-color: $_hover_bg_color; }
-    &:active { 
+    &:active {
       background-color: $selected_bg_color;
       color: $selected_fg_color;
     }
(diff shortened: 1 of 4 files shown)


madsrh:
It might be my stupidity, but I can't test this 😞 Don't know what I'm doing wrong.

clobrano:
After refreshing the shell `alt-f2 r`, move a window till the border of the screen to snap it on one side

Feichtmeier:
image

This works but shouldnt this also be adapted?

image

clobrano:
The point is that one is an indication (tile preview) and the other is a selection

madsrh:
Yes, the rubberband needs to match. This white works and is very neutral but I wonder if white is visible in Nautilus on the white background?

clobrano:
I tested it and choose a transparency that should be enough to make it visible on bright backgrounds

Feichtmeier:
I think the snappy thing to select elements is more like a "hover". When you are done, then the elements are selected. But I have the feeling that if we would make it white, the selected items on the desktop would also be white :)

clobrano:
It might work

clobrano:
> I tested it and choose a transparency that should be enough to make it visible on bright backgrounds

I take back that. This white is really not visible with white backgrounds

image

clobrano:
Here are some variations (consider that this PR has porcelain with 80% of transparency)

Porcelain 70% transparent
image

Yaru light (it's invisible :disappointed:)
image


Aubergine 80% transparent?
image
image
image

Feichtmeier:
I agree, rather aubgerine than white! Or just keep orange :man_shrugging:

madsrh:
Can you add a box shadow to the white version? So if it's on a white background, you could still see the frame.

The aubergine is definitely visible but isn't it odd to have orange for selection and then suddenly aubergine for window snapping? I might prefer to just leave it as it is, IDK

clobrano:
> The aubergine is definitely visible but isn't it odd to have orange for selection and then suddenly aubergine for window snapping? I might prefer to just leave it as it is, IDK

If we change snapping, I'll change selection as well

clobrano:
> Can you add a box shadow to the white version? So if it's on a white background, you could still see the frame.

unfortunately, since the background is transparent, the box shadow will be visible behind it

madsrh:
What about something like:

```
.tile-preview {
  background-color: transparentize(#3d3d3d,0.9);
  border: 1px solid #ffffff; }
```

clobrano:
Here is the test. This colour seems more flexible indeed :+1: , maybe just with yaru dark, and in general dark backgrounds, the shadowed area is not visible enough? Considering this, the orange one might be better?

image
image
image
image
image
image
image

madsrh:
I don't have a strong opinion here, but I will say that these latest screenshots are more neutral, which I really like. Replacing orange with purple doesn't really change anything.

I see the problem on dark backgrounds but it is still visible because it has the border (actually MS Windows only uses a border, no background).

clobrano:
I see, so I can try some more iteration to find a good balance

madsrh:
I can't take screenshots but here's another attempt at the neutral version:

```
.tile-preview {
  background-color: transparentize($ash,0.77); 
  border: 1px solid $selected_fg_color;
}
```

madsrh:
Then we could do this for GTK:

```
.rubberband,
rubberband {
  background-color: transparentize($ash,0.77); // transparentize(#cccccc,0.75);
  border: 1px solid lighten($ash, 5%);
```
dark

light

clobrano:
> I can't take screenshots but here's another attempt at the neutral version:
> 
> ```
> .tile-preview {
>   background-color: transparentize($ash,0.77); 
>   border: 1px solid $selected_fg_color;
> }
> ```

this one seems perfect!

madsrh:
> this one seems perfect!

THIS one is perfect :smile: 

```
  background-color: transparentize($ash,0.8);
  border: 1px solid $silk; 
```

Feichtmeier:
Yoooo If you guys are on it, you think we could get the same effect with osd fg/bg colours?

clobrano:
`$osd_bg_color` and `$osd_fg_color` are either quite dark (jet) and bright (#eeeeec).
The good of ash and silk is that they give better results in both light and dark variants

image

image

NOTE: I'm going to move latest changes on tweaks

Feichtmeier:
Okay - if we find a color later with desaturate(...) darken(..) lighten(...) of osd_bg/fg_color that matches ash and porcelain we can do this in a different PR. Did not want to block you :+1:

clobrano:
Sorry, but it doesn't make much sense tweaking other colors to have one we already have defined IMHO

Feichtmeier:
Okay - I personally would just prefer to not use colors from the palette outside of _colors.scss
I think it is semantically vague. Whereas colors fom _colors.scss are not since you know what they mean if you read it. Again, it's nothing super important

clobrano:
I understand your point, I just see it as we have a palette made of both semantic and normal colors, and we should use them both in the proper manner. Widgets that have same meaning might share a semantic color, as for selected state, but I don't think it's mandatory otherwise; anyway, this can be discussed later when a broader palette will be exposed in gtk.

Thanks!

friendlyapple:
I think the borders would also be better with a little transparency.

clobrano:
> I think the borders would also be better with a little transparency.

in rubberband, snap preview or both?

friendlyapple:
@clobrano Both, I think. But especially on the selection (rubberband I suppose), like macOS or Windows.

madsrh:
This seems broken in master (or I messed up my system) ATM


> I think the borders would also be better with a little transparency.

Sure, let's try it :smiley: 

Also we should apply this to rubberband for select-area screenshots too

clobrano:
Alright, it seems that shell's rubberband does not really style the border, but it's ok to have all the pieces synchronized. Shall I merge this?

madsrh:
Ok. Yes please 😀

Feichtmeier:
Yes, I just think we need to re-apply it once the big shell PR is merged

Issue #1814, opened 2020-01-26
Updating build system for coloured variants


clobrano:
Current build scripting has a lot of duplicated code and its optimization would permit supporting more color variants

#designfest

Issue #1331, opened 2019-05-01
meson: future version copy symlink instead of file


clobrano:
The following message shows that a future change in Meson will break our build

> Warning: trying to copy a symlink that points to a file. This will copy the file,
> but this will be changed in a future version of Meson to copy the symlink as is. Please update your
> build definitions so that it will not break when the change happens.

We use symlinks on `icons` only, so the possible solutions are:

1. Find a way to instruct meson to resolve the symlink before do the copy
2. Change `generate-symlinks.sh` into `generate-copy.sh`

Solution 1 would be the one with less impact in our filesystem, but I still didn't find anything. It is even possible that Meson developers will provide something new before the final switch.

Solution 2 is the easier to implement (I already tested a solution and I works fine), even if it requires a long first step to transform each symlink into actual file and the project size will increase a lot (consider that actual files are already copied into the filesystem, so the installation size won't change).

PR #1884, opened 2020-02-09
build: deduplicate meson.build and introduce Flavours

clobrano:
- default, light and dark themes are now built as Yaru _flavours_. It is easier now to introduce new flavours with different colorschemes
- deduplicate all the common meson.build code. The build instructions are all in `gtk/src/meson.build` file.
- deduplicate files: most of the original files are in the **default** flavour, while **light** and **dark** flavours have simlinks, however `gtk-dark.scss` for **default** and **light** flavours are symlink to `dark/gtk.scss` to have the same "prefer-dark" look and feel.

~~- replace `gresource` binary with actual assets and css files: this is a side effect of building all the files in the same directory (`gtk/src`). In place than a gresource binary, this change installs the actual `gtk.css`, `gtk-dark.css` and all the assets under `/usr/share/themes/<theme>`.~~ 

^ Fixed thanks to @jhenstridge hint and @3v1n0 meson.build in gnome-shell, used as reference

**NOTE**: For testing, I renamed the project **YaruFlavours**, so that in Tweaks you should see the entries *YaruFlavours*, *YaruFlavours-dark*, *YaruFlavours-light* alongside the usual Yaru to simplify the comparison. Before merging, I will change it back to Yaru.

Closes:  #1814
Closes:  #1331


jhenstridge:
If you can keep the gresource files, please do.  It massively reduces the number of files an application needs to open on start.

You can use the `.gresource.xml` file to map files to different paths within the resulting `.gresource` file.  We already use that to rename the CSS files in master:

    <file alias="gtk.css">gtk-generated.css</file>
    <file alias="gtk-dark.css">gtk-dark-generated.css</file>

The same can be done for the other assets if the paths in the source tree don't match what you want in the final install.

clobrano:
Thanks @jhenstridge! I saw something similar made by @3v1n0 in latest gnome-shell meson script and I'll try here as well

clobrano:
Hey @didrocks, let me know if you have time to review this, or if you feel we can merge it right away. I used it quite a lot, and it seems to work fine

clobrano:
Alright, I'm going to merge this 🤞

chrisjbillington:
With this change the Arch AUR packages are no longer able to build. Fix is in #2206

Issue #1868, opened 2020-02-08
Missing mime type for Nim-lang


jeremie2:
Hi!
A mime type icon for [Nim](https://nim-lang.org/) language files is missing. I tried to create a text-x-nim.svg file:
![nim](https://user-images.githubusercontent.com/30863448/74078068-277df400-4a26-11ea-8dad-c2a86b2528e6.png)

Is it OK?

madsrh:
Fantastic work @jeremie2 🎉 

Three things:
1. Could you make a PR so we can test this?

2. On my 19.10 default installation there is no support for .nim 😞 Did you install something that added support for this?

![image](https://user-images.githubusercontent.com/3986894/74082147-6b3d2180-4a57-11ea-8a98-f0da0c101ed2.png)


3. I know that we have other icons with a similar "issue", but could you add a shadow? I don't know if this is out of style - @ubuntujaggers will know 😉 
![image](https://user-images.githubusercontent.com/3986894/74082118-02ee4000-4a57-11ea-8cc1-165d57067609.png)

meetdilip:
Other white + logo mime types have a shadow through the middle. Existing icons

![application-vnd microsoft portable-executable](https://user-images.githubusercontent.com/6965443/74083407-b5d79180-4a89-11ea-8f2e-ad1deeb1285f.png)

![application-json](https://user-images.githubusercontent.com/6965443/74083410-b708be80-4a89-11ea-98ca-34a9a752ecf9.png)

It could be just the logos maybe.

Feichtmeier:
This might help:

https://snwh.org/suru/guidelines

jeremie2:
> Fantastic work @jeremie2 tada

Thank you!

> 3. I know that we have other icons with a similar "issue", but could you add a shadow?

I didn't put any shadow because I have seen that all the mime-type icons related to program language files got only a "flat" official logo with no shadow.
So I think this is a guideline.

<br>
That said, Huston we have problem.... I'm a noob :stuck_out_tongue:
It's the first time I'm doing something like that and a few things are actually not clear.<br><br><br>

>  2. On my 19.10 default installation there is no support for .nim disappointed Did you install something that added support for this?

I found a few howtos about adding a new line into `/etc/mime.types`, but didn't work..

Then I found [this tutorial](https://coderwall.com/p/qjda2q/create-new-mime-type-and-assign-an-icon-to-it-in-ubuntu) (custom xml file into `/usr/share/mime/packages`) and it worked.
[...]

PR #1888, opened 2020-02-10
added icon for mimetype nim

jeremie2:
Closes: #1868


madsrh:
Looks good 😀🎉

You only need to edit mimetypes.list if you add a symlink. If looks like you wrote the same filename two times

jeremie2:
> Looks good grinningtada
> 
> You only need to edit mimetypes.list if you add a symlink. It looks like you wrote the same filename two times

OK.. [noob mode/on]  😬

Should it be a string like:
```
text-x-nim.png text-nim.png
```

Also, shell I do another PR and this will not be accepted?

madsrh:
It is not required to add a symlink (maybe push without and as it later?). If you want to create a symlink then yes, that line looks better, but you also have to generate the link.

./generate-symlinks.sh -m text-x-nim

No need to create a new PR. Just add, commit and push again

madsrh:
Where did you find `text-nim`? I'm not sure this symlink is necessary 

Hmm, I don't know what to think about this: https://github.com/Bonandry/suru-plus-ubuntu/issues/49

My problem is that I can't test this PR. Anyone?

clobrano:
Let me try

clobrano:
```
 carlo ● ~ » mimetype example.nim*                                                                                                                                                                           ~
example.nim:  text/x-nim
example.nims: text/x-nim
```
image

I added the support for nim (not easy), but the icon does not show up

image

jeremie2:
@madsrh you are right, no need to create a symlink.

I repeated all the steps on a clean Yaru installation till the `render-bitmaps.py text-x-nim` command. ..and it  works.
I think last night I didn't realize that after the correct `render-bitmaps.py` command everything was already working. Sorry!

I'm going to send a new commit.

@clobrano is it possible that the .nim file in your screenshot is an empty file?
Because in that case it doesn't show any logo.

Here an example:
Schermata del 2020-02-11 18-10-21

clobrano:
They both have a minimal script

```
#!/usr/bin/env nim

echo "info"
```

the weird thing is that the file `example.nims` even works, `example.nim`
return an error, but both are recognized by nautilus and `mimetype` command.

Could you run `mimetype helloworld.nim` on your machine?

jeremie2:
> Could you run `mimetype helloworld.nim` on your machine?

Sure!

~$ mimetype helloworld.nim:
```
helloworld.nim: text/x-nim
```

clobrano:
okay, it's the same as mine, so the problem might be just my mimetype
configuration

clobrano:
Let's merge this, do you agree @jeremie2? Or do you need some more work?

jeremie2:
> Let's merge this, do you agree @jeremie2? Or do you need some more work?

Yes I agree, for me it's ready :wink:

clobrano:
Cool, thanks for your work, really appreciated :+1:

gusbemacbe:
@jeremie2, @clobrano and @madsrh 

I was called when you referenced the topic on my Suru++ Ubuntu project.

You must remember that Linux does not support it yet, therefore, you have to create a new file association, using File Type Manager on GTK-based environment or using File Association from KDE System Settings on KDE. 

If you want Linux to support it, you must report to FreeDesktop and GNOME developers.

clobrano:
> You must remember that Linux does not support it yet

Yeah, I noticed, I had to create some files to test it properly, but if the
association is in place, the icon is available, so I think it's good enough

> If you want Linux to support it, you must report to FreeDesktop and GNOME
developers.

This might be cool, but it's out of the scope of this project for me

jeremie2:
After some tests I have finally understood why `.nim` files were visible on my system.

To understand how mime types icons get loaded I followed a tutorial about creating an xml file. Then I deleted the files I have created but I didn't noticed that some strings remained in some existing files... that's why everything was working on my system, while **at this stage nothing should work!**

Here is what I've done:

1. Created a nim.xml file:
    ```
    sudo touch /usr/share/mime/packages/nim.xml
    ```
1. Saved inside it the following strings:
    ```xml
    <?xml version="1.0" encoding="UTF-8"?>
    <mime-info xmlns='http://www.freedesktop.org/standards/shared-mime-info'>
	    <mime-type type="text/x-nim">
		    <sub-class-of type="text/plain"/>
	    	<comment>Nim Text file</comment>
	   	<glob pattern="*.nim"/>
	   </mime-type>
   </mime-info>
    ```
1. Run the command:
   ```
   sudo update-mime-database /usr/share/mime
   ```

The last command is going to add some lines into several files in **`/usr/share/mime`**:
* **types**:
   ```
   text/x-nim
   ```
* **globs** and **globs2**:
   ```
   text/x-nim:*.nim
   ```
* **subclasses**:
   ```
   text/x-nim text/plain
   ```
Unfortunately some of these strings were still there after I've run *sudo ninja install*.. 🙁

<br>

>   ...I had to create some files to test it properly, but if the association is in place, the icon is available, so I think it's good enough

Yes, at least the icon is already there 🙂

Well, I wanted to start with something "easy" but it went a bit more complicated 😛
Anyway, I'll have a look at freedesktop [docs](https://www.freedesktop.org/wiki/Software/shared-mime-info/) and I'll see...

Issue #1725, opened 2020-01-11
Remove hardcoded colors from Software Center


madsrh:
Seems like there's no way we can change these hardcoded colors. Would fixing this upstream be considered low hanging fruit for the desktop-team @flexiondotorg?

- Review stars
- Checkmark for installed applications
- Progressbars

![image](https://user-images.githubusercontent.com/3986894/72209691-42098f80-34b1-11ea-81ef-28bfd7bd36bc.png)

![image](https://user-images.githubusercontent.com/3986894/72224551-fec82300-357b-11ea-89a0-39d0a49dd487.png)

Here's the mockup from https://github.com/ubuntu/yaru/issues/226

![36940877-252a4794-1f4e-11e8-8dfb-5c385bf7809e](https://user-images.githubusercontent.com/3986894/72209560-10dc8f80-34b0-11ea-84aa-2a9c2fd647e6.gif)

Feichtmeier:
Here is the old upstream issue for this, still not closed:

https://gitlab.gnome.org/GNOME/gnome-software/issues/553

Feichtmeier:
Good news :tada: 
![image](https://user-images.githubusercontent.com/15329494/74218591-bac75b80-4caa-11ea-87b1-072066112d04.png)

Robert applied my patch really fast https://bugs.launchpad.net/ubuntu/+source/gnome-software/+bug/1862056

Now the bonus work is to get this into upstream + we can now remove the "box-shadow" hack for the .installed-icon from _aps.scss

clobrano:
Awsome!

PR #1893, opened 2020-02-11
Apps: remove the box-shadow hack for counter-label

Feichtmeier:
It's no longer needed since gnome-software now uses success_color for this: https://github.com/ubuntu/yaru/issues/1725


Feichtmeier:
@clobrano should we keep it for older versions?

clobrano:
Yeah, I think so

Feichtmeier:
I'll see if the box Shadow actually goes beyond the borders

Feichtmeier:
ya

No negative effect, let's keep it

PR #1897, opened 2020-02-11
Taking folds off network-server

ubuntujaggers:
I actually think this looks good without the folds (if simple) and will go much better with other devices:

image

image image


ubuntujaggers:
Actually, let me darken the grey to match the dark grey on the hard disk.

Feichtmeier:
Looks definitely better. Previously the three folded effect it looked a little bit like a squeezebox :)

madsrh:
I kind of liked that you could see the separate units. How about adding some lines? 

image

ubuntujaggers:
Something like this?

image

Personally I like it without the lines now...

madsrh:
> Personally I like it without the lines now...

It just isn't as obvious what this is IMO because it's more common to the icon as individual server. This is not important to me, just wanted to share my point of view 😃

clobrano:
Considering that the details are very little, the rows might improve recognition at small sizes, but for both versions are an improvement over the current one

Feichtmeier:
Any consensus here from you guys?

madsrh:
That's more +1's than my +0 :smile: I've shared my concerns so feel free to merge it

friendlyapple:
I think like @madsrh 😄 The lines are cool!

ubuntujaggers:
I'm going to merge now because it's better than the concertina, but I'm not opposed to pushing an attractive version with lines if we have it in time for 20.04 👍

madsrh:
I didn't create an issue for this @ubuntujaggers, but I think we should revisit this icon. Especially in the small sizes it isn't working IMHO. Here's a few others in comparison: 

image

PR #1898, opened 2020-02-11
Light and dark theme headerbar button border improvements

Feichtmeier:
- unify the color of the :checked button borders in the headerbar, because uneven border colors make the button look more pressed than necessary without improving the legibility
- this is also more consistent with the default theme :checked headerbar buttons

Before:
image
image

After:
image
image

Default theme:
image

@madsrh @clobrano @ubuntujaggers ?

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -71,6 +71,17 @@ headerbar *, button * {
   -gtk-icon-shadow: none;
 }
 
+// Leave border-color of headerbar buttons evenly colored
+// But darken the border-color a little bit to keep legibility
+// This is consistent with the "default theme"
+headerbar, .titlebar {
+  stackswitcher button:checked,
+    button.toggle:checked {
+      border-color: darken($borders_color, 6%);
+      border-top-color: darken($borders_color, 6%);
+    }
+  }
+
 // blue spinner
 spinner {
   &:not(:backdrop) {


Feichtmeier:
Peek 2020-02-11 20-05
Peek 2020-02-11 20-04

clobrano:
It's not so clear :D, can you rephrase it? The title as well

Feichtmeier:
Done - hope this is more clear? :D

clobrano:
Yes thanks. I think I prefer the press effect in light/dark master though 😖

Feichtmeier:
Why exactly? It's less visible in master and adds a 3d effect. There is no benefit and it looks inconsistent with the rest of the theme

clobrano:
Because the press effect is 3D by definition. Basically the PR replace the
press with a stronger border

Feichtmeier:
The press effect is pressing the leveled up button down. 
It's also less visible. This was the reason we've changed the default theme btw. (Which also has no second layer 3d effect)

clobrano:
I know, but now that I see the difference closely, I prefer the 3D effect.
It's just my opinion though, if the others prefer this PR I have
no problems :)

ubuntujaggers:
Looking at the gifs I slightly prefer the change - but possibly because my eyes don't interpret the uneven border as a 3D effect, it just looks like one of the edges being darker (almost like a slight mistake rather than a design choice).

clobrano:
Alright then :), feel free to merge it +1

PR #1900, opened 2020-02-12
Remove orange tint on icons

madsrh:
A third (#1831 and https://github.com/ubuntu/yaru/pull/48) attempt at removing the orange tint on the icons in Nautilus.

This is VERY close, but I need the wizard skills of @Feichtmeier or @clobrano to fix the label color when right clicking :disappointed: 
I don't know why, but the label is gray in the dark theme but "invisible" in the light theme.

files

filesdark 01-47

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -124,19 +124,17 @@ $ambiance: null !default;
         border-radius: 0px 0px 0px 0px;
     }
 
- widget.view {
-          background-color: rgba(255, 255, 255, 0.1); // reduces orangeness of selected folder icons
-          color: $text_color;
+    widget.view {
+      background-color: rgba(255, 255, 255, 0.1); // reduces orangeness of selected folder icons
+      color: $text_color;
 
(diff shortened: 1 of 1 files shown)


clobrano:
I think I got it

clobrano:
image

also I fixed the backdrop a little

nautilus

Feichtmeier:
I love when orange becomes gray in the backdrop. Offtopic but it would be so nice if this would become the backdropselectedbgcolor :)

madsrh:
Can I merge this? 😄 

One tiny thing: when right clicking shouldn't the label still be colored? You can't really see which file is selected. Definitely not an issue blocking this PR, but would be nice.

clobrano:
> One tiny thing: when right clicking shouldn't the label still be colored?

It should and it is in the dark variant, but not in the light one, and I am still unsure why. I'll try again tomorrow

madsrh:
@clobrano how would you feel about merging this one now a creating a new issue for the right clicking label?

clobrano:
Yes, sorry didn't really have enough time yesterday, it sounds a good idea :)

PR #1902, opened 2020-02-12
Default theme: headerbar separator improvements

Feichtmeier:
The separator color in the backdrop was too strong.
With this commit we bind the color of the separators in the headerbar of the default theme to the headerbarcolor which also improves the :backdrop color. Changing the headerbar color should now change the separator color accordingly.

Before:
image
image

After:
image
image

Note: if you think the active window color needs improvement, we can do this. But anything is better than a set hex color :D (which was probably me in the rush of 19.10)

### DIFF
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -1,7 +1,7 @@
-
-$_headerbar_button_bg_color: lighten($headerbar_bg_color, 6%);
-$_headerbar_button_border_color: darken($headerbar_bg_color, 9%);
-$_backdrop_headerbar_button_border_color: $headerbar_bg_color;
+// Color variables dependant on $headerbar_bg_color defined in _colors.scss
+$_button_bg_color: lighten($headerbar_bg_color, 6%);
+$_button_border_color: darken($headerbar_bg_color, 9%);
+$_backdrop_button_border_color: $headerbar_bg_color;
 
 // overwriting the headerbar styling from common, for the inverted look in the ambiance theme
 headerbar,
@@ -45,17 +45,17 @@ headerbar,
   filechooser .path-bar.linked>button,
   .path-bar button {
 
-    @include button(normal, $_headerbar_button_bg_color, $headerbar_fg_color);
+    @include button(normal, $_button_bg_color, $headerbar_fg_color);
     &.flat {
       @include button(undecorated);
     }
 
     &:hover {
-      @include button(hover, $_headerbar_button_bg_color, $headerbar_fg_color);
+      @include button(hover, $_button_bg_color, $headerbar_fg_color);
     }
     &:active,
     &:checked {
-      @include button(active, $_headerbar_button_bg_color, $headerbar_fg_color);
+      @include button(active, $_button_bg_color, $headerbar_fg_color);
       border-color: black;
     }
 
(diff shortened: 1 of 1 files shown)


clobrano:
Backdrop is improved indeed, but on focus was better before

Feichtmeier:
Ok. Will change. A bit brighter you say?

clobrano:
Same as it is in master it's ok for me

Feichtmeier:
Ok I'll check at home. Better screen

Feichtmeier:
@clobrano 
image
I also renamed the vars and also made the ex "dark border color" now "headerbar_button_border_color" dependant on headerbar_bg_color instead of a set value

gtk/src/default/gtk-3.20/_headerbar.scss
```diff
@@ -1,7 +1,7 @@
 
-$_headerbar_button_bg: lighten($headerbar_bg_color, 6%);
-$_dark_border_color: rgb(20, 19, 19);
-$_btn_backdrop_border: $headerbar_bg_color;
+$_headerbar_button_bg_color: lighten($headerbar_bg_color, 6%);
+$_headerbar_button_border_color: darken($headerbar_bg_color, 9%);
```
clobrano: that's better indeed

gtk/src/default/gtk-3.20/_headerbar.scss
```diff
@@ -1,7 +1,7 @@
 
-$_headerbar_button_bg: lighten($headerbar_bg_color, 6%);
-$_dark_border_color: rgb(20, 19, 19);
-$_btn_backdrop_border: $headerbar_bg_color;
+$_headerbar_button_bg_color: lighten($headerbar_bg_color, 6%);
```
clobrano: I know this has started by me, however, since we are inside the `_headerbar.scss` file, I would drop the *_headerbar* prefix everywhere, it is just superfluous. Eventually, same story for the *_color* postfix for background variables, since it can only be a color.

Feichtmeier: I agree on the headerbar suffix
But I would keep
Border
bg
or
fg

Feichtmeier: It could also be a whole background just like in drawing

clobrano: border, bg and fg are correct, keep them, while for background it's not used anywhere as whole background here, so I don't expect misunderstandings

Feichtmeier:
Just have a look now. Also removed the box Shadow now correctly. Previously it had no effect since switches never had it. Only the sliders.

About the variable names: I think it's clearer to say what the variable belongs to. A whole background, which can consist of gradients plus edges and what not, or simply a color

Feichtmeier:
Also a border can be
$border: 1px solid $border_color

clobrano:
> Also a border can be
> $border: 1px solid $border_color

indeed, I was only referring to `background`

Feichtmeier:
A background can also consist of more than just a colour:
```scss
background: $c $gradient;
```

clobrano:
> A background can also consist of more than just a colour:
> 
> ```css-scss
> background: $c $gradient;
> ```

clobrano:
sorry, I wanted to close the message, not the PR

Feichtmeier:
Np. I'll try to revert the box Shadow commit later

Feichtmeier:
Yo now we have a nice case for squash @clobrano 
This commit message gets generated, do you like it like this or shall we just remove adding then then reverting the box Shadow commit?
Screenshot_20200212-224310

clobrano:
For the changelog, the only thing that matters is the commit title, so something like "lighter headerbar separator in backdrop" might be enough.
About the commit message, I'd say just clean it up a little and remove the lines about the box-shadow :+1:

PR #1904, opened 2020-02-12
Dark theme: darken entries in notebooks

### LINKED ISSUE #1784: Entryfields in dark theme are hard to spot
Entryfields are really hard to spot:
image

Feichtmeier:
Since notebooks and entries both use $base_color, entries drown in the dark theme, looking almost disabled. Darkening them for the dark theme only in notebooks helps here. We reduce the change for notebooks only since it is safe to say that the $base_color won't be changed by apps. Everywhere else it is not clear where the entries are drawn.

Closes #1784


madsrh:
LGTM :smiley:

Issue #1903, opened 2020-02-12
tortoisehg nautilus extension uses symbolic instead of colour icons


chrisjbillington:
Tortoisehg's nautilus extension is ending up with symbolic emblems instead of full colour ones:

![image](https://user-images.githubusercontent.com/1044087/74359407-37cd0080-4d91-11ea-80fa-4dc490865566.png)

Looking at its source code, it is passing the following strings to `Nautilus.FileInfo.add_emblem()`:

* added: `'list-add'`
* clean: `'default'`
* modified: `'important'`
* untracked: `'dialog-question'`

Not clear if `Nautilus.FileInfo.add_emblem()` automatically prefixes `'emblem-'` to the icon names before looking them up, but it appears to be getting:

* `'list-add'` -> `/usr/share/icons/Yaru/scalable/actions/list-add-symbolic.svg`
* `'important'` -> `/usr/share/icons/Yaru/scalable/emblems/emblem-important-symbolic.svg`
* `'dialog-question'` -> `/usr/share/icons/Yaru//scalable/status/dialog-question-symbolic.svg`
* `'default'` -> `/usr/share/icons/Yaru/scalable/emblems/emblem-default-symbolic.svg`

It seems that non-symbolic icons for these are not present in Yaru.

tortoisehg also uses passes "danger" to `Nautilus.FileInfo.add_emblem()` for unmerged files, this one does end up with a colour icon, presumably `/usr/share/icons/Yaru/<size>/emblems/emblem-danger.png` since that one is present in Yaru:

![image](https://user-images.githubusercontent.com/1044087/74360630-6ea41600-4d93-11ea-81a9-4125803275cc.png)

PR #1905, opened 2020-02-12
Emblems update

madsrh:
New icon for `list-add` and `dialog-question` and a bunch of symlinks.
Added an emblem background to `emblem-system` to match the others.

This should fix https://github.com/ubuntu/yaru/issues/1903. Could you give it a try @chrisjbillington ?

image


chrisjbillington:
Much better!

image

PR #1909, opened 2020-02-13
Making folder icons lighter

ubuntujaggers:
Give it a whirl!

image


friendlyapple:
@ubuntujaggers the folders are a little too bright. Would it be possible to find a compromise between the current folders and these ?

For the moment, I keep a preference for the current folders.

Feichtmeier:
Making them darker than this would barely make any difference to the current situation where some people think that the contrast to the white window is too high @friendlyapple

ubuntujaggers:
Tbh I think I prefer master, how many are in favour? Happy to go with majority.

friendlyapple:
@ubuntujaggers I prefer the current folders (master) like you. The folders of this PR are too bright.

Feichtmeier:
@ubuntujaggers master looks definitely better in the dark theme. And I can't say that it looks too strong in the light theme. So I probably agree with you. Let's wait for the rest

fcole90:
I'm not sure these are the best version we could have, but personally I like them much more than the darker ones. I feel like this colour to be more neutral, so that the mood is mainly conveyed by the accent colour hinted by the folder inside, rather than from the outer gray :relaxed: 

Pertaining the contrast with the dark theme, I didn't feel them being to bright, but I don't use it daily, so this may affect my opinion :thinking: 

light_v_dark_folders

fcole90:
image

Feichtmeier:
You are not wrong @fcole90 
I think they obviously look better in the light theme(s). 
But I think the master icons look better in the dark theme. 
I guess we need to understand on which theme we want to focus on. 
An argument for the light theme(s) could be that the mixed theme is the default theme.

real-amano:
Hmm. I guess that making it theme dependant is still not possible. To have Yaru-Bright, Yaru-mixed icons (a symlink to Yaru bright) and Yaru-dark, which would be a copy of Yaru-bright but with some icons adapted for contrast.

fcole90:
@Feichtmeier I think it would make sense indeed to focus on the mixed theme, which is the default, as you say. It seems Mac OS did the same, using icons which likely fit better the light theme. :thinking: 

image

friendlyapple:
Maybe we can find a compromise between the two styles? We can slightly lighten the current folders but not as much as those of this PR?

Feichtmeier:
Any update here @madsrh @clobrano @ubuntujaggers ?
Pref for the light themes or the dark theme? :)

clobrano:
I still prefer master, or at least this is not an improvement for me respect master

madsrh:
> I still prefer master, or at least this is not an improvement for me respect master

+1

ubuntujaggers:
I use mixed theme and actually prefer master, so it's not a light versus dark for me. Perhaps we should leave it as it is for now but be open to trying more things?

clobrano:
I agree. I think that the solution might be out of the simple "make them lighter/darker"

Feichtmeier:
Okay - closing this then, since it looks that none of you three really likes this solution and I am also not very sure about this

PR #1918, opened 2020-02-14
Increase contrast for dark buttons and switches

Feichtmeier:
Before:
image_2020-02-14_15-43-49

After:
image_2020-02-14_15-46-55

WIP: switch screenshots coming later


clobrano:
It looks good to me. It's not a complete solution, but it improves it a little (PR on the right side)

image

Feichtmeier:
Switches and headerbar buttons after:
image

Switches and headerbar buttons before:
image

Feichtmeier:
@madsrh @clobrano you guys decide, I am pretty neutral on this. At least I've found a solution which leaves buttons who don't use $bg_color untouched :)

clobrano:
Switch BG does not change (still #585858), button BG changes from #434343 to #484848. The improvement is probably visible only on side-by-side comparison

Feichtmeier:
Ah switch bg does not change because it blends with the $base_color I think. Right, better screenshots would be better

madsrh:
I tried this branch a bit last night and IMO it's much better except for `$popover_bg_color: $bg_color;` but I guess we can't have both 🤷‍♂ And this seems like a good tradeoff.

IDK if making the popover 1% lighter would make any difference, since 2% is the same color as the buttons have.

PR #1920, opened 2020-02-14
Adding new non-uniform app icon templates

ubuntujaggers:
As step one in my icons spring clean, and to help new contributors, I've removed the old squircle template and added one for each of our core icon shapes.

I took the opportunity to do the following:

- Standardise and correct the drop shadow on the full size icon
- Remove unnecessary clipping (we can discuss the policy on clipping in full, but there's no point using it when it isn't needed)
- Other minor improvements like the 16px circle icon, which was a tiny bit wonky.

Here they are:

image

image

image

image


Feichtmeier:
Fantastic work Stuart!
Thank you!

ubuntujaggers:
I think we can make the circle in the smallest size bigger, it's currently 14px to make room for a barely visible 1px drop shadow. The other 16px icons forgo this because space is a premium. I'll update the template and then update all the actual icons.

ubuntujaggers:
Dunno how to change an existing PR to draft so am changing the title for now...

Issue #1914, opened 2020-02-14
Missing label background when right clicking


madsrh:
A continuation of https://github.com/ubuntu/yaru/pull/1900 
Missing label background when right clicking in Nautilus
![image](https://user-images.githubusercontent.com/3986894/74530175-bf4d7780-4f29-11ea-8140-74707c6c7ca5.png)

PR #1921, opened 2020-02-14
nautilus: complete do not color folder on selection

Feichtmeier:
- only style .nautilus-canvas-item, not all widget.view
- remove the outline ring, not needed because the focus is already orange and it corrupts the label
- remove the background then bring it back for dir(ltr)

Peek 2020-02-15 00-35

@clobrano @madsrh 
I hope you like the backdrop color
I also removed the outline ring because I don't think that it's nessecary because I think there is no room for the focus ring in this tiny label - WDYT?

Closes: #43
Closes: #1914


madsrh:
Looks good @Feichtmeier 

Fixes #1914

PR #1924, opened 2020-02-15
All png images are optimized for low ram usage

paledega:


paledega:
I used optipng.

```shell
#creating png list
find -iname *.png > pnglist.sh
#regex 
sed -i "s/^/optipng -o7 /g" pnglist.sh
#run generated script
sh pnglist.sh
#clear
rm -f pnglist.sh
```

clobrano:
Hi @paledega, thanks for contributing. Could you explain better the original problem and how this PR should solve it?

friendlyapple:
@paledega There is no problem of ram usage. The only argument would have been to optimize disk space... but Yaru is already incredibly light for such a complete theme.

Why lose quality, if there is no problem?

Feichtmeier:
If the quality didn't drop and the size is still reduced it's okay I think?

ubuntujaggers:
This is a ginormous PR that replaces thousands of files.  I would suggest:

- Do some comparisons/testing of individual pngs to check the output is as good as the original pngs
- If so and if possible, **build this into the full colour export script** so it's part of the process
- ...and also do all the existing pngs as per this PR?

If I understand Optipng correctly, there should be zero loss of quality  :)  but I think we should proceed carefully, given the size of the change, and build it into our process so we don't have to separate optimise the output of the png script.

clobrano:
As @ubuntujaggers said, better integrate the tool than regenerate all the pngs this way

PR #1926, opened 2020-02-16
Added mimetype icon for firmware

madsrh:
I'm going to push this one again, because I noticed that it also exists in Adwaita. We're still missing the symbolic icon, but perhaps we can slim-down the upstream version? 🤷‍♂ 

image

Adwaita:
image


Feichtmeier:
Nice!
I think the symbolic icons should be made in a separate PR to reduce conflicts
I think the new dconf icon also misses a symbolic

friendlyapple:
@madsrh Sorry for the delay. I think @ubuntujaggers had changed the style of this icon in a previous PR.

See: https://github.com/ubuntu/yaru/pull/1683#issuecomment-569956857

PR #1927, opened 2020-02-16
WIP: Added fullcolor dialog-warning icon to status

madsrh:
WIP! This still needs some fine tuning. Could you work your magic @ubuntujaggers ?

Screenshot from Evolution:
image


Feichtmeier:
I could take care of this I think :thinking: 
But we use baseplates for status icons, so if you are okay with a square or a circle, I could do it @madsrh

madsrh:
Oh, you don't want to keep the triangle? ⚠️

Feichtmeier:
No, we don't use triangles for status :) The other danger icon with the "!" also sits on a circle :) (consistency police :policeman: )

madsrh:
Okay, then it's easy copy and paste 👍 Feel free to fix it then.

But I think we should have a triangle shape anyway

Feichtmeier:
Okay, if you think about adding a triangle baseplate, let's wait for @ubuntujaggers opinion and ideas on this

clobrano:
Triangle and Circle have a different meaning IRL, circle is mandatory, triangle is warning. This is not real life, but it makes sense to keep this difference if possible

madsrh:
> Triangle and Circle have a different meaning IRL, circle is mandatory, triangle is warning. This is not real life, but it makes sense to keep this difference if possible

I agree

Feichtmeier:
The logic is unquestionable. It's just that we don't use triangle baseplates.
If you follow a design pattern and/or guidelines and break it too often, maybe you should think about different guidelines then. We only use none baseplate icons for devices. That's all I'm saying

clobrano:
Let's adapt the guidelines to the logic that makes sense then, if it
doesn't work reject it

Feichtmeier:
Hasn't this big fixed? 😆 
🍵 
Lol we are so lazy currently 🤣

madsrh:
I can't check atm but yes. Did you create a separate issue for the icon? @ubuntujaggers will know 👼

Feichtmeier:
Yes, here it is: https://github.com/ubuntu/yaru/pull/2143 👍 
Thanks @ubuntujaggers

PR #1931, opened 2020-02-16
Mimetype icons for Krita, desktop, makefile + symlinks

madsrh:
image

Desktop and makefile are borrowed from Yaru++

Krita did render with an error (because of the gradient?), but images are fine.


Feichtmeier:
Nice!
Krita is very colorful, but so is their brand, right?

madsrh:
> Krita is very colorful, but so is their brand, right?

Indeed. I've just used the gradients [from the icon](https://user-images.githubusercontent.com/3986894/74662537-67697780-519a-11ea-9973-6a39c06ef1f3.png) and added them to the background

Feichtmeier:
If you successfully tested them, please squash and merge :)

PR #1932, opened 2020-02-16
Symlink for selection-checked

madsrh:
Adding a symlink should fix https://github.com/ubuntu/yaru/issues/1686 ✔️ 

Could you test @Feichtmeier ?

### DIFF
--- a/icons/Suru/16x16/actions/selection-checked.png
+++ b/icons/Suru/16x16/actions/selection-checked.png
@@ -1,0 +1,1 @@
+dialog-apply.png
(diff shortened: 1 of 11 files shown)


Feichtmeier:
Oh, awesome!
Need to get that dialogue again. It only came because I was having the partial upgrade because I am using the daily ISO on my host. I think it will come again soon. I'll let you know, or maybe we'll find another app. 
How did you find the icon name? :D

madsrh:
> How did you find the icon name? :D

Yaru++ again, they got a ton of stuff we're missing. Sadly mostly monochrome

Feichtmeier:
As discussed on telegram, sadly does not fix the issue but it's questionable if 18.04 or 19.10 users will have a version of yaru installed where this icon is in.
So let's just merge it? One more symlink can't hurt =)

PR #1946, opened 2020-02-19
Correct number of project parts

Dragon8oy:
Pretty self explanatory, unless I've misinterpreted it


clobrano:
You are right, thanks a lot! And welcome for your first contribution :+1:

PR #1952, opened 2020-02-20
DRAFT: Icons spring clean: round icons

ubuntujaggers:
This branch contains the polished round icons.  When I've got all the icons in master I'll make any necessary changes to the templates and readme and push those (until they're in master we can point any would-be contributors to the draft PR where they'll find the current svgs).

The main differences are that the icons use a more structured template (e.g., the highlight is in its own layer so you can easily fiddle the opacity) and the smaller icons use the full canvas, like so:

image

Also I've removed any clipping I noticed.  Removing clipping probably makes it easier for new contributors to reuse elements from the icons because what you see is what you get.

I'd also like us to consider a couple of design changes for inclusion in master.  Firstly, the Transmission grey is a bit more metallic - give it a whirl if you aren't sure, I think it looks much better on the app grid:

image

Secondly, this branch includes one of the new software-properties designs:

image


ubuntujaggers:
Damn, this was meant to be a draft!

madsrh:
Looks great 😃

Issue #1939, opened 2020-02-18
Sync icontheme, gtk2, gtk3, and shell checkmark assets


Feichtmeier:
1. Sync the colors (aubergine, golden focus rings)
2. Sync the checkmark assets

TODO:

- [ ] https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/source-symbolic.svg checkmark icon

- [ ]  https://github.com/ubuntu/yaru/blob/master/gnome-shell/src/checkbox.svg + the others, already started with this in #1746 

- [x] https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-3.20/assets.svg -> render -> real assets -> https://github.com/ubuntu/yaru/issues/1730

- [x] https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-2.0/assets.svg -> https://github.com/ubuntu/yaru/issues/1730

clobrano:
I'm sorry, but I can't fix the shapes of 

- https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/source-symbolic.svg checkmark icon
- https://github.com/ubuntu/yaru/blob/master/gnome-shell/src/checkbox.svg + the others, already started with this in #1746

the other two items are already covered by #1730 and fixed in #1963

Feichtmeier:
I can fix the shell for sure. No idea about the source-symbolic thing. That's prbly better something for @ubuntujaggers

Issue #1730, opened 2020-01-13
gtk2: use aubergine for assets


clobrano:
All gtk2 assets and some variables in gtkrc needs to be changed to adopt the new aubergine second accent color

- [x] wait for #1731

PR #1963, opened 2020-02-22
color update for gtk 3 and gtk 2 assets

clobrano:
Closes: #1730 
Closes: #1939

TODO
- [ ] dark/gtk2.0 are too old, requires deep rework


Feichtmeier:
omg... selfslapping... just wasted two hours of my life doing exactly the same thing... :| How far are you? Maybe I can at least share the work

clobrano:
Ready for review, but I'd  really appreciate some help @ubuntujaggers for the dark/gtk-2.0 assets-external.svg has old text-selection assets

image

the right one (light variant) is

image

Feichtmeier:
@clobrano 
Looks good so far,
menu bullets look like gtk
also the aubergine sliders and checks :+1: 
The only thing missing is that new checkmark here 
image

ubuntujaggers:
Tomorrow I'll take a break from legacy icons and tick these assets off

Feichtmeier:
Ah one thing, I am not sure if we should make those "Border less" OSD bullets and checks for gtk2 at all
Because they are used in weird places we would not call menus:
image

maybe we better make them all purple + white?

Feichtmeier:
omg sorry missclick

clobrano:
Yeah, I am not that good at shaping on inkscape :sweat: 
can we update this in a next PR? :smile:

clobrano:
> Ah one thing, I am not sure if we should make those "Border less" OSD bullets and checks for gtk2 at all
> Because they are used in weird places we would not call menus:
> image
> 
> maybe we better make them all purple + white?

uhm, right, I didn't observed this in the app I was using as reference

Feichtmeier:
Sure just merge if you think we are one step closer to the goal.
No need to block this

clobrano:
which app is that, btw?

Feichtmeier:
Inkscape settings

clobrano:
> Ah one thing, I am not sure if we should make those "Border less" OSD bullets and checks for gtk2 at all
> Because they are used in weird places we would not call menus:
> image
> 
> maybe we better make them all purple + white?

this seems to happen only for radio buttons, I wonder why, but it needs a redesign, because a grey border does not fit very well, nor a pich dark since it doesn't have enough contrast with the bg :(

clobrano:
How does this look?
image

Feichtmeier:
Disabled :smile: 
I would just leave them purple for gtk2

clobrano:
They have the same color as the labels, so everything should seem disabled, but ok

Feichtmeier:
Let's wait for @madsrh , was just my impression, doesn't mean that this is reality

madsrh:
Hey, [this](https://github.com/ubuntu/yaru/pull/1963#issuecomment-592151172) doesn't look great 🤔 but [this](https://github.com/ubuntu/yaru/pull/1963#issuecomment-592458520) looks perfect 😃 

+1 from me

clobrano:
Alright, merging then, thanks!

PR #1967, opened 2020-02-22
IconTheme: add auth-fingerprint to source

### LINKED ISSUE #1108: Missing auth-fingerprint icon
In Adwaita:

auth-fingerprint-symbolic

Feichtmeier:
@ubuntujaggers and/or @madsrh 

I've added the icon to source symbolic but I can't manage to export the SVG. I've only commited the source change, so please fix it :laughing: :pray: 

image

ToDo:

- [x] Correctly export the icon

Closes #1108


ubuntujaggers:
Hi @Feichtmeier,

I pushed the following:

- Tweak to source-symbolic - moving the design 0.5px was the right thing to do so it's sharp, but the empty "canvas" object had moved too - so moved it back to whole pixels
- Exported the individual icon using **cd ~/yaru/icons/src/scalable && ./extract-symbolic-icons.rb auth-fingerprint**

image

Do you have a fingerprint scanner to test?

Feichtmeier:
So only because it is not exactly on the grid the export doesnt work? :D Thank you!!!
I don't have something to test, the name is correct I would just merge it if you don't disagree

PR #1970, opened 2020-02-23
WIP update for legacy icons edit-redo and edit-undo

madsrh:
I'm not sure about these @Feichtmeier 🤔 WDYT?

image


Feichtmeier:
Imho that's totally fine. Thank you!

Feichtmeier:
@madsrh the only thing that could be improved is if you would add back the fold effect :thinking: Would this be possible?

madsrh:
It's in the file. The layer is simply turned off (might need some refitting though)

PR #1975, opened 2020-02-23
Color change for document-export icon

### LINKED ISSUE #1968: Legacy icon for export and import are too similar
It's really hard to tell the two legacy icons for export and import apart. Perhaps we should adopt that purple 😉 

image

The same two icons in Humanity:
image

madsrh:
Fixes #1968 

Is this okay @ubuntujaggers ?

image


Feichtmeier:
His thumb says yes, merge if tested! :)

PR #1978, opened 2020-02-23
Add gaming controller icon

### LINKED ISSUE #1951: Missing icon for game controller
## Actual Behavior

Capture d’écran du 2020-02-20 11-09-09

## Adwaita

Capture d’écran du 2020-02-20 11-04-01

Super Nintendo colors 😋

  1. Open Gnome-settings
  1. Go to Bluetooth
  1. Connect a game controller (I connected a Nintendo Switch Pro controller)

Jupiter007-43:
Closes #1951


Feichtmeier:
Looks very good!
Just one tiny thing - do they need to be in legacy folders? I thought about fullcolor -> devices, no? @ubuntujaggers

Jupiter007-43:
I think yes : https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/src/fullcolor/legacy/audio-speakers.svg

ubuntujaggers:
Gnome does have audio-speakers and input-gaming in legacy folders, but that might be because they're the old style rather than the modern style.

Generally, devices used in current apps (e.g. that show up in Nautilus when they're plugged in) go in fullcolor --> devices.  Legacy folders are for colour designs that show up in older apps (like the Inkscape buttons).  I suggest we put it in both places to be safe?

This is a fantastic job, @Jupiter007-43 - I've got a couple of small tweak requests.  For some reason I get an error if I try to pull a branch that isn't one of ours, so are you able to make the changes?

First of all, the smaller icons have been resized perfectly, but the gradient for the top and bottom edges has been left behind, which makes the highlights invisible.  The highlights object is in the Highlights layer and is positioned here:

image

If you use the node tool on it, you can see that the gradient starts and ends off canvas:

image

Click just under the outline to select this object.  Using the node tool, make it so the gradient starts and ends 1px from the outline (you have to do this for the four small sizes):

image

Also, the controls on 24px look a bit high compared to the other sizes - how about moving the d-pad and four coloured buttons down by 1px on that icon?  Revised svg looks like this:

image

These are minor issues, I think it's fantastic :+1: If you can fix the highlight on sizes 16-48px and copy the icons into the full colour devices folder then I say merge.

Jupiter007-43:
I didn't seen the highlights 😓
I'm still a beginner with Inkscape.

Jupiter007-43:
Capture d’écran du 2020-02-24 11-30-37

ubuntujaggers:
I think that's very smart, should I merge?  @clobrano @Feichtmeier @madsrh?

Feichtmeier:
Better ask @3v1n0 I think he needs to rebase then

PR #1982, opened 2020-02-24
Prepare for ubuntu release 20.04.2

3v1n0:
Let me know if this is fine to release as it is up to commit e402f7b8.

Please don't merge anything before this, or let me know in case so I can rebase.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,95 @@
+yaru-theme (20.04.2) UNRELEASED; urgency=medium
+
+  [ ubuntujaggers ]
+  * trying new icon for dconf editor
+  * tweaking new icon for dconf editor
+  * trying more suru version of media-flash
+  * adding missing shadow to 48px
+  * pixel fitting floppy disk icon
+  * tweaking edging
+  * adding highlights
+  * trying version of network-server without folds
+  * Symlink drive-removable-media-usb -> media-removable (#1892)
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Nothing urgent on my list that would be ready now. Eventually @ubuntujaggers has something in the pipeline?

Feichtmeier:
@3v1n0 one thing... Which version of gnome shell well you put in now? Because the keyboard had a change to use symbolic icons instead of assets. This needs also a change to the theme

ubuntujaggers:
I've got some near future bits for templates, the rest of the spring clean, and legacy buttons - will we get chance to do them after?

3v1n0:
> @3v1n0 one thing... Which version of gnome shell will you put in now? Because the keyboard had a change to use symbolic icons instead of assets. This needs also a change to the theme

It would be 3.35.91, from what I see it seems to work properly with current master right now, but we have time to fix it later if I'm missing something

Feichtmeier:
Alrighty, then!
I need to apply the fix later then, it's nice that they remove more and more assets from shell :+1:

Feichtmeier:
@3v1n0 

Okay we have some changes in master - please rebase!
(And really make sure that your version of gnome shell doesnt have the on screen keyboard asset change it, otherwise I need to make a change to the theme)

PR #1983, opened 2020-02-24
Revert to upstream tabs but make them less heavy

Feichtmeier:
I think this "looks" a bit better - I don't have anything else to say about this. No usability argument or something else. Just the look :man_shrugging:

Before:
image
(hover above "Papierkorb" tab :) )
image

After:
image
(hover above "Papierkorb" tab :) )
image

Edit: some more pics 
image
image

@clobrano @madsrh @ubuntujaggers 

(maybe a little plus is much less code)
image


gtk/src/default/gtk-3.20/_tweaks.scss
```diff
@@ -316,343 +316,27 @@ treeview {
...
-
-          &:checked {
-            background-color: $base_color;
-            border: 1px solid $borders_color;
-            border-bottom-color: transparent;
-            box-shadow: none;
-
-            &:hover { background-color: $base_color; }
-            &:backdrop { background-color: $backdrop_base_color; }
+// Make the tabs slightly less heavy
```
clobrano: We can remove or rephrase the comment, I think, since it is not really clear ("less heavy" respect what?)

Feichtmeier: Indeed, sorry for the weird comment. I'll update it if you guys like the idea

clobrano: I still need to test the branch. I am positive, but since there's a poll, let's wait for the other replies

ubuntujaggers:
Just tried it - I like both!  More than happy to switch to this if people prefer it, I suppose it's a bit sleeker/more modern.

gtk/src/default/gtk-3.20/_tweaks.scss
Feichtmeier: What comment would you suggest btw? :smile_cat:

madsrh:
It's a +0 from me 😄 I like both 🙈

gtk/src/default/gtk-3.20/_tweaks.scss
clobrano: None? :D

clobrano:
Can we try to use the same color for both the label and the underline on hover? Both light and dark

image

clobrano:
also, remember to delete undershot, we don't need it anymore if this PR goes to master

clobrano:
Use the same headerbar color in nautilus

image

Feichtmeier:
> Can we try to use the same color for both the label and the underline on hover? Both light and dark
> 
> image

Good idea!
Changing the hover inset box-shadow would mean to re-insert the 300 lines again :S Or at least some of it.
Stopping the hover color change for not(:checked) is easy though
image
For the dark theme it's still different though. Because they use $borders_color for the box-shadow :face_with_head_bandage: 
image

Peek 2020-02-24 23-08

Feichtmeier:
> Use the same headerbar color in nautilus
> 
> image

My idea was the transparent bg, so the background of the widget the notebook is put into is taken. 

If the notebook is placed on $base_color it's that. If it's $bg_color it's this. And so on.

clobrano:
The problem is that on Nautilus the separation between notebook and
headerbar is not visible and misleading

clobrano:
> if the notebook is placed on $base_color it's that. If it's $bg_color it's this. And so on

What if we change the background color behind the headerbar only in Nautilus then?

Feichtmeier:
Sounds like a plan, I'll try this

Feichtmeier:
image

image

image

Like this @clobrano ?

clobrano:
As said, I don't like much the dark variant hover, but it's a minor issue, so ok for me

madsrh:
> I don't like much the dark variant hover

+1

I had the same thought

Feichtmeier:
I try to improve it without re-inserting so much code in a second PR :+1:

PR #1985, opened 2020-02-24
Adding legacy humanity actions to Yaru

ubuntujaggers:
This change protects buttons in older apps from being overwritten by a similarly named (and sometimes wrong) Yaru icon.  Inkscape now looks like:

image

This is the safest and most comprehensive version of the change: I've taken all the full colour actions from Humanity.  The more conservative approach would be to only add Humanity icons as and when we know they're prone to getting "bumped" by Yaru ones.  Happy to do that instead, although this approach protects us from problems we don't know about yet.


Amr-Ibra:
Does that mean that humany-icon-theme does not need to be installed by default in Ubuntu?
https://launchpad.net/ubuntu/+source/humanity-icon-theme

I'm asking to decrease duplication on disk space if possible.

Feichtmeier:
@ubuntujaggers one thing that I am not okay with: 

please keep the new checkmark icons because they show up in the installation process of Ubuntu ✅

Feichtmeier:
Can't we achieve the same thing with using Inherit=Humanity in the index.theme? @ubuntujaggers

ubuntujaggers:
> @ubuntujaggers one thing that I am not okay with:
> 
> please keep the new checkmark icons because they show up in the installation process of Ubuntu

Yep, very reasonable  :)

I'm also happy to go with the approach where we only add Humanity icons as and when we know there's a problem, no strong feelings.

clobrano:
> Can't we achieve the same thing with using Inherit=Humanity in the index.theme? @ubuntujaggers

Isn't this as we already do?

Feichtmeier:
Okay. But there must be a better way than to copy humanity icons that are already installed in 20.04 nonetheless. No offense, just fearing we do the wrong thing

Amr-Ibra:
And this should be kept in mind that legacy icons in Yaru are only "needed" because of Qt apps. GTK 2 and 3 are happy without them.
If only Qt apps could choose the right icons for document-*!

Issue #1832, opened 2020-01-31
Shell popover styling


madsrh:
Here's a wishlist @Feichtmeier. Let's see if any of them are fixable.

1) **Less padding**
Actually, the line height isn't much different between GTK and the shell, but the padding around the separators are huge. 
![image](https://user-images.githubusercontent.com/3986894/73555408-dbdbb100-444d-11ea-8df7-c408c0b8312c.png)

2) **Background is different**
This might be deliberately but the background color is #fff in the GTK context menu and #f7f7f7 in the shell menus 🤷‍♂ 

3) **Shell does not respect font change**
Not sure if this is even fixable, but it would be great if the two were linked. Perhaps we could tag didrocks ?
![image](https://user-images.githubusercontent.com/3986894/73555005-1db82780-444d-11ea-86cb-b5cddf02dac8.png)

Feichtmeier:
2) is easy and I'm sure it's fixed in the WIP PR if not three secs

3) is luckily fixed with the new shell version 

1) could be difficult to not break anything but I'll check

Feichtmeier:
@madsrh concerning the colour, which yaru version did you try?
The popovers are f7f7f7 in master
https://github.com/ubuntu/yaru/blob/9c5f466d866eaf25abfdb2a832ff6291118a7677/gnome-shell/src/gnome-shell-sass/_colors.scss#L6
->
https://github.com/ubuntu/yaru/blob/9c5f466d866eaf25abfdb2a832ff6291118a7677/gnome-shell/src/gnome-shell-sass/_common.scss#L600

madsrh:
Oh, I don't recall 🤔I'll check later.

Do you prefer the f7f7f7 or white?

Feichtmeier:
f7f7f7, like in gtk

madsrh:
👍 As long as they use the same color I'm happy 😀

madsrh:
@Feichtmeier 

![image](https://user-images.githubusercontent.com/3986894/75099542-2cfd4180-55c3-11ea-9237-770d7ae38a22.png)

Feichtmeier:
Master?

madsrh:
> Master?

Yes, apprentices?

Sorry, couldn't resist. Yes, it's in master

Feichtmeier:
Well then it's a problem with how the shell renders the colors, have a look yourself:

[...]

PR #1987, opened 2020-02-25
colors: use brighter background for hover elements

clobrano:
All popover in both gtk and shell use bg_color (#F7F7F7), while the
context menu is using base_color (#FFFFFF).

Use bg_color for context menu as well (also has better contrast over
Nautilus notebook wich is pure white

see: #1832


madsrh:
🙈 It's actually the (#F7F7F7) I don't like. Better (#FAFAFA) everything or as @Feichtmeier just suggested [here](https://github.com/ubuntu/yaru/issues/1832#issuecomment-590813601), white. What do you think?

clobrano:
Let's try with FAFAFA

clobrano:
With the new organization of the shell files, the background is based on bg_color (instead that on osd colors for example), so using FAFAFA we're introducing a new color not belonging to the palette. I'm not against, I just want you to know :)

clobrano:
wait, there is a problem with the dark variant

clobrano:
@madsrh have a look :wink:

madsrh:
+1 I think this is good. Are you also +1 @Feichtmeier / @friendlyapple?

Feichtmeier:
I need to test it - it's for both shell and gtk, right?

madsrh:
okay, **EDIT:** I tried this before "Merge branch 'master' into sync-context-menu-color-with-popover" but now it looks like the shell is F7F7F7 again 🤷‍♂

Feichtmeier:
Gtk looks good I think - but I am still not sure if the menu really needs to be #fafafa as well :thinking: 
@clobrano I think you need to change this here
https://github.com/ubuntu/yaru/blob/sync-context-menu-color-with-popover/gnome-shell/src/gnome-shell-sass/widgets/_popovers.scss#L8
to ```#fafafa```

madsrh:
Doh, I forgot to restart the shell 🤦‍♂ 


image

Feichtmeier:
It's still f7f7f7 in the shell @madsrh - you can even see it in your sceenshot :D 
We should be able to use lighten for it just to avoid introducing another hex in there:
image

madsrh:
> you can even see it in your sceenshot :D

I think it's an optical illusion 👁👁 I've used the picker and it checks out. Below I've cut out a piece of the shell and moved it over the popover 🤷‍♂ It's the same background

image

Feichtmeier:
Oh... lol! I am an idiot again :D 

So carlo changed the whole bg_color for the shell :man_shrugging: I don't think that this makes much of a different, since there is not much more visible in the shell than this

I think this looks better overall. The contrast between the calendar and the cards on top if it is enough
The only thing that could not provide from this is the on screen keyboard and buttons, because the hover effect is already almost invisible. This reduces the hover difference for elements that use button(normal) -> button(hover) even more
If you are okay with this I am, too. Otherwise maybe better change the color in _popovers.scss

clobrano:
> Otherwise maybe better change the color in _popovers.scss

I prefer to change bg_color, since in popover everything is related to this color, so the change this way is minimal

Feichtmeier:
Okay, I think the contrast is still fine for the calendar:

```#f7f7f7```
Screenshot from 2020-02-27 15-02-51
```#fafafa```
Screenshot from 2020-02-27 15-04-25

But the hover effect for the keyboard buttons is basically gone: 
Screenshot from 2020-02-27 15-08-03

We could fix this later then, explicitly for the keyboard. So no matter how you do it, there is diff :smile:

Feichtmeier:
Okay, I think the contrast is still fine for the calendar:

```#f7f7f7```
Screenshot from 2020-02-27 15-02-51
```#fafafa```
Screenshot from 2020-02-27 15-04-25

But the hover effect for the keyboard buttons is basically gone: 
Screenshot from 2020-02-27 15-08-03

We could fix this later then, explicitly for the keyboard. So no matter how you do it, there is diff :smile:

Feichtmeier:
Yeah me too! 
Just wanted to remind (me at least) to check the keyboard hover :+1:

clobrano:
> Otherwise maybe better change the color in _popovers.scss

I prefer to change bg_color, since in popover everything is related to this color, so the change this way is minimal

PR #1990, opened 2020-02-25
Update check radio assets

Feichtmeier:
- slim down the SVGs
- export the PNGs

Before:
Screenshot from 2020-02-25 13-19-54
After:
image_2020-02-25_19-19-25


friendlyapple:
Much better :) However, is it me or the dash is slightly blurred?

Feichtmeier:
😱 I hope not... @ubuntujaggers could you investigate if I did it correct in Inkscape?

fitojb:
Some gridfitting tips to avoid blurry exports: https://graphicdesign.stackexchange.com/questions/63403/snap-to-grid-in-inkscape

PR #1994, opened 2020-02-26
shell: re-use silver color for mouse selection

clobrano:
re-apply #1883 to the shell, since it was lost after upstream rebase

Closes: #1727


madsrh:
Wow, that was fast! I've tested it and it works 🎉 

image

PR #1997, opened 2020-02-26
Shell theme upstream fixes

Feichtmeier:
1) This includes new changes from https://github.com/ubuntu/yaru/pull/1955
2) Using a ppa from @3v1n0 I was able to fix things for a more recent shell version:

- add fixes from upstream for calendar, login screen, lock screen, dialogs, looking-glass, panel, misc and more
- add eye assets from upstream for the login entry, even if it is not loaded from within the theme -> reduces diff
- reduce our diff in drawing, but keep our style for focus rings and button shadows like in our gtk theme
- use an orange dot for the running dots in the app grid
- use $inkstone fg $porcelain bg buttons in the login screen in both variants, because user themes cant change it
- only change font-weight in our little "tweaks" section in _widgets.scss and not in _panel or _calendar or _dialogues to reduce the diff
- re-apply our style for the window-picker (no diff here)

Visible changes are, the login and lock-screen, the fonts no longer streteched, the system logout menu looking like upstream intended it to be:
image_2020-02-26_20-35-43
image_2020-02-26_19-32-57

@3v1n0 please test


3v1n0:
Works great, do we want the white background (when focused) here though?
I find the contrast a bit strong, given we've an orange border too.

immagine

Or maybe, the wait thing is fine, I don't like much the border though, even in the search (where maybe having it purple makes more sense, but I see that this way it maches gtk:
immagine

As per the entry is a bit centered on the right side, but this could be an upstream problem too.

Feichtmeier:
Thanks for the review!
The search entry has only been touched concerning hover, because they (upstream) failed to make it work in both variants (light/dark).

The login/lock screen entry... I forward this to @madsrh and @clobrano. I like the white but I could also be okay with a dark one

Feichtmeier:
Okay Carlo agreed to merge.
I'll propose different login and search entry designs in a separate PR

PR #2003, opened 2020-02-27
Adapt gtk to shell spinner

### LINKED ISSUE #1938: Process-working spinners across the desktop - from the boot to shell to the application
Not sure if we want this, but we can investigate .

[Vanilla style](https://github.com/canonical-web-and-design/design-vanilla-framework/blob/master/Icons/icon-set.md) GTK spinner NOT the wait mouse cursor (for now)

Feichtmeier:
We can not change the shell spinner in Yaru because it is not loaded from within the shell theme and not from the icon theme, it is a gresource we can not access from yaru.

So this...

Reverts process-working to upstream …

- the upstream spinner is used in the shell and in the boot screen (for new PCs) and both are not changeable in our project
- reverting to upstream in color and form reduces the inconsistency

Bildschirmfoto von 2020-02-27 23-49-24

The shell spinner still does not look _exactly_ like the gtk spinner (should ask upstream) but it is damn close and it looks also exactly like the boot screen spinner you see when you have a vendor boot screen. 

The only thing left would be to update the mouse cursor if we want but the three slided spinner looks cool on a cursor and it is a different thing imho

At least the inconsistency is reduced.
Opinions @clobrano @madsrh @ubuntujaggers (thanks Stuart for the help)


madsrh:
If we keep the wait cursor, I'm good :smile: Nice work @Feichtmeier

Feichtmeier:
Leave it as a draft - there is a new chance that the shell spinner becomes our beloved Yaru spinner 
#1938

PR #2015, opened 2020-03-01
Gtk3: Make all :not:(.empty) level bars green

### LINKED ISSUE #2014: Battery levels have the wrong color
Battery levels should not use aubergine 🔋 

image

Feichtmeier:
Closes #2014 

Before:
image

After:
image


popey:
Looks great! (before and after)
Screenshot from 2020-03-01 21-28-38
Screenshot from 2020-03-01 21-23-00

Feichtmeier:
I think for our problem having so many different colors this is fixed.
However there is def. something wrong with gnome control center there.
Because it doesn't always use the correct style classes for .full, .high and .empty
Found an upstream bug for this https://gitlab.gnome.org/GNOME/gnome-control-center/issues/725

We could still see this as fixed for our case though

Issue #2016, opened 2020-03-02
disabled switch has the same style as non-disabled


clobrano:
In default and light variants, the disabled switch has the same style as the non-disabled one

![switch-disabled-before](https://user-images.githubusercontent.com/2883614/75659227-eb773100-5c69-11ea-8c10-55ad580db828.gif)

PR #2017, opened 2020-03-02
switch: stronger disabled state in default and light variants

clobrano:
In default and light variant, the disabled checked switch is the same as non-disabled one. In the dark variant, the only difference is the handle.

This is due to the need to move our switch style to tweaks to reduce upstream diff in common, so the fix consists in porting some missing styling from common to tweaks.

Closes: #2016

switch-disabled-after

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -227,33 +227,30 @@ switch {
 
   &:checked {
     background-color: $switch_bg_color;
+    &, & slider { border-color: $switch_border_color; }
 
-    &, slider {
-      border-color: $switch_border_color;
-    }
-    &:disabled slider { @include button(insensitive); }
     &:disabled {
       color: $insensitive_fg_color;
       border-color: $borders_color;
       background-color: $insensitive_bg_color;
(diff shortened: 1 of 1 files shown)


clobrano:
This change uses the "semantically correct" insensitive colors, but I remember that in the past we differentiated the disabled checked from disabled non-checked. I could update the PR with a similar change if you prefer

Feichtmeier:
Oh shit, I guess I removed to much in that "clean up" - Thanks :+1:

Feichtmeier:
> This change uses the "semantically correct" insensitive colors, but I remember that in the past we differentiated the disabled checked from disabled non-checked. I could update the PR with a similar change if you prefer

i think the way you did it here is fine
you see if it's checked at the position of the knob/slider

imo you can merge it :+1:

philipzae:
Hope the fix works as intended because the bug i saw affected popey as well with his dark theme.
https://twitter.com/popey/status/1234212400255074304

clobrano:
The fix is for all variants

switch-disabled-after-dark

The picture posted on twitter showed that there was a change in dark variant for disabled switch (the handle). Of course it wasn't enough and I made all the variants act the same

PR #2018, opened 2020-03-02
ShellTheme: color app grid rename dialog label

Feichtmeier:
So it stays variant independent, should be moved to upstream since they have forgotten this

Before:
Screenshot from 2020-03-02 08-37-28

After:
Screenshot from 2020-03-02 08-40-08
Screenshot from 2020-03-02 08-43-06


Feichtmeier:
Upstream MR: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1059

Issue #2012, opened 2020-03-01
Symbols differ in aspect ratio for sidebar vs. folder icon


friendlyapple:
I want to reopen an old issue in which @matthewpaulthomas said that the symbols were different between the sidebar and the folder icons.

The problem is still here. Moreover, it concerns all folders and not only the download folder.

Original issue: https://github.com/ubuntu/yaru/issues/865

Feichtmeier:
Fixed with #2052

PR #2024, opened 2020-03-02
Changed symbolic download icon to better match folder download

madsrh:
This fixes (only) the download icon in #2012 - which also is the most eye catching one.

image

image


ubuntujaggers:
Can we try a halfway house, making the hollow part of the stalk 3px instead of two? In that case, to ensure the image is sharp and aligned with the grid, you would then have to make the horizontal line across the bottom only 15px, and have the whole design either to the left or right of the canvas - whichever looks more balanced when the icon is in use (lots of symbols actually only use 15 of the px to allow an odd-numbered element to be cleanly centred).

ubuntujaggers:
Oh wait... was it already like that ☺️

friendlyapple:
Wouldn't it be better to leave the sidebar icon and change the folder icon? The download icon looks "compressed"

clobrano:
Indeed, the folder glyph seems a bit out of proportion, like too tall
respect the wide, but I'm basically ok with it, if you prefer to keep this
style.

ubuntujaggers:
@madsrh , maybe:

image

image

??

Tweaks:

- arrow stalk is 5px across (grey, white, white, white, grey) and the top of the stalk is lowered by 1px
- horizontal line is made to match the new arrow
- whole image is slightly to the right;

image

madsrh:
@ubuntujaggers This looks good 👍 I'm still AFK, but could you make a pull request for that (since you've already made it 😄)?

Feichtmeier:
Fixed by Stuart in the other pr

PR #2025, opened 2020-03-03
GNOME Shell generate gresource

### LINKED ISSUE #1938: Process-working spinners across the desktop - from the boot to shell to the application

3v1n0:
Lots of refactories in meson for gnome-shell, reducing duplications and hardcoded strings.

Add a meson option to generate gresource file for the Yaru gnome-shell theme
as it is upstream so that we can inject this into the shell and be able to
theme even more assets (this requires some light patching to the shell, not yet included in ubuntu or upstream).

 - The *.scss files use now again resource:// path by default to reduce
   diff with upstream
 - In case we don't use gresources, such files are reconfigured to use path
   code has to be hacky if we don't want to repeat the same for each folder
 - In case gresources is generated we install it and set in the yaru.json
   but no asset is installed in /usr/share/gnome-shell, but we do keep them
   in /usr/share/themes/gnome-shell for compatibility with extensions

Closes #1938


clobrano:
Awesome work @3v1no, thanks!

Feichtmeier:
Have this output after this merge when launching gtk3-widget-factory:

```bash
frederik@phoenix:~$ gtk3-widget-factory 

(gtk3-widget-factory:31234): Gtk-WARNING **: 10:58:29.650: GModule (/usr/lib/x86_64-linux-gnu/gtk-3.0/3.0.0/immodules/im-ibus.so) initialization check failed: GLib version too old (micro mismatch)

(gtk3-widget-factory:31234): Gtk-WARNING **: 10:58:29.650: Loading IM context type 'ibus' failed
```

Feichtmeier:
@3v1n0 do we need a newer shell than 3.35.91 for this?
Ths top panel and login screen spinner is still the upstream one here:
Bildschirmfoto von 2020-03-04 11-21-27

3v1n0:
@Feichtmeier yeah, we'll release this as part of 3.35.92, as right now it would actually break it. :)

Feichtmeier:
ok!

PR #2029, opened 2020-03-03
Fix headerbar button regression caused by #2010

### LINKED ISSUE #2028: gnome-software stackswitcher central button has odd white border
image

Feichtmeier:
Remove this from tweaks as it introduced bugs. The only riskfree way of changing the headerbar outside of _common is to fully re-style it, like we did in _headerbar.scss for the default theme. Asking upstream to fix this uneven borders so the diff is gone again.

Bildschirmfoto von 2020-03-03 20-25-24

Closes #2028


Feichtmeier:
@clobrano it turned out that the whole block I've made was crap. If you are okay with this I'd rather take these two lines of diff and take personal care to get this into upstream later

PR #2030, opened 2020-03-03
Adding colour icons for headphones and headset

### LINKED ISSUE #1930: Full color headset icon is missing
image

ubuntujaggers:
I have no idea if this will look any good in context - would someone with bluetooth headphones mind testing?  It shouldn't be this saturated in reality, the images are compressed here:

image

image

It's extremely hard to come up with a Yaru-styled image that works at the small sizes too  :(

Closes #1930


madsrh:
I'm really not a fan of the colored backplate. I haven't got a better suggestion, but just wanted to mention it.

Feichtmeier:
Bildschirmfoto von 2020-03-03 21-25-14
I disagree - I like it, especially when you see it in the app. Only the colors are a bit too strong as discussed

Feichtmeier:
Bildschirmfoto von 2020-03-03 21-31-24
Bildschirmfoto von 2020-03-03 21-33-39

I like it ^^

clobrano:
What might be against the colored backplate is that on a light background it kinda stole the attention from the actual icon, but as Mads I didn't have a better option, and this is anyway a good one, so :+1:

Issue #2031, opened 2020-03-03
Gtk3: scale troughs could need a border


Feichtmeier:
Adwaita-dark:
![image](https://user-images.githubusercontent.com/15329494/75828761-e1aa1680-5dac-11ea-881b-7d1dd56f1c79.png)

Yaru-dark:
![image](https://user-images.githubusercontent.com/15329494/75828798-f71f4080-5dac-11ea-9277-09d03b367385.png)

Adwaita:
![image](https://user-images.githubusercontent.com/15329494/75829042-7ca2f080-5dad-11ea-9e9b-c896e6606648.png)

Yaru:
![image](https://user-images.githubusercontent.com/15329494/75829083-93e1de00-5dad-11ea-833c-e1cb55cd21d4.png)

In my opinion, the border-less filled scale does not work in gtk. 
I would suggest to align to upstream here
https://github.com/ubuntu/yaru/blob/96db78eb3aafaca39dc7e21c350289e593ad64c4/gtk/src/default/gtk-3.20/_colors.scss#L76
https://github.com/ubuntu/yaru/blob/96db78eb3aafaca39dc7e21c350289e593ad64c4/gtk/upstream/gtk%2B3.0/Adwaita/_colors.scss#L75
->
https://github.com/ubuntu/yaru/blob/96db78eb3aafaca39dc7e21c350289e593ad64c4/gtk/upstream/gtk%2B3.0/Adwaita/_colors.scss#L12

PR #2039, opened 2020-03-05
Gtk3: darken progressbar border colors

Feichtmeier:
when looking at the unfilled progressbars with a border versus the part of the progressbar that is filled, it looks like the progressbar is 1px bigger when filled, which is not exactly clean.
Darkening the progressbars slightly in the dark and light themes fixes this and also increases the contrast to the background, no matter if the progressbar sits on $bg_color or $base_color

Screenshot from 2020-03-05 09-20-35
Screenshot from 2020-03-05 09-20-29

Closes https://github.com/ubuntu/yaru/issues/2031


clobrano:
The light variant looks a bit too heavy, don't you think?

Feichtmeier:
okay I try less there

Feichtmeier:
Screenshot from 2020-03-05 14-39-14
This okay @clobrano ?

clobrano:
I had to see it in a VM, this way looks good :+1:

PR #2033, opened 2020-03-04
headerbar: custom outline for suggested-action button

clobrano:
headerbar.scss is parsed after tweaks.scss, removing the outline definition. The less risky action is to re-set it in headerbar

Closes: #2019

PR #2040, opened 2020-03-05
headerbar: fix missing states in suggested button outline

clobrano:
Fix headerbar suggested button outline's states styling missing from #2033


Feichtmeier:
tested and works, thanks!

Issue #1989, opened 2020-02-25
Background in top of Templates folder


madsrh:
I know that per app tweaks isn't ideal, but IMHO the label in the Templates folder lacks discoverability which is key for it's usefulness. Opposed to the Trash folder, the top banner here acts as information and would benefit from a background.

WDYT?

![image](https://user-images.githubusercontent.com/3986894/75255230-e20d4500-57e1-11ea-96b2-f75ed6ef76bc.png)

madsrh:
Okay I was pretty alone on this one 🙈

friendlyapple:
@madsrh I think it's a good idea :)

madsrh:
👍 @friendlyapple 😁

I tried this to do this with `infobar.info` but couldn't make it work. Is this even possible @Feichtmeier ?

clobrano:
It is
![image](https://user-images.githubusercontent.com/2883614/75606199-ab3d7480-5aea-11ea-85fe-386537257662.png)

clobrano:
Ah, "unfortunately" also the Thrash folder has an infobar
![image](https://user-images.githubusercontent.com/2883614/75606317-b644d480-5aeb-11ea-83f9-298b41bd41af.png)

Feichtmeier:
Haha infobars again xD 
I think that yellow looks like that 3M post-its so I think it looks good for a info :+1: 
But what I don't like is that upstream just ignores the differences between the info bars: warning, info, error and so on

madsrh:
> "unfortunately" also the Thrash folder has an infobar

Why is unfortunately inside a " " 😂

> Shall we go with your suggestion instead (#f3ef82)?

Sure or even lighter perhaps?

> upstream just ignores the differences between the info bars: warning, info, error and so on

So, you're both saying that it isn't possible?

Feichtmeier:
No we had it once... back in time... I think... let me see...
https://github.com/ubuntu/yaru/issues/856#issuecomment-423826338

clobrano:
It totally makes sense. We went through a plethora of ideas :D
Right now, upstream does not make any difference, however yellow might suit better to a warning message, even more in dark variant, where the yellow should be darker.

What do you think of this?
![nautilus-infobar](https://user-images.githubusercontent.com/2883614/75681668-5ccbd980-5c94-11ea-80eb-33e07a155d63.gif)

Feichtmeier:
That looks def. better for an information
[...]

PR #2041, opened 2020-03-05
apps/nautilus: make infobar more visible

clobrano:
Closes: #1989

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -26,6 +26,8 @@ $ambiance: null !default;
     @at-root %dim_icons_in_backdrop,
     & *scrolledwindow:backdrop { opacity: 0.9; }
 
+    infobar { background-color: if($variant=="light", lighten($silk, 10%), $slate); }
+
     paned box.floating-bar {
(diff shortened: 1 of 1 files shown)


Feichtmeier:
What I don't understand is, that I believe you want the users to actually see this better, but isn't a yellow color more visible than a gray color? :thinking:

clobrano:
My point is to make visible that there is a label there, so just a different color is enough I think. Yellow is a warning color, I wouldn't use it

Feichtmeier:
I don't have strong feelings on this and can live with both
Proceed as you wish :+1:

clobrano:
Okay, I'll wait for @madsrh feedback, since he requested the change in the first place

madsrh:
> For the trash I think it is a little bit more "dangerous" and the yellow fits, imo.

That's funny - IMO the yellow doesn't fit in Trash 😄 because it acts more like a header, where in Templates in highlights the actual information the users should read 🤷‍♂ 

What I was thinking (but failed to illustrate in the initial mockup) was something subtle like #fff4c9

image

It's pretty annoying that we can't target only Templates and not Trash, so gray might work better for both. That was a long way of saying +1 for this PR

Issue #2034, opened 2020-03-04
shell: missing border in dark shell slider


clobrano:
Is this a regression or an upstream change?

![image](https://user-images.githubusercontent.com/2883614/75859893-2a88bc00-5dfb-11ea-834e-732ea40e86e2.png)

Feichtmeier:
Uh that's weird, last time I checked the dark shell theme was yesterday from master and it was not like this. I do like the white slider but I know others don't. So I guess this is a regression

Feichtmeier:
https://github.com/ubuntu/yaru/blob/f5452e62ef1da89d9345f00e1699750bae64a46b/gnome-shell/src/gnome-shell-sass/widgets/_slider.scss#L22-L24

```$fg_color``` is ```f7f7f7``` for the dark theme

clobrano:
Yes, the handle is supposed to be dark, probably, but I was referring to the slider border. At the sprint, or soon after, we decided to add a border because aubergine blends too much with the dark panel

Feichtmeier:
Yes sorry. That's what weird :) I can fix both of you haven't already started

clobrano:
I started no worries :)

Feichtmeier:
if you like those: https://github.com/ubuntu/yaru/pull/2039
could you also adapt the light variant to that branch's look=?

PR #2042, opened 2020-03-05
Issue2034/shell slider borders

clobrano:
shell/slider: revert upstream change to yaru style

- changed border color to pre-rebase since current one is too strong against panel background
- use new progress_border_color from gtk and adapt the overdriver border color to it.
- on the slider, don't use the same color for background and border

Closes: #2034

image
image

Note: this might be a little tricky to test on dark shell because the new meson build does not symlink anymore Yaru-dark shell. I've already asked @3v1n0 if there is a reason not to symlink, and if not, the fix should be easy. Anyway if you don't see any difference in the dark shell, try to double check the symlink and do it manually eventually.


clobrano:
updated the handle's borders as in gtk

Feichtmeier:
Looks good from your pictures, but I can't test it locally since I don't know how to set the $variant to dark manually :) any tips?

clobrano:
Can't you install user themes?

Feichtmeier:
Alright I'll try later or tomorrow, if you successfully tested it yourself (which I guess, since there are screenshots xD) you could also just merge it

Issue #2043, opened 2020-03-06
Dark shell theme no longer selectable in gnome-tweaks


chrisjbillington:
A few commits ago, the dark shell theme was installed to `/usr/share/themes/Yaru-dark/gnome-shell`, and now it has moved to `/usr/share/gnome-shell/themes/Yaru-dark`.

I see that this brings it in line with the light shell theme, which lives in `/usr/share/gnome-shell/themes/Yaru`, but is symlinked to from `/usr/share/themes/Yaru/gnome-shell`.

However, The dark shell theme is no longer selectable in gnome-tweaks (at least on gnome 3.34).

Should it get a symlink from `/usr/share/themes/Yaru-dark/gnome-shell` like the light theme has? Or are things different in gnome 3.36 such that those symlinks are not necessary for the themes to appear in gnome-tweaks (in which case the light theme's symlink can be removed)?

clobrano:
Hi @chrisjbillington,

I noticed this yesterday, the fix should be easy, I'm jus waiting for a feedback from desktop team

fra98:
The Yaru-dark entry disappeared even with the Yaru release from January (19.10.5), so it doesn't seem to be related to the latest changes, any idea?

chrisjbillington:
@fra98 for me I saw it in gnome-tweaks until a few commits ago and can pinpoint the cause as the move of the dark shell theme out of `/usr/share/themes`. However, I'm building from source (well, in the context of an Arch Linux package).

It's possible that for e.g. the Debian packages something could be packaged wrong leading to a similar issue - what OS and packages are you using?

@clobrano thanks! Feel free to close my PR, and if you @ me in the changes I'll test them.

PR #2044, opened 2020-03-06
Add symlink to dark shell theme in /usr/share/themes

chrisjbillington:
Fixes issue #2043 

The dark shell theme was moved from `/usr/share/themes/Yaru-dark/gnome-shell` to
 `/usr/share/gnome-shell/themes/Yaru-dark`.

As with the light shell theme, a symlink should be left in its old location so that it is visible in gnome-tweaks.

### DIFF
--- a/gnome-shell/src/install-shell.sh
+++ b/gnome-shell/src/install-shell.sh
@@ -7,3 +7,4 @@ install_prefix="${MESON_INSTALL_PREFIX}/share"
 
 mkdir -p "${destdir_prefix}/themes/${project_name}"
 ln -sf "${install_prefix}/gnome-shell/theme/${project_name}" "${destdir_prefix}/themes/${project_name}/gnome-shell"
+ln -sf "${install_prefix}/gnome-shell/theme/${project_name}-dark" "${destdir_prefix}/themes/${project_name}-dark/gnome-shell"


Feichtmeier:
Thanks for the PR
I don't remember what's the deal but I'm sure @3v1n0 had some reason for it. Let's please wait for his reply

clobrano:
Hi Chris, I've already contacted @3vih0, and I have a patch to the
meson.build ready, which I think it's more appropriate to fix this.

3v1n0:
Mh, couldn't be that I just maybe generate the install dir with a wrong path?

Maybe this can be fixed with meson only changes?
As if we use gresources we can just install everything under `/usr/share/gnome-shell/themes/Yaru-dark`, at least that was the point. While otherwise it may stay under shell domain.

clobrano:
Indeed, the fix - when not using gresource - is simply to call the install script with Yaru-dark name. I'll push the PR asap

> As if we use gresources we can just install everything under /usr/share/gnome-shell/themes/Yaru-dark, at least that was the point. While otherwise it may stay under shell domain.

the alternate directory install under `/usr/share/themes/Yaru-dark/gnome-shell` is used by User Themes extension, that's why we need it, I don't know if it is still necessary when using gresource

chrisjbillington:
Closing in favour of #2050

PR #2050, opened 2020-03-06
build/shell: symlink shell variants to alternate theme directory

clobrano:
in order to make use of User Themes extension, both theme variants need
to be under `/usr/share/themes/<theme-name>/gnome-shell, but Yaru-dark
was missing.

Add back the symlink to Yaru-dark shell variant

Closes: #2043

### DIFF
--- a/gnome-shell/src/meson.build
+++ b/gnome-shell/src/meson.build
@@ -145,6 +145,12 @@ if gnomeshell_use_gresource
     install_dir: gnomeshell_theme_dir,
 )
 else
-  # Symlink default variant
-  meson.add_install_script('install-shell.sh', meson.project_name())
+  # Symlink variants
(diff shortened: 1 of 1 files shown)


clobrano:
@chrisjbillington et alt. :)

This is how I tested this.

1. switch to a different theme, totally, both shell and gtk
2. delete `/usr/share/themes/Yaru-dark` and `/usr/share/gnome-shell/theme/Yaru-dark`
3. build Yaru
4. check `/usr/share/themes/Yaru-dark` shall have a symlink to `/usr/share/gnome-shell/theme/Yaru-dark`

chrisjbillington:
Sweet, this works for me with the Arch package.

Seems like a better fix than just duplicating lines in install-shell.sh

PR #1896, opened 2020-02-11
Bionic sync20.04

clobrano:
Manually sync bionic to the latest master for 20.04

~~(PR opened only to have the testing snap, **not** actually ready for review)~~

I finally fixed all the build issues. The PR **is** open for review, even if probably some more details have to be merged since in the meantime master branch moved forward

Things I suggest to check carefully:
- bionic specific style: e.g. the shell
- icons

PR #2051, opened 2020-03-06
build: build gtk-common-themes for master branch

clobrano:
- do not skip gtk-common-theme build for master

Also
- keep building communitheme snaps for bionic branch only
- keep xenial sources list targets since snapcore/snapcraft:stable is a
xenial machine

### DIFF
--- a/build-helpers/prepare-build-snap
+++ b/build-helpers/prepare-build-snap
@@ -21,18 +21,15 @@ if [ "$TRAVIS_BRANCH" == "bionic" ]; then
         # this is an official edge snap
         channel="edge"
     fi
-else
-    echo "skipping: snaps are only for bionic builds"
-    exit 0
+    echo "Push to bionic ubuntu repo: will release to $channel"
+
+    cmd="$cmd && echo \"$SNAP_TOKEN\" | snapcraft login --with - && snapcraft push *.snap --release=$channel"
+
+    docker run -v $(pwd):$(pwd) -t snapcore/snapcraft:stable sh -c "$cmd"
 fi
 
-echo "Push to bionic ubuntu repo: will release to $channel"
 
-cmd="$cmd && echo \"$SNAP_TOKEN\" | snapcraft login --with - && snapcraft push *.snap --release=$channel"
-
-docker run -v $(pwd):$(pwd) -t snapcore/snapcraft:stable sh -c "$cmd"
-
-# build now gtk-common-themes
+echo "Build now gtk-common-themes"
 # FIXME: we won't need this once snapd can accept and auto-install theme snaps.
 mkdir -p ../gtk-common-themes/snap
 cd ../gtk-common-themes
@@ -49,6 +46,8 @@ docker run -v $(pwd):$(pwd) -t snapcore/snapcraft:stable sh -c "$cmd"
 
 # write on PR install instructions for the snap branch
 [ "$TRAVIS_PULL_REQUEST" == "false" ] && exit 0
-curl -H "Authorization: token ${GITHUB_TOKEN}" -X POST \
-        -d "{\"body\": \"A new test snap version is available using: \`snap refresh communitheme --channel=$channel\`.\nFurther available updates will track that pull request then.\n\nSwitch back to stable or edge snap with \`snap refresh communitheme --stable\` or \`snap refresh communitheme --edge\` once you are done with it!\nYou may want to download as well \`gtk-common-themes\` snap from \`$channel\` channel to test your snaps with those changes.\n\nReplace \`refresh\` with \`install\` if you haven't installed the snap yet.\"}" \
-        "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/issues/${TRAVIS_PULL_REQUEST}/comments"
+if [ "$TRAVIS_BRANCH" == "bionic" ]; then
+    curl -H "Authorization: token ${GITHUB_TOKEN}" -X POST \
+            -d "{\"body\": \"A new test snap version is available using: \`snap refresh communitheme --channel=$channel\`.\nFurther available updates will track that pull request then.\n\nSwitch back to stable or edge snap with \`snap refresh communitheme --stable\` or \`snap refresh communitheme --edge\` once you are done with it!\nYou may want to download as well \`gtk-common-themes\` snap from \`$channel\` channel to test your snaps with those changes.\n\nReplace \`refresh\` with \`install\` if you haven't installed the snap yet.\"}" \
+            "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/issues/${TRAVIS_PULL_REQUEST}/comments"
+fi
--- a/build-helpers/prepare-build-snap
+++ b/build-helpers/prepare-build-snap
@@ -10,9 +10,10 @@ cmd="sed -i s/xenial/bionic/g /etc/apt/sources.list && apt update -qq && cd $(pw
 # Only release as a snap from main repo (ubuntu slug), enabling other forks to test build the snap.
 # Notes that the tokens are only available when the source repo is under ubuntu/ project.
 YARU_BRANCH="$TRAVIS_BRANCH"
-if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-    echo "building PR $TRAVIS_PULL_REQUEST_BRANCH"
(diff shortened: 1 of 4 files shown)


clobrano:
This has the same problem I had with #1896 

```
15249
Error while processing...
15250
The store was unable to accept this snap.
15251
  - package contains external symlinks: share/themes/Yaru/gnome-shell
```
@didrocks, @jhenstridge the symlink is needed to support "User themes" extension without copying the whole gnome-shell twice (one under `/usr/share/gnome-shell` directory and one under `/usr/share/themes/yaru-dark/gnome-shell` which is used by User Themes).

Do you think is viable a patch to User themes to look into the right path as well? Viable in the sense that "I can do it, as well" ? :smile:

jhenstridge:
Can you provide some details about where the symlink is from and to?  I'm unfamiliar with what you mean by "user themes" here.

clobrano:
Of course, sorry :)
[User themes](https://gitlab.gnome.org/GNOME/gnome-shell-extensions) is an extension that allows to change gnome shell theme from gnome tweaks.

This extension consider 1 *default* theme under `/usr/share/gnome-shell/theme/` and looks for alternative themes under `/usr/share/themes/<theme-name>/gnome-shell`. So in Ubuntu session the default shell theme is Yaru, while in vanilla session the default is Adwaita.

On master branch (which is not released as snap) we started to provide

1. a symlink from `/usr/share/gnome-shell/theme/Yaru-dark` to `/usr/share/themes/Yaru-dark/gnome-shell`, so that one can select Yaru-dark shell theme from gnome-tweaks, in both Ubuntu and Vanilla session

2. a symlink from `/usr/share/gnome-shell/theme/Yaru` to `/usr/share/themes/Yaru/gnome-shell` so that one can select Yaru gnome-shell also in Vanilla session.

chrisjbillington:
I filed an issue with the user themes extension suggesting it look in `/usr/share/gnome-shell/theme` as well:

https://gitlab.gnome.org/GNOME/gnome-shell-extensions/issues/214

It looks fairly trivial, just adding an additional path to an existing list of paths, but both user themes and gnome-tweak tool will need to be patched.

clobrano:
Thanks @chrisjbillington, 

I think I can takle the extension patch, since - as you said - seems easy (unless there are reasons I am not aware to keep it this way)

clobrano:
Oh, I'd probably avoided to talk about what Yaru does :D, but it's ok

jhenstridge:
Could you just turn this into a relative symbolic link?  That is, rather than:

    /usr/share/gnome-shell/theme/Yaru-dark -> /usr/share/themes/Yaru-dark/gnome-shell

Use:

    /usr/share/gnome-shell/theme/Yaru-dark -> ../../themes/Yaru-dark/gnome-shell

The problem the snap review-tools are complaining about is that you have a symlink that points outside of the snap.  A relative symlink would point at the right content, no matter where the snap is mounted.

chrisjbillington:
Does that imply that the link and its target aren't in the same snap? (I am unfamiliar with whether there is more than one snap). If so, shouldn't that be resolved instead?

I remember an issue in the past where a symbolic link to the shell theme was in the deb package for the gtk theme, and it should have  been in the .deb package for the shell theme, obviously - despite the location in the filesystem of the link being next to the gtk themes.

clobrano:
thanks @jhenstridge, I'll change the script

jhenstridge:
The review-tools check will trigger on any absolute symlink, so the fix is to create a relative symlink.  The fix would be to change

https://github.com/ubuntu/yaru/blob/master/gnome-shell/src/install-shell.sh#L9

to:

    ln -sf "../../gnome-shell/theme/${project_name}" "${destdir_prefix}/themes/${project_name}/gnome-shell"

This should behave equivalent to what you have in non-snap contexts, and avoid review failure in the snap case.

clobrano:
It seems that the review-tools are well designed :smile:. It still detects the external symlink and forbits the push on the store

clobrano:
I proposed the upstream patches to get rid of those symlinks

- https://gitlab.gnome.org/GNOME/gnome-shell-extensions/-/merge_requests/111
- https://gitlab.gnome.org/GNOME/gnome-tweaks/-/merge_requests/53

jhenstridge:
It's not at all clear that your rel_install_prefix change actually produces the correct link path.  Have you actually tried what I suggested?

I don't understand why you're using realpath here: the relationship between the source and destination of the symlink is fixed, so you know exactly what the relative path will be.

clobrano:
In a test version I was printing the path, it was correct, but I can also double check using fixed path.

I used `realpath` to support other packagers (e.g. arch) I am not sure they use the same paths since they define DESTDIR

jhenstridge:
The logic just looks wrong.  Let's say that you had configured the project with `MESON_INSTALL_PREFIX=/usr`, and then performed a `DESTDIR` build to `/tmp/install`.

In this case, I think we have:

    install_prefix=/usr/share
    dest_dir=/tmp/install/usr/share/themes/Yaru

That should end up setting `rel_install_prefix=../../../../../../usr/share`, so you end up with a symlink like:

    /tmp/install/usr/share/themes/yaru/gnome-shell -> ../../../../../../usr/share/gnome-shell/theme/Yaru

Note that this relative path backtracks past our `DESTDIR` of `/tmp/install`.  This is what review-tools is complaining about, since it means the symlink could point at files outside of the snap.  What you want is a symlink that looks like this:

    /tmp/install/usr/share/themes/yaru/gnome-shell -> ../../gnome-shell/theme/Yaru

Don't try to be clever with realpath: just use the correct relative path in your `ln -s` call.  The code will be simpler, and will actually work.

clobrano:
Okay, I'll try that, thanks

jhenstridge:
Could you try updating the build script to run "unsquashfs -ll" on the gtk-common-themes snap before it tries to publish it?  That should tell us exactly what the symlink looks like inside the snap.

clobrano:
> Could you try updating the build script to run "unsquashfs -ll" on the gtk-common-themes snap before it tries to publish it? That should tell us exactly what the symlink looks like inside the snap.

I'll check how, but also now I don't understand the difference between the two CI checks. They seem to do the same thing now, and the second one succeeds pushing the snap. You can even see which current working directory `install_shell.sh` is running into and the symlink

jhenstridge:
I was suggesting you run "unsquashfs -ll filename.snap" on the result of the snap build.  I don't think the meson build is the place to put such an experiment.

Looking at the CI script, it probably needs to go in `build-helpers/prepare-build-snap` script prior to pushing the snap to the store.

clobrano:
I had to reduce the output of `unsquashfs` due to CI logging limits, so I filtered only Yaru related things, hopefully it's enough.

jhenstridge:
Sorry for the delay getting back to you.

In the CI log, I see:

```
lrwxrwxrwx root/root                30 2020-03-29 15:38 squashfs-root/share/themes/Yaru/gnome-shell -> //share/gnome-shell/theme/Yaru
lrwxrwxrwx root/root                35 2020-03-29 15:38 squashfs-root/share/themes/Yaru-dark/gnome-shell -> //share/gnome-shell/theme/Yaru-dark
```

So that's two absolute symbolic links.  It doesn't match up with the code in your branch, so I wonder what is going on?  It's hard to tell what source code is actually being built from the log.

clobrano:
It's definitely not running my latest commit. The `install-shell.sh` calls `set -x` so we should see all the commands in that script. I wonder what happened

clobrano:
Alright, it seems to work now, but I still need to change one thing. I want to build gtk-common-theme snap only for Pushes to master, not for PRs

Feichtmeier:
Very nice @clobrano :tada: 

One thing: since there are some gtk apps as snaps installed in focal, any idea how we dodge the "UI freeze" problem? Or is @flexiondotorg okay with a "fluid design" approach in Ubuntu, similar to what Windows10 does?

clobrano:
Good point. It's probably wrong rebuild gtk common for pushes on master, since the yaru counterpart is not released as well.

One more thing to check (maybe on Monday) is what happened to the gtk-common-themes just build by previous CI job. Since it took the content of this PR it might be out of dates

clobrano:
I think more and more that goal of this PR is incorrect. I am going to close it, sorry for wasting your time @jhenstridge

PR #2058, opened 2020-03-10
prepare for release 20.04.3

clobrano:
- [x] wait for Launchpad bug IDs


didrocks:
Pushed some small fixes. As we need to wait for the new Shell to be in focal to release it, let’s just prepare and merge that to master.
Thanks!

PR #2071, opened 2020-03-14
Removing unnecessary objects from legacy action svgs

ubuntujaggers:
Just realised a load of these still have "work in progress" objects left outside the canvas.  They don't affect appearance (because they're not in view) but needlessly bloat the files.  Fixed with this commit.


Feichtmeier:
Looks good :+1: Thanks!

Issue #2070, opened 2020-03-14
Style dark headerbar similar to light theme


madsrh:
I'm not sure how big the diff is here, but what I immediately noticed in backdrop was the button borders when using the Terminal (dark theme) in the light theme.

![image](https://user-images.githubusercontent.com/3986894/76688820-bf2bbf00-6630-11ea-949e-e441f046cf7d.png)

madsrh:
@clobrano What do you think about a border color somewhere in between the two current ones?

clobrano:
Not sure, but I'll post some attempts

PR #2072, opened 2020-03-14
Reduce differences between default and dark variant headerbar buttons

clobrano:
- use the same backdrop border color
- use the same background color

Fixes: #2070

### DIFF
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -1,7 +1,8 @@
 // Color variables dependant on $headerbar_bg_color defined in _colors.scss
 $_button_bg_color: lighten($headerbar_bg_color, 6%);
 $_button_border_color: darken($headerbar_bg_color, 9%);
-$_backdrop_button_border_color: $headerbar_bg_color;
+// deducted from the color used in common which is $backdrop_borders_color: mix($borders_color, $bg_color, 80%);
+$_backdrop_button_border_color: darken($inkstone, 9%);
 
 // overwriting the headerbar styling from common, for the inverted look in the ambiance theme
 headerbar,
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -91,7 +91,7 @@ headerbar,
     }
 
     &:disabled {
-      @include button(insensitive, if($variant=='light', darken($headerbar_bg_color, 15%), $headerbar_bg_color), $headerbar_fg_color);
+      @include button(insensitive, if($variant=='light', darken($headerbar_bg_color, 9%), $headerbar_bg_color), $headerbar_fg_color);
       border-color: darken($headerbar_bg_color, 6%);
 
       &:active,
--- a/gtk/src/default/gtk-3.20/_headerbar.scss
+++ b/gtk/src/default/gtk-3.20/_headerbar.scss
@@ -1,5 +1,5 @@
 // Color variables dependant on $headerbar_bg_color defined in _colors.scss
-$_button_bg_color: lighten($headerbar_bg_color, 6%);
+$_button_bg_color: lighten($headerbar_bg_color, 7.5%);
 $_button_border_color: darken($headerbar_bg_color, 9%);
 // deducted from the color used in common which is $backdrop_borders_color: mix($borders_color, $bg_color, 80%);
 $_backdrop_button_border_color: darken($inkstone, 9%);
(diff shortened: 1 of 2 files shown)


clobrano:
This is the base change, I am going to attempt some different backdrop borders as discussed

clobrano:
image

In the image above, the window at the bottom is nautilus in default variant, while the one in between is a terminal, so in dark variant.
The difference between the two backdrop button borders is 2% (default is 2% lighter). I think that, looking at the whole desktop, this 2% lighter is already a bit too much and the borders start to fading into the background. What do you think @madsrh ?

gtk/src/default/gtk-3.20/_headerbar.scss
```diff
@@ -1,5 +1,5 @@
 // Color variables dependant on $headerbar_bg_color defined in _colors.scss
-$_button_bg_color: lighten($headerbar_bg_color, 6%);
+$_button_bg_color: lighten($inkstone, 2.5%);
```
Feichtmeier: Hmmmm... I would not do it like this. Inkstone is too different from headerbar_bg_color
Better in _drawing here:
https://github.com/ubuntu/yaru/blob/master/gtk/src/default/gtk-3.20/_drawing.scss#L199
This line is too long anyways. Just lighten them up by 4% for light and dark version

Feichtmeier: Like this:
$button_fill: if($c == $bg_color, image(lighten($c, 4%), image(lighten($c, 2%))) !global;

clobrano:
This isn't the same I think. Drawing is used in headerbar as well, if you change that, default and the others variants will differ anyway

clobrano:
Nah, sorry, I got it now

madsrh:
This looks good @clobrano 👍 Are you +1 @Feichtmeier ?

There's still a tiny difference - the button border in the light theme is slightly darker, but you REALLY have to look hard to see it 😄 

We should leave the rest, right? border-top and header background

image

Feichtmeier:
We could also think about making headerbar bg colour just descending from inkstone to have full dark theme sync there

clobrano:
In that case, we'll drop the warm color I'd prefer to keep

madsrh:
> In that case, we'll drop the warm color I'd prefer to keep

We can keep it if you prefer - my issue was only regarding the button borders.

Feichtmeier:
Sure, okay, then just brighten them up in drawing and the contrast is as good as in the dark theme :+1: 
(funny that I was first quiet resistant about this change :) )

gtk/src/default/gtk-3.20/_headerbar.scss
clobrano: It turns out that since dark variant is #4a4a4a, the closest change is increasing `headerbar_bg_color` by 1.5%, which is  #4d4a4a, while using 4% lighten in drawing makes #494646

clobrano:
> We should leave the rest, right? border-top and header background
> 
> image

Oh, nice catch. This comes from the window decorator that is the one used for light variant. I'll try to change it

Feichtmeier:
@clobrano could you put more screenshots when you're done?
Wiped focal again because world of warcraft stopped to work on focal and now back on bionic :robot:  :D

clobrano:
image

Feichtmeier:
Looks good
What happened to disabled/checked/hover/backdrop and so on after you've lightened the button bg up? All good?

clobrano:
> What happened to disabled/checked/hover/backdrop and so on after you've lightened the button bg up?
These states use the same base color, so everything changed accordingly

backdrop
image

insensitive
image

Colors can't be exactly the same, of course, but they are now the closest possible.

madsrh:
Tested ✔️  👍 

Could we make the disabled button slightly more disabled looking?

- My mockup (less border, background closer to headerbar and less white "text")
- This PR

image

Feichtmeier:
Looks very good Carlo +1

clobrano:
> Could we make the disabled button slightly more disabled looking?

both default and dark, right?

madsrh:
> both default and dark, right?

Yes

clobrano:
I had a look at the possible change and I think we're a bit late to do this now. On default variant is easy and limited, but on light and dark variants we should change the `insensitive` colors which are widely used, this means a lot of testing to avoid regressions.
To give an idea, changing the `insensitive_fg_color` to green and `insensitive_border_color` to red, you can see how many widgets are affected

image

I would suggest keeping this action item as a separate issue and work on it >20.04, what do you think?

Feichtmeier:
I agree

madsrh:
Alright, I see. What about changing the just the border to something less than 6% in this PR?

Feichtmeier:
I would rather merge it as it is mads :+1:

clobrano:
There would be a difference between default and dark variants

Feichtmeier:
Can you put this into bionic? :innocent: @clobrano

PR #2080, opened 2020-03-16
Reduce differences between default and dark variant headerbar buttons

clobrano:
cherry-pick #2072

* headerbar: use the same backdrop border color of dark variant
* headerbar: make button background color the same as dark variant
* common: remove whitespaces
* reduce difference in headerbar insensitive buttons

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -920,14 +920,14 @@ button {
 
     $_border_bg: linear-gradient(to top, $alt-borders-color 25%, $borders-color 50%);
 
-    &:not(.flat):not(:checked):not(:active):not(:disabled):not(:backdrop) {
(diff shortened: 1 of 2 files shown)


didrocks:
A new test snap version is available using: `snap refresh communitheme --channel=edge/yaru-pr2080`.

You may want to download as well `gtk-common-themes` snap from `edge/yaru-pr2080` channel to test your snaps with those changes.

Feichtmeier:
Very nice thanks! Could you also include the other changes to the dark theme headerbar we've made in the past week? 
We can just cherry pick again? Now after you've synced?

clobrano:
The last sync was for the icons, so I'm not sure we can just cherry pick.
For the Last scss I just merged the files

Feichtmeier:
If the scss files are the same as in master everything should be fine! Thank you and sorry for being such and help currently

PR #2083, opened 2020-03-17
Legacy icon for object locked / unlocked

madsrh:
Trying to fix the issue where the lock icon for layers in Inkscape isn't visible because of the red / green clashing with the orange selection.

Before:
image
 
After:
image


ubuntujaggers:
Nice work 👍

ubuntujaggers:
Nice work 👍

PR #2093, opened 2020-03-19
Re-add nautilus-desktop window styling

Feichtmeier:
Re-adding this to master, so we can cherry-pick it into bionic. And after that we should have no difference (hopefully).
Are you okay with this @clobrano even if it is not used for newer nautilus versions?

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -145,6 +145,13 @@ $ambiance: null !default;
     }
 }
 
+.nautilus-desktop-window {
+  .nautilus-desktop.view {
+    color: $selected_fg_color;
+    text-shadow: 0px 1px 2px rgba(0, 0, 0, 0.5);
+  }
+}
+
 // Nautilus disk space dialogue styling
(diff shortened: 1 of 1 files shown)


clobrano:
I'd prefer not having things in master that we don't use

Feichtmeier:
Then we should also remove the gnome-software green box-shadow ;)

clobrano:
If it's not used sure, but here you want to add something new that we already know it's unused. It doesn't make much sense to me.

Feichtmeier:
it's 5 lines in _apps that are annoying to care about if you sync with master, just like it was remove in last sync. :man_shrugging: 
We also have nemo styling in apps

clobrano:
I understand it's a bit annoying, but the point of having two branches is to keep the differences. We can add some comments to help when updating and also cheery pick should not mess up with these differences

Feichtmeier:
There are no differences in the gtk folders except those nautilus-desktop styling, even though we have other styling in _apps that is not used in master

clobrano:
I think we should remove anything not used in master because it's misleading

PR #2094, opened 2020-03-19
Remove the livepatch icon from the yaru repository

Feichtmeier:
To re-add it in the ubuntu repository


clobrano:
Do we need to wait until Ubuntu includes the patch before merging this?

Feichtmeier:
I think sebastian bacher is currently looking for a way to keep the yaru icon in yaru and but then let the livepatch indicator use it's correct icon. They have different names so it should be possible. So eventually we can just close this PR when he is successful. The reason is that they may not want the yaru icon in their repo is that the livepatch is also used in different distros which do not use yaru

clobrano:
I see, thanks

Feichtmeier:
Seb fixed it, so this is not needed: https://launchpadlibrarian.net/472380880/update-notifier_3.192.30_source.changes

PR #2104, opened 2020-03-23
Shell theme asset updates

Feichtmeier:
- keyboard updates: add .symbolic icons to from upstream into the src folder and into the gresource file
- remove the old keyboard assets from the src folder and from the gresource file
- change the corresponding lines in the keyboard.scss file to take the new symbolic assets which should work in the dark and light versions theoretically
- remove useless diff for font-weight: bold asignments because we use a wildcard in _widgets.scss for this
- remove useless diff for media control hover
- remove useless diff as in whitespaces and new lines

@3v1n0 please test with your local gnome-shell 3.36 installation if they keyboard assets show up


gnome-shell/src/data/gnome-shell-theme.gresource.xml
```diff
@@ -35,6 +27,11 @@
     <file alias="icons/pointer-drag-symbolic.svg">pointer-drag-symbolic.svg</file>
     <file alias="icons/pointer-primary-click-symbolic.svg">pointer-primary-click-symbolic.svg</file>
     <file alias="icons/pointer-secondary-click-symbolic.svg">pointer-secondary-click-symbolic.svg</file>
+     <file alias="icons/keyboard-caps-lock-filled-symbolic.svg">keyboard-caps-lock-filled-symbolic.svg</file>
```
3v1n0: Minimal, but a diff with upstream... Remove one leading space char

3v1n0:
Ok, I've merged this with the fixes and most of all with commit 2a36fa1f7463a4806a5eaf7092f8a50bd2d298ef that was the trick to fix the problem we had with the icons being ignored.

So we can close... Actually, github, this is MERGED! :)

Issue #2114, opened 2020-03-26
Bring back the focus ring to the tab and nautilus widgets


Amr-Ibra:
Referring to this [comment](https://github.com/ubuntu/yaru/issues/2109#issuecomment-604145456) and the discussion afterwards:
the focus ring indicates the transition between an element which is selected and an element that "just got the focus".

Issue #2109, opened 2020-03-25
Hover and press effect on tabs looks better in Adwaita


Amr-Ibra:
### Steps to reproduce the problem
1. Install geany (sudo apt install geany)
2. Hover the mouse pointer and press on the left and bottom tabs, including the arrow tabs
3. Change theme between Yaru and Adwaita

![Hover_Press](https://user-images.githubusercontent.com/45657128/77522785-188fbb80-6e85-11ea-8922-f001fccd0d38.gif)

Notice the difference in the hover and press effect, especially on the arrow tabs, between Yaru and Adwaita. In my opinion, Adwaita looks better.

Notice that clicking twice on a tab does not draw a dotted rectangle around it, which indicates that the tab has keyboard focus. Go up, down, left, right; the dotted rectangle moves along. Adwaita does that, Yaru does not.

### Software that presents the issue
- Name: geany
- Version: 1.36

### Yaru version
apt show yaru-theme-gtk
Package: yaru-theme-gtk
Version: 20.04.4

Feichtmeier:
This is exactly the hover effect @madsrh and @clobrano wanted by design, so I'll close this

Amr-Ibra:
But clicking twice on a tab does not draw a dotted rectangle around it, which indicates that the tab has keyboard focus. Go up, down, left, right; the dotted rectangle moves along. Adwaita does that, Yaru does not.

clobrano:
If you notice, the tab doesn't really have the outline. Once it is selected, it is on focus

Amr-Ibra:
The dotted rectangle is needed for keyboard navigation. If the keyboard focus is somewhere, and you use only the Tab key to switch the focus around the elements, you wouldn't know where the focus stands at the moment.

That also applies to Nautilus:
1. Open many tabs in Nautilus
2. Focus the keyboard in the file area
3. Switch the focus with the Tab key around all Nautilus elements
4. Observe that the tabs do not indicate whether they have keyboard focus or not
5. Now try the same thing with Adwaita

clobrano:
What I meant is that, when we discussed this, we observed that the selected tab and the dotted rectangle seems to always go together. Do you observe a different behaviour in adwaita?

Amr-Ibra:
> Do you observe a different behaviour in adwaita?

Yes, don't use the mouse and use only the keyboard. Follow the steps I mentioned for nautilus and switch the focus using the `Tab` key.

clobrano:
[...]

PR #2117, opened 2020-03-27
tweaks: outline for tabs

clobrano:
In order to show when the tab has "just got" the focus with keyboard the
outline is still necessary.

Add a thinner style for tab's outline, since the normal 2px outline
seems too heavy for this widget.

Fixes: #2114

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -148,8 +148,9 @@ $ambiance: null !default;
       // First, remove the background
       background-color: rgba(255, 255, 255, 0.1);
       // Then bring it back for when it receives a direction
-      &:dir(ltr) {
-        background-color: $selected_bg_color;
+      &:selected:not(:backdrop) {
+        &:dir(ltr) { background-color: $selected_bg_color; }
+        &:not(:dir(ltr)) { background-color: rgba(175, 175, 175, 0.5); }
       }
 
       &:backdrop {
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -327,7 +327,8 @@ notebook > header {
       &, &.right, &.left, &.top, &.bottom {
         background-color: transparent;
         tab {
-          outline-color: transparent;
+          outline-width: 1px;
+          outline-offset: -1px;
           &, &:dir(ltr), &:dir(rtl) {
             &, &:backdrop {
               &, &:checked {
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -328,7 +328,7 @@ notebook > header {
         background-color: transparent;
         tab {
           outline-width: 1px;
-          outline-offset: -1px;
+          outline-offset: -2px;
           &, &:dir(ltr), &:dir(rtl) {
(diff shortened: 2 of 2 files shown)


clobrano:
Peek 2020-03-27 13-07

Amr-Ibra:
Thanks @clobrano 

Does this also work for files/folders in Nautilus? I explained the same principle in this [comment](https://github.com/ubuntu/yaru/issues/2109#issuecomment-604284860).

Feichtmeier:
No but we can use $colored_focus_border for the .selected Elements

clobrano:
Why no?

Peek 2020-03-27 16-46

Amr-Ibra:
> Why no?

I mean apply the same analogy to file/folder icons in Nautilus using your screencast. You will see that there is a transition between a file selected (blue highlight only on the file name) and a file that "just got the focus" (blue highlight on both file name and icon). Adwaita does that, Yaru does not.

grafik

clobrano:
Uhm, I see what you mean here, let me check

clobrano:
What do you think of something like this?
Peek 2020-03-27 20-19

Feichtmeier:
Hm.... the tabs looked good in the widget factory but in nautilus :( not so much
COnerning the tint... I guess this question goes more to @madsrh

clobrano:
It looks better in light and dark variants, indeed.

Amr-Ibra:
> Hm.... the tabs looked good in the widget factory but in nautilus :( not so much

If you think about it, the outline serves a purpose then disappears. It's short-lived.
It looked better because the tabs in nautilus are wide. But I said it's short-lived.

madsrh:
> COnerning the tint... I guess this question goes more to @madsrh

I think it's a good solution 👍

clobrano:
Alright, this the full circle

Peek 2020-03-27 22-23

I also notice a "bug", in both Yaru and Adwaita, so I think it's something related to Nautilus itself: if no folder is selected, when the focus jumps from the tab to the icon, nothing seems selected.

In below gif, I pressed TAB, when the focus ring disappears from the tab and nothing appeared in the view. Only when I pressed the arrow key, the folder got both selection and focus.
Peek 2020-03-27 22-27

Feichtmeier:
I am really not sure about the focus ring on the tab :| Same as last time we discussed it. Maybe if you make the focus ring smaller?

clobrano:
The reason of this outline offset is to draw the bottom line of the outline over the orange underline of the tab. Otherwise the tab surface is crowded with too many orange line and looks messy for me

clobrano:
This is 1px smaller 
image

Two pixels smaller, already makes the bottom rounded corners visible and kind of ruins the effect
image

Feichtmeier:
I am fine with 1px smaller :+1: Even if I would love to just make it invisible, but @Amr-Ibra seems really attached to this issue emotionally, so :+1:

clobrano:
Ok for the change. Just a note, this is an accessibility issue, so it is important to fix it

clobrano:
This one might need a UIFe, so let's wait for merging it

Feichtmeier:
How does multi select look in nautilus?

clobrano:
This one, do you mean?
image

Feichtmeier:
Great that's unchanged +1
Just thought it's worth a double check after I've seen your commit :) good night

clobrano:
I think this can be considered as a bug actually, so we can merge it

PR #2118, opened 2020-03-28
Adding symbols for system-reboot and system-log-out

### LINKED ISSUE #2073: system-log-out-symbolic and system-reboot-symbolic do not exist in Yaru
`system-log-out-symbolic` and `system-reboot-symbolic` do not exist in Yaru. These icons are supposed to be used in gnome-shell when you search for "logout" and "reboot".

As a result, gnome-shell shows the icon `application-exit-symbolic`, which is basically an [X](https://github.com/ubuntu/yaru/blob/master/icons/Suru/scalable/actions/application-exit-symbolic.svg), instead when you search for "logout".

**Adwaita icons**

[system-log-out-symbolic](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/Adwaita/scalable/actions/system-log-out-symbolic.svg)
grafik

[system-reboot-symbolic](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/Adwaita/scalable/actions/system-reboot-symbolic.svg)
grafik

ubuntujaggers:
When I build, I don't see any change in behaviour when I search for "reboot" and "logout" in the overlay...  but here are two symbols with the requested names:

image

image

Also, is there any way to edit a commit message, because I accidentally called system-reboot "system-shutdown" in the first one :)

Closes #2073


Feichtmeier:
@ubuntujaggers 
> Also, is there any way to edit a commit message, because I accidentally called system-reboot "system-shutdown" in the first one :)

yes!

```bash
git commit --amend
```

Opens the commit editor with your last local commit which you can change
Then you need to

```bash
git push --force
```

To overwrite the commit in the repo

Feichtmeier:
Alternatively you should be able to change the final commit with the squash and merge button

ubuntujaggers:
> Alternatively you should be able to change the final commit with the squash and merge button

I'll do that this time and remember the command for next time :+1:

Feichtmeier:
@didrocks do we need an UIFE for this?

Amr-Ibra:
I don't think this needs a UIFE because the icons are not visible by default (according to [the UserInterfaceFreeze wiki](https://wiki.ubuntu.com/UserInterfaceFreeze)), and this fixes an issue in Yaru that doesn't exist in Adwaita (according to the [UserInterfaceFreeze Exceptions wiki](https://wiki.ubuntu.com/FreezeExceptionProcess#UserInterfaceFreeze_Exceptions)).

didrocks:
Sorry, I missed it. This is slightly incorrect, you need an UIFE because it’s something installed by default and so, can appear (it’s not only "not visible by default" as the rule. I suggest anyway a big UIFE request before the next Yaru release with the list of changes for the documentation team.

clobrano:
Indeed, but what if only some of the changes are not accepted?

didrocks:
We’ll remove them from master. I doubt anything major is changing though, correct? So, it should be easy to get it accepted :)

clobrano:
Yeah, indeed, I was just curious about the process :)

PR #2132, opened 2020-04-05
gnome-shell/calendar: darken event dot

clobrano:
In the default Yaru shell theme, the calendar dot of days having an
appointments is not visible. The dark shell theme is not affected.

The problem is due to the color of calendar-today.svg being too similar
to the calendar background.

Fix consists in changing the color of calendar-today.svg in a way that
is both visible in the light and dark shell theme.

Fixes: https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1870938

image
image

### DIFF
--- a/gnome-shell/src/calendar-today.svg
+++ b/gnome-shell/src/calendar-today.svg
@@ -14,7 +14,7 @@
    height="24"
    id="svg10621"
    version="1.1"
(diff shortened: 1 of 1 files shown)


clobrano:
To make the change effective, let build gnome-shell gresource file

PR #2133, opened 2020-04-06
gnome-shell/calendar: do not change color of day with events

clobrano:
When a non-working day has events, the color is different from other
non-working days without any event.

This happens because the color of days with events is set explicitly,
this might be wanted, but the view looks uneven this way.

To prevent this, do not explicitly set a specific color for days
with events. The color is expected to stay the same regardless the day
being a working or non-working one.


clobrano:
This is another change that might require a UIFe

clobrano:
To make this work, enable gnome-shell build to generate the gresource

Issue #2079, opened 2020-03-16
Doesn't build with libsass 3.6.3


chrisjbillington:
I'm on Arch Linux, which recently updated to libsass 3.6.2, though I've since manually updated to 3.6.3. I see that [3.6.3 will likely be in Ubuntu Focal](https://bugs.launchpad.net/ubuntu/+source/libsass/+bug/1867116), so that version seems relevant.

Anyway, Yaru fails to build with libsass 3.6.2 or 3.6.3 - it hangs at the following on 3.6.3:

```$ ninja -C build
ninja: Entering directory `build'
[54/66] Generating light-gtk-3.20-gtk-generated.css with a custom command.
WARNING on line 283, column 15 of /tmp/yaru/gtk/src/default/gtk-3.20/_headerbar.scss:
Compound selectors may no longer be extended.
Consider `@extend *, :link, :selected` instead.
See http://bit.ly/ExtendCompound for details.

[55/66] Generating gtk_gresource with a custom command.
```

The hanging might be a bug in libsass, but the warning should presumably be addressed.

clobrano:
Thanks Chris, I'll have a look

chrisjbillington:
Relevant bug report in gtk:

https://gitlab.gnome.org/GNOME/gtk/issues/2237

Not sure if this implies an update to glib will fix it - I have glib 2.64, which looks to be the most recent

chrisjbillington:
A relevant bug report in libsass: 

https://github.com/sass/libsass/issues/3033

I do see memory usage growing indefinitely

AdsonCicilioti:
Hi ,
I am also on the latest Linux arch and my build stops at step 47/66. It accuses several non-existent .css files, and stops at a step where it seems to be renaming files.

output of `meson build`:

```
 meson build              
The Meson build system
Version: 0.53.2
Source dir: /home/adson/.src/yaru
Build dir: /home/adson/.src/yaru/build
Build type: native build
Project name: Yaru
Project version: 19.10
Host machine cpu family: x86_64
Host machine cpu: x86_64
Configuring cursor.theme using configuration
Configuring index.theme using configuration
Program meson/post_install.py found: YES (/home/adson/.src/yaru/icons/meson/post_install.py)
Program gresources-xml-parser.py found: YES (/home/adson/.src/yaru/gnome-shell/gresources-xml-parser.py)
[...]

PR #2135, opened 2020-04-07
do not extend compound selectors in gtk-3.0

clobrano:
with libsass>3.6.2 compound selectors may not be extended anymore.

using placeholder selector instead (%)

Closes: #2079


Dragon8oy:
Builds fine on Debian Unstable with libsass 3.6.3

clobrano:
Awesome, thank you!

PR #2143, opened 2020-04-09
Adding dialog-warning and symlinks

### LINKED ISSUE #2103: Missing symlinks for dialog warning
grafik
grafik

Could be either
```gtk-dialog-warning```
or
```important```
or
```messagebox_warning```
or
```stock_dialog-warning```

that need a symlink to

```dialog-warning```

@ubuntujaggers could you fix that?

ubuntujaggers:
Closes #2103

image


clobrano:
I wonder if this is considered more than a bug fix, since without this the old icon was shown

@didrocks, @seb128 what do you think?

seb128:
It sounds like a fix but it could impact on screenshot so it would be good to ask Gunnar/the document team how they feel about the change

clobrano:
Right, I'll add it to the LP bug then

PR #2145, opened 2020-04-10
Suggested change to Extensions app - is this more "Yaru"?

ubuntujaggers:
image

image

image

?


real-amano:
Personally I find that the most pleasing suggestion up till now.
And a good compromise between all pieces colored and three pieces the same grey.

PR #2154, opened 2020-04-16
Shell: import some changes from 3.36 upstream branch

3v1n0:
I didn't notice anything wrong with them (no actual changes TBH), so feel free to look at these deeper :)


clobrano:
LGTM 👍

clobrano:
Do we need this before 20.04 release?

3v1n0:
> Do we need this before 20.04 release?

Personally I didn't find any difference with or without it... So if you can look better maybe you notice whether it's needed or not.

But from what I see I think we can wait

clobrano:
I don't see differences as well, let's wait

Thank you 😉️

clobrano:
I think we can take this now

PR #2157, opened 2020-04-19
Adding full colour app icon for accessories-dictionary

ubuntujaggers:
Not the most exciting icon in the world  :)  but I installed a dictionary app from Software Centre which picked up one of our symbols:

image

...so I've added an app icon to stop that happening:

image

image

image

image


Feichtmeier:
that's a stylish suprise!
+1 looks amazing

Issue #2060, opened 2020-03-11
Default theme: Gnome clocks "hdysqueezer" buttons have wrong hover color


Feichtmeier:
![Bildschirmfoto von 2020-03-11 12-28-54](https://user-images.githubusercontent.com/15329494/76412485-faf62880-6393-11ea-859f-355c0776c88a.png)

Probably because they are not included in our _headerbar.scss styling
Issues so far:

- [ ] wrong hover/active/backdrop styling for those squeezer buttons
- [x] our compact mode is not working there, and the squeezer buttons stretch the whole headerbar including the buttons
- [x] circular buttons are stretched in the "add new" dialogue

clobrano:
I'm afraid this is a common problem for non-adwaita themes. This is Materia

![image](https://user-images.githubusercontent.com/2883614/76452728-e042a480-63d1-11ea-8c7e-8c3294ddc97a.png)

clobrano:
It is basically using the exported background color. If the exported theme_bg_color is green, also the hover effect becomes green

![image](https://user-images.githubusercontent.com/2883614/76454477-07e73c00-63d5-11ea-96a8-a8ad569775bc.png)

Feichtmeier:
Yes it's a problem with light themes with a dark headerbar :) 
So, what shall we do?

clobrano:
Basically we can't fix it

clobrano:
I'd try to build ghome-clocks without some styling and propose a patch if can find one, but not really soon

ubuntujaggers:
Do we have to supply some symbols for these widgets or do we think the design is baked into the app?

Feichtmeier:
I think those are installed by the app under hicolor
Couldn't find those here https://gitlab.gnome.org/GNOME/adwaita-icon-theme/blob/master/src/symbolic/gnome-stencils.svg

Feichtmeier:
@clobrano okay, then let's keep it opened so people don't own new bugs? Or rather close it?
I would suggest as a long term goal for 22.04 to either introduce headerbar colours into gtk or drop the mixed theme

clobrano:
Let's keep it open for now

clobrano:
This is a promising step forward https://aplazas.pages.gitlab.gnome.org/blog/blog/2020/04/02/coloring-api.html

Feichtmeier:
Concerning stretched sizes:

[...]

PR #2162, opened 2020-04-20
Yaru bigger buttons

clobrano:
Draft project to compare standard Yaru with a Yaru that uses upstream button sizes.

This PR installs a project called YaruBigBoss :sweat_smile: in three variants (default, light, dark), so it is easier to compare with standard Yaru (which is not overwritten)


clobrano:
did anyone test this? 😄

Feichtmeier:
Not yet sorry, I will later!

Feichtmeier:
Works 👍 🚀 
image

Now what is the plan with this? 😆

clobrano:
😄 it has no purpose except let @madsrh and @ubuntujaggers compare current Yaru with this one

ubuntujaggers:
I actually forgot I had the big buttons for a while so if there's a big supportibility reason i wouldn't mind. I'll build it again tonight 👍

madsrh:
Sorry for not getting around to testing the earlier 😬 This is pretty cool @clobrano You're great with names 😆 

It seems unnecessary huge at first, but after a while I didn't notice it. I still prefer the current Yaru sizes though. Are the diff from Adwaita causing issues or do you just prefer big button (and big shadows) yourself? 😀

clobrano:
Honestly, I don't like it 😆
On 12" screen is terrible, but it was suggested by @Feichtmeier to overcome some button issues, so I provided a test branch

Feichtmeier:
Yeah I also don't like it visually. I think we should postpone it as long as we can. Depending on the future of gtk apps we might need to do it anyways. It's a one line change sooooo let's wait. I think

Feichtmeier:
> Sorry for not getting around to testing the earlier grimacing This is a pretty cool way to compare @clobrano and you're great with names laughing
> 
> It seems unnecessary huge at first, but after a while I didn't notice it. I still prefer the current Yaru sizes though. Are the diff from Adwaita causing issues or do you just prefer big button (and big shadows) yourself? grinning

Just to clarify: our sizes are fine as long as people do not put Adwaita-only-sized widgets into the headerbar and stick with common widgets like buttons, entries and so on. See #2060

Feichtmeier:
Shall we close this then @clobrano ? If no one likes it visually :shrug:

PR #2186, opened 2020-05-07
Dash to dock padding fixes

3v1n0:
And sync settings from [upstream stylesheet](https://github.com/micheleg/dash-to-dock/blob/ubuntu-dock/stylesheet.css).


Feichtmeier:
@3v1n0 odd place to ask, but since it looks to be that you have became one of the main dash to dock contributors I thought I could ask:

is there any way to reduce the yaru dock styling to a minumum? 
I think this idea how we use this dock.scss file comes from the very beginning of the project. 
I often thought that it was weird to paste in the whole style sheet there and that the file is a little bit confusing because you also have the dash styling which comes from upstream (like 95% of our styling)
In the end - I think - our changes break down to two changes:

- green notification badges
- different padding

Issue #2182, opened 2020-05-06
GtkSpinner is blue on destructive buttons


Feichtmeier:
![grafik](https://user-images.githubusercontent.com/15329494/81177557-2453bf00-8fa7-11ea-9765-594977d18e93.png)
This is simplescan

Imo it should be white

clobrano:
It makes sense. I wonder though if it's too hard to add a minimal dark border also, so that if it appears over white surfaces it'll be still visible

Feichtmeier:
I would rather special case this only for those situations. Please see the PR

PR #2187, opened 2020-05-08
Special case  spinners in destructive, suggested action buttons

Feichtmeier:
Bright blue spinners on bright green/red buttons looks weird. This can happen in simple-scan.
This special cases those situations and uses selected_fg_color instead, keeps the blue spinner for the rest.
Closes https://github.com/ubuntu/yaru/issues/2182

@clobrano please test this


clobrano:
Not tested yet, sorry, but the change looks good to me

Feichtmeier:
Tested in a VM, works:

Screenshot from 2020-05-25 10-37-14

PR #2198, opened 2020-06-01
Make wacom overlay visible

### LINKED ISSUE #2099: Text in Wacom "Map buttons" is unreadable
Text in Wacom "Map buttons" is unreadable

image

Feichtmeier:
Style .button with %button and use osd colors for the window

Screenshot from 2020-06-01 11-05-35

@madsrh @clobrano they want to fix it for 3.38 I think (looking at their tags), and since I think this bug should be fixed as soon as possible for 20.04 wacom users, let's get this in?

Closes #2099

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -34,7 +34,7 @@ $tooltip_borders_color: $osd_outer_borders_color;
 $shadow_color: transparentize(black, 0.9);
 
 //insensitive state derived colors
-$insensitive_fg_color: mix($fg_color, $bg_color, 50%);
+$insensitive_fg_color: mix($fg_color, $bg_color, 70%);
 $insensitive_bg_color: mix($bg_color, $base_color, 60%);
 $insensitive_borders_color: $borders_color;
 
--- a/gnome-shell/src/gnome-shell-sass/widgets/_osd.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_osd.scss
@@ -30,7 +30,12 @@ $osd_levelbar_height:8px;
 /* Pad OSD */
 .pad-osd-window {
   padding: 32px;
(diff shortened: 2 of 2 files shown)


Feichtmeier:
Ps.: the button at the bottom is an "insensitive" shell button. I am not sure if we should maybe change all insensitive shell buttons to look better, but I have no idea how to test this outside of this dialogue

clobrano:
I agree this shall get it as soon as possible, but I can't test this as well :(

madsrh:
+1 we need this fixed for the LTS

If you have a fix for the "insensitive" button that would be great :)

Feichtmeier:
Btw the weird position is fixed in 3.36 shell and will land in Ubuntu some day

Regarding the insensitive button, this is how they look in the auth dialog
grafik

Feichtmeier:
@madsrh somehow they behave differently than in dialogues and have some strange fog-like stuff overlayed above them
We could try to strengthen the text color a bit for those wacom buttons
grafik
Would this be okay?

Feichtmeier:
Cool, dark shell also works. :+1: 
@clobrano proceed as you wish ^^ I've lost orientation of releases and so on

clobrano:
Merging then :+1:

gnome-shell/src/gnome-shell-sass/_colors.scss
```diff
@@ -34,7 +34,7 @@ $tooltip_borders_color: $osd_outer_borders_color;
 $shadow_color: transparentize(black, 0.9);
 
 //insensitive state derived colors
-$insensitive_fg_color: mix($fg_color, $bg_color, 50%);
+$insensitive_fg_color: mix($fg_color, $bg_color, 70%);
```
Feichtmeier, line 37: Oh man, this happens if you do too much in one time. Did not want to commit this @clobrano , sorry. 
WIll make a PR for this specific line later

Issue #2171, opened 2020-04-29
Icon Request: Difference between iso and img file


avil25:
## Expected Behavior

iso(application/x-cd-image) and img (application/x-raw-disk-image) are almost similar archival formats. They both should have the same or similar type of icon. 

## Actual Behavior
Both do have some connection with cd and the icon of ISO looks good but the IMG icon has the same icon as of plain text.
![Screenshot from 2020-04-29 21-07-25](https://user-images.githubusercontent.com/40562410/80617087-4b6a3800-8a5f-11ea-9da6-d6793011ede7.png)

Also, object file (application/x-object) contains machine code and it does have an icon similar to the document type which looks little misplaced. An Icon like this will look better for binary files.
![application-x-object](https://user-images.githubusercontent.com/40562410/80618282-c08a3d00-8a60-11ea-9825-ed5a73b2adc7.png)

I do not know if it is a good idea but what do you about making mime type icon colorful like this.
![Screenshot from 2020-04-29 21-35-29](https://user-images.githubusercontent.com/40562410/80618817-81a8b700-8a61-11ea-888d-753e925c85f6.png)

I do know all these are non-necessary things and Yaru team does have lots of important work to do but I wonder if someone with a few spare time can do some of the things.

Anyways thank you, everyone, for making Yaru wonderful.

madsrh:
> iso(application/x-cd-image) and img (application/x-raw-disk-image) are almost similar archival formats. They both should have the same or similar type of icon.

Actually they have exactly the same icon 😄 but sure, we could do something similar to the audio files like:

![image](https://user-images.githubusercontent.com/3986894/80621890-4483fb00-8a48-11ea-9ac8-04d9f7ac0562.png)

> ...what do you about making mime type icon colorful like this.

I'm not sure it's a good idea. Any comments @ubuntujaggers @Feichtmeier @clobrano ?

> Also, object file (application/x-object) contains machine code and it does have an icon similar to the document type which looks little misplaced. An Icon like this will look better for binary files.

So, use the binary icon for object files (application/x-object) too?

> Anyways thank you, everyone, for making Yaru wonderful.

😃 Thank you for providing this feedback @avil25. I see you've made [.png mockups](https://user-images.githubusercontent.com/40562410/80618817-81a8b700-8a61-11ea-888d-753e925c85f6.png), but can you create the icons in Inkscape too?

clobrano:
Thanks @avil25 for the feedback, really appreciate it.
[...]

PR #2213, opened 2020-06-21
Added text to img and iso icons

madsrh:
A fix for https://github.com/ubuntu/yaru/issues/2171

I also added the same style icon for squashfs.

image


madsrh:
We could also do something like this: (inspired by [We10x](https://github.com/yeyushengfan258/We10X-icon-theme/blob/master/src/scalable/mimetypes/application-x-cd-image.svg))

image

clobrano:
+1 for the light version :)

clobrano:
@madsrh, what do you think about moving the label few pixels towards the bottom? If you tried already, or just prefer this way, no worries, I think it's already good :+1:

madsrh:
@clobrano I agree. The placement here is simply taken from the mp3 icon, so I knew the text would render sharp. I could try to move them all slightly down, but @ubuntujaggers is really the pixel-fitting ninja. Perhaps he could take a look?

Feichtmeier:
I agree with carlo moving them a bit down is a good idea @madsrh :)

ubuntujaggers:
Is this text low enough:

madsrh:
Perfect 🎉

clobrano:
Yes 👍

ubuntujaggers:
Pushed :+1:

Issue #2183, opened 2020-05-07
nemo: renaming file on icon-view has wrong highlight color


fcastilloec:
## Expected Behavior

I'm trying to rename a file while on icon-view, and the highlighted text (filename) should be a distinctive one.
Here's a GIF on how it looks on list-view, where the highlighted text is working correctly in nemo:
![list](https://user-images.githubusercontent.com/1280530/81338425-7822de80-9061-11ea-95b7-f233a1f467ac.gif)

## Actual Behavior

The highlighted text is not apparent. It feels like nothing was highlighted; if looking carefully, it seems like it's in bold at most.
![icon](https://user-images.githubusercontent.com/1280530/81338455-8244dd00-9061-11ea-9da5-dff03c57e760.gif)

## Steps to Reproduce the Problem

  1. Make sure nemo is set as icon view.
  1. Rename a file (either by pressing F2, or from a menu)

## Software that presents the issue

- Name: nemo
- Version: 4.4.2

## Yaru version

Official version from Ubuntu repositories: `v20.04.6`

Feichtmeier:
We hardly support Nemo, but since there is some Nemo code inside yaru, we could eventually fix it.

rddunphy:
There is a fix for this issue in Nemo, as well as a related issue where the background of split screen panels is shaded incorrectly: https://github.com/linuxmint/nemo/issues/1173. The fallback CSS with the fix is only applied to themes that don't support Nemo, which, if I've understood correctly, it determines by searching for the string "nemo" in the theme's code, so it's not applied to Yaru. Adding [this](https://raw.githubusercontent.com/linuxmint/nemo/master/gresources/nemo-style-fallback-mandatory.css) to `gtk.css` fixes both issues for me. (The same is the case for the default, light, and dark themes.)

clobrano:
thanks @rddunphy!

PR #2218, opened 2020-06-28
Nemo selected text highlighting and inactive pane shading

rddunphy:
Added missing highlighting for selected text when renaming files in Nemo (#2183), as well as shading for the unfocused pane in split pane mode (the right pane in the screenshots). (Based on the [fallback code](https://raw.githubusercontent.com/linuxmint/nemo/master/gresources/nemo-style-fallback-mandatory.css) that implements this in Nemo for unsupported themes.)

Screenshot from 2020-06-28 19-11-54
Screenshot from 2020-06-28 19-12-07

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -419,6 +419,11 @@ terminal-window {
 
     }
 
+    .nemo-inactive-pane .view, .nemo-inactive-pane iconview {
+      // Shades the background of the inactive pane
+      background-color: $bg_color;
+    }
+
     .nemo-window-pane {
       widget.entry {
         // Add border around rename text entry
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -416,6 +416,26 @@ terminal-window {
           &:not(:backdrop):hover { background-image: image($base_hover_color); }
         }
       }
+
+    }
+
+    .nemo-window-pane {
+      widget.entry {
+        // Add border around rename text entry
+        border: 1px solid;
+        border-radius: 3px;
+        color: $fg_color;
+        border-color: $selected_bg_color;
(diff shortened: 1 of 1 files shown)


gtk/src/default/gtk-3.20/_apps.scss
```diff
@@ -416,6 +416,31 @@ terminal-window {
           &:not(:backdrop):hover { background-image: image($base_hover_color); }
         }
       }
+
+    }
+
+    .nemo-inactive-pane .view, .nemo-inactive-pane iconview {
+      // Shades the background of the inactive pane
+      background-color: $bg_color;
```
clobrano, line 424: shouldn't this be `$backdrop_bg_color`?

rddunphy: Yeah, that makes sense, thanks! I've changed it and pushed the commit to this branch, but it's not appearing under the PR - do I need to manually update it somehow?

Feichtmeier: Maybe you pushed it to the wrong branch?

rddunphy: That's definitely the sort of thing I would do, but when I click on the link to the branch at the top of this page it says there are 3 commits. Then if I click on "View pull request" it takes me back here, where there are only 2. Very odd!

clobrano: Weird indeed. You can either let us fix it for you in a next PR, or propose a new PR altogether if you like

Feichtmeier: Maybe it's related to the new UI and a UI bug here. Looked into your repo and saw the three commits.

rddunphy: I did get a couple of 500 errors from GitHub yesterday, so maybe it's just a fluke. I'll try opening a new PR.

PR #2220, opened 2020-06-30
Nemo selected text highlighting and inactive pane shading

rddunphy:
Added missing highlighting for selected text when renaming files in Nemo (#2183), as well as shading for the unfocused pane in split pane mode (the right pane in the screenshots). (Based on the fallback code that implements this in Nemo for unsupported themes.)

(2nd PR, as #2218 didn't update with new commit)

Screenshot from 2020-06-28 19-11-54
Screenshot from 2020-06-28 19-12-07


clobrano:
Thanks @rddunphy!

PR #2229, opened 2020-07-06
Added mimetype icon for log files

madsrh:
Added a mimetype for log files. Inspired by Papirus.
image

Since this is the app icon for the log viewer...
image

...it would make sense to make logs look like this:
image

...but IMO that doesn't really work, so I tried to use the line from the system-monitor app. What do you think? @Feichtmeier @clobrano @ubuntujaggers


Feichtmeier:
I think the pulse is a good design idea for a log file :+1: (untested)

clobrano:
Tested :+1: very good :)

PR #2232, opened 2020-07-06
Gtk3: increase menu visibility

### LINKED ISSUE #2221: Gtk3: context menus could use more shadows
Similar to the shell shadows, I preferred the "Yaru 1.0" shadows. I think @madsrh could remember what I mean.
This is not enough in my opinion and it looks like it's on the same layer as the background:

grafik

Feichtmeier:
_colors.scss: Switch menu background color back to white to let them "pop" out more for the light theme(s), no change for the dark theme

_tweaks.css: Increase the menu shadow for all themes, to make them more visible on the background they are drawn upon

Closes #2221 

@clobrano and/or @madsrh and/or @ubuntujaggers and/or @Jupiter007-43 please test all three gtk3 themes and share your opinion. 
Preview screenshots following but I would really test this on your screens etc etc:
grafik
grafik

### DIFF
--- a/gtk/src/default/gtk-3.20/_colors.scss
+++ b/gtk/src/default/gtk-3.20/_colors.scss
@@ -19,7 +19,7 @@ $link_visited_color: if($variant == 'light', darken($link_color, 20%), lighten($
 $top_hilight: $borders_edge;
 $dark_fill: mix($borders_color, $bg_color, 50%);
 $headerbar_color: if($variant == 'light', lighten($bg_color, 5%), darken($bg_color, 3%));
-$menu_color: if($variant == 'light', $base_color, lighten($inkstone, 1%));
+$menu_color: if($variant == 'light', #FAFAFA, lighten($inkstone, 1%));
 $popover_bg_color: if($variant == 'light', #FAFAFA, lighten($inkstone, 1%));
 $popover_hover_color: transparentize($fg_color, 0.85);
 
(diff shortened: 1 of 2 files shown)


Feichtmeier:
@madsrh @clobrano Maybe a little improvement...1px **less** in Y direction down ?
grafik
grafik

clobrano:
I prefer 1px more 😀
Otherwise it looks like a border, but I'm a fan of big shadows

Feichtmeier:
To sum the telegram chat a bit up:
@clobrano  @madsrh prefer the gray menu
@madsrh prefers 1px y offset
@clobrano prefers 2px y offset
So what should I do? :smile:

clobrano:
Ok for 1px y, it wasn't a big concern for me

Jupiter007-43:
I just test it and I think the shadow looks more natural with just 1px offset. 👍
In fact I have two screen, one with a very bad contrast (on which 2px y is better), but on the second 2px is too much. So I think you should set 1px (which looks good on both screen).
I'm sorry I didn't answer earlier. I wasn't at home yesterday.

Feichtmeier:
Okay, so the final (?) result looks like this:
grafik

madsrh:
This looks good to me. Are you happy with this @Feichtmeier and does it fix your initial issue "increase menu visibility"?

Feichtmeier:
yup :)

PR #2233, opened 2020-07-07
Use CD image icon for apple disk images

wesinator:
DMG files


madsrh:
Looks good to me 😃 

Thanks @wesinator

avil25:
If you can add same style like added in #2213 it will look more consistent.

madsrh:
> If you can add same style like added in #2213 it will look more consistent.

That's a great idea but I'm AFK for vacation and SRU is approaching. So unless you, @Jupiter007-43 or @ubuntujaggers can push this now, I think we'll merge this PR and add the consistent style later.

PR #2244, opened 2020-07-10
Add missing Simple-scan symbolic icons

Jupiter007-43:
I add the x-office-document icon (same as text-x-generic):

Capture d’écran du 2020-07-10 14-00-10

This icon is a mimetype so it was easy for me to know where to put it into source-symbolic. But where should I put the two others (scan-type-adf-symbolic, scan-type-batch-symbolic) which are Scan specific? 🤔 I was thinking into Actions but I don't want to do anything stupid...


Jupiter007-43:
Capture d’écran du 2020-07-11 12-01-15

What do you think of the scan-type-adf icon (second item of the menu).

Jupiter007-43:
I resolved the conflicts.

Issue #2247, opened 2020-07-10
Name issue for mimetype-application-x-rss+xml


Muqtxdir:
Hi,

The same name issue as in #2241 

The svg-filename is **application-rss+xml** instead of **application-x-rss+xml**
The svg-baseplate name is **application-x-rss_xml** instead of **application-x-rss+xml**

**Screenshot:**
![rss](https://user-images.githubusercontent.com/54065734/87199208-a69a8f80-c318-11ea-8778-a6042330fe1d.png)

Thanks!

madsrh:
Nicely spotted again @Muqtxdir 💪 So great to see a new contributor!

Muqtxdir:
Hi @madsrh,

> Nicely spotted again @Muqtxdir

I noticed this when I was inverting the mime-types  to dark (negative) mime-types in Inkscape for [Yaru-blue](https://github.com/muqtxdir/yaru-blue) dark theme.

> So great to see a new contributor!

Thanks! and  btw I admire the work you all have put into Yaru to look this cool

Muqtxdir:
Closed with Pull-request: #2248

PR #2248, opened 2020-07-10
Mimetypes application x rss+xml

Muqtxdir:
Change of icon name in Baseplate from **application-x-rss_xml** to **application-x-rss+xml**
Change of Filename from **application-rss+xml** to **application-x-rss+xml**

Issue No: #2247 
Please check this


ubuntujaggers:
Tested, it works :+1:

PR #2255, opened 2020-07-12
Improving calendar-app-symbolic

ubuntujaggers:
Current calendar symbol is a slightly unsuccessful 16px version of the app icon:

image

This is especially bad if it appears at a larger size in a notification:

image

This suggested PR changes it to the standard calendar appointment symbol used elsewhere:

image

image

Any thoughts/preferences?


Feichtmeier:
I must admit that I like he difference between the app and the "function" ... Hmmm. Let's see what the others say

clobrano:
I like this PR, also it is more in line with upstream approach

Jupiter007-43:
I think the calendar looks better indeed.

ubuntujaggers:
Any objections to merging?

PR #2260, opened 2020-07-17
Add display Settings icons

### LINKED ISSUE #2257: Missing display settings icons
## Actual Behavior

Capture d’écran du 2020-07-15 19-43-56

😱 In addition Adwaita's ones look blurry...

I've already test to replace one icon:

Capture d’écran du 2020-07-15 19-16-06

Icons names are: **video-joined-displays-symbolic**, **view-mirror-symbolic** and **video-single-display-symbolic**.
These are 24px icons.

Jupiter007-43:
Closes #2257


Jupiter007-43:
Which one do you prefer?

With a number like the original icon:
Capture d’écran du 2020-07-17 20-19-40

Or without:
87817307-05a45b00-c869-11ea-88a0-4ee7f2b8f3f2

Feichtmeier:
Personally I think I prefer the one without the "1".
How does it look when you cycle through the options with Windowskey+P ?
You can make a delayed screenshot to catch it (Flameshot->indicator->open launcher)

Jupiter007-43:
I prefer without the number too.

Jupiter007-43:
Apparently not, the text of the two last options on you screenshot is the same on mine (only external and only internal) although I don't have a built-in screen.

So:

- The icon for the "only external" option is the same as the "single monitor" option (video-single-display-symbolic).
- The icon for the "only internal" option use the "computer-symbolic" icon (a screen with a keyboard in front) so we cannot modify it without alter the meaning of it.
- And I think that the two others (join and clone options) icons look good (at a pinch perhaps "join" can be modified if you want).

In short, I can only adapt the "video-single-display-symbolic" icon. Have you got an idea for it (not having a laptop, I don't know what to do)?

Thanks 🙂

Jupiter007-43:
Is your screen settings page looks like this on your laptop?

87817307-05a45b00-c869-11ea-88a0-4ee7f2b8f3f2

Join screen | Clone | Single screen

Feichtmeier:
Thanks for the clarification. I hope you see this as problematic as I do ^^ 
Don't want to lay any stones in the way but I always thought that this windows+P cycling has a very high chance to be misunderstood.
computer-symbolic should indeed be as generic as possible. Uhhh I need to think
@madsrh any idea?

Feichtmeier:
Maybe include little cable for the external display icon?

Jupiter007-43:
Very good idea! Do you think of something like this:
Capture d’écran du 2020-07-20 16-04-05
Capture d’écran du 2020-07-20 16-05-49

Jupiter007-43:
For comparison, Adwaita icons:

Capture d’écran du 2020-07-20 16-11-00

I think the two last options are totally unintelligible... 😳 And what's the last icon supposed to represent?

Jupiter007-43:
Small adjustment: remove cable rounded end path.

Capture d’écran du 2020-07-20 16-24-29

Jupiter007-43:
Final version:

Capture d’écran du 2020-07-20 17-02-33

Jupiter007-43:
Revert view-mirror icon to upstream-like style:
Capture d’écran du 2020-07-20 17-48-46

Feichtmeier:
Tested the branch, too!
Looks very cool @Jupiter007-43 - and I agree with @ubuntujaggers

madsrh:
I did not test, but the PR looks good and I also agree with @ubuntujaggers **BUT** IMHO the metaphor isn't very good (_not your fault_). 
image I wish upstream would adopt [this one](https://raw.githubusercontent.com/ubuntu/yaru/c48a600fffb783aa849675ac29f4a3bd54cd9e78/icons/Suru/scalable/status/view-mirror-symbolic.svg) ❤️ 

image Using the butterfly / mirror icon could also lead to the assumption that you would get this 😄 
image

Jupiter007-43:
I completely agree with all of you. This time I think we're good.

> image Using the butterfly / mirror icon could also lead to the assumption that you would get this smile
> image

Unity... 😢
Hum... The mirror icon is completely inappropriate indeed. Strange screen config by the way, I didn't know it was possible to doing that.

So everyone is okay with this version:
Capture d’écran du 2020-07-20 17-48-46

Amr-Ibra:
Great icons, thanks @Jupiter007-43.

It took me some time to recognize, without reading the text, that the left icon, video-joined-displays-symbolic, is actually two monitors next to each other horizontally. I was looking at it vertically. Does anyone else think the same?

Jupiter007-43:
Hi @Amr-Ibra, thank you!
Sorry, but what do you mean with _I was looking at it vertically_ (sorry I'm French 😥)?
Do you mean that the two screens should be one on top of the other?

Amr-Ibra:
> Sorry, but what do you mean with _I was looking at it vertically_ (sorry I'm French disappointed_relieved)?

I meant that I was trying to recognize what the icon meant by looking at it from top to bottom and vice versa, not from left to right. So I couldn't see the two screen from the first glance.

> Do you mean that the two screens should be one on top of the other?

No, they are fine next to each other, but maybe I couldn't recognize it because they are two half monitors.

Jupiter007-43:
> I meant that I was trying to recognize what the icon meant by looking at it from top to bottom and vice versa, not from left to right. So I couldn't see the two screen from the first glance.

👍 Ok, sorry I didn't understand.

I can try something like the upstream one:
Capture d’écran du 2020-07-15 19-43-56

Jupiter007-43:
Capture d’écran du 2020-07-26 20-17-39

What do you think of?

In fact I've already tried something like this before but I preferred the current one because like that, this icon is 4px smaller in  height than the two others.

Amr-Ibra:
It's more recognizable, but you are right, it looks smaller than the other two icons, and the right tall monitor looks weird.

Jupiter007-43:
In fact I had the same thinking as you with the two half screen but it was my best idea so... (@ubuntujaggers have you a best one?)
And I really don't like the upstream icon with a little and a big screen. 😖

But thank very much you for your feedback 👍

Jupiter007-43:
And what do you think of this @Amr-Ibra?

1. Capture d’écran du 2020-07-27 11-45-03
2. Capture d’écran du 2020-07-27 11-47-41
3. Capture d’écran du 2020-07-27 11-48-38

Amr-Ibra:
Number 3 is the best among them, but I actually prefer the one before, without the ubuntu logo. Besides, Yaru icons would be used outside Ubuntu.

Jupiter007-43:
I agree with you, it's just for improve readability. Do you think that it works?
If yes I will try with another logo (even if I don't know which one to use yet).

Amr-Ibra:
> I agree with you, it's just for improve readability. Do you think that it works?

Yes, it improves readability.

Jupiter007-43:
Hi @Amr-Ibra,
please take a look at #2300.

Issue #2251, opened 2020-07-11
application-apk icon is misleading


madsrh:
Since the Android APK mime type icon uses the zipper, it is associated with a compressed file.
IMHO it would be more correct to use the package metaphor (although a package is also an archive).

This issue is just a reminder to revisit the icon and align the design with Flatpacks, Snaps and the likes, once some design guidelines are in place.

Feichtmeier:
I think those archives for packages are generally a good idea. But I agree that one should understand what's inside. How does it currently look compared to .jar or .Deb ?

avil25:
What do you think about something like this
![application-apk-2](https://user-images.githubusercontent.com/40562410/87229624-2ed76e00-c3c7-11ea-86ea-8030db6e204c.png)

or something similar to the one we already have for .deb files
![application-apk](https://user-images.githubusercontent.com/40562410/87229637-54647780-c3c7-11ea-9e33-6b2457fcde3f.png)

Sorry I am not really good at designing but trying to give some ideas.

Jar icon is similar to the one we have for the archive. Snap and flatpacks have a simple file icon with there logo in the middle.
similarly, windows icon also look misleading and unrecognizable(but I think we can't use the windows logo that's why it's like that)

madsrh:
I personally prefer your second suggestion.

> similarly, windows icon also look misleading and unrecognizable(but I think we can't use the windows logo that's why it's like that)

We cannot use the official Windows logo

ubuntujaggers:
That little robot in the bottom right is going to be indistinguishable at normal sizes, but I like the concepts :+1:

Muqtxdir:
Hi,
What about this, too simple and boring right?
![apk](https://user-images.githubusercontent.com/54065734/87244317-f896fe80-c459-11ea-8a3c-3ea1b33a228c.png)

Thanks!

Feichtmeier:
The thing what I like about that Zipper metaphor, is that it correctly says what that file is. 
A jar can be run by a java runtime but it also contains other files, compressed. If you remove the zipper from the icon this metaphor goes away

madsrh:
This one also looks good @Muqtxdir 😄

> The thing what I like about that Zipper metaphor, is that it correctly says what that file is...

IMHO we should not merge a new design before we've decided on some kind of guidelines. I will try to come up with a suggestion soon.

And thank you so much for contributing! We love feedback but we lloooooovvvveeee contributions ♥️

avil25:
[...]

PR #2269, opened 2020-07-21
Updated snap mimetype icon with a more colorful version

madsrh:
Current:
image

Is more colorful better?

image


Feichtmeier:
What's inside that snap file?
Binary or text?

madsrh:
> What's inside that snap file?
> Binary or text?

You're asking me? 😨 

I simply ran with the comment from @clobrano [here ](https://github.com/ubuntu/yaru/issues/2251#issuecomment-658113935) 😄

clobrano:
It's a compressed filesystem with binaries

Feichtmeier:
Okay, I really like the colourful mimetypes.
I would leave mimetype icons for files which include text of any kind (XML, txt, Java code, whatever code, yaml.... ) With a white background though, so you still have the association with the piece of paper - wdyt?

madsrh:
> I would leave mimetype icons for files which include text of any kind (XML, txt, Java code, whatever code, yaml.... ) With a white background though, so you still have the association with the piece of paper - wdyt?

Exactly my thinking! 🤘

ubuntujaggers:
Are we happy to merge this one? I like it 👍

PR #2274, opened 2020-07-22
New icon for ms-dos-executable mimetype

madsrh:
This PR might be going straight to hell 🔥 👿 because this clearly violates Microsofts guidelines on page 7 here: https://query.prod.cms.rt.microsoft.com/cms/api/am/binary/RE1voQq 🤷‍♂️ 

**EDIT:** _Okay, I see now that I've committed the wrong files_ 🤦‍♂️ _I'm going to sleep now!_ 

image


Jupiter007-43:
It looks very good!
🤔 But it is possible to add something to show that is an executable and not Windows into a file?
If not never mind, it's still much better than the old icon. 👍

Feichtmeier:
Lil :gear: maybe?

Jupiter007-43:
> Lil gear maybe?

Yes something like that 👍

madsrh:
> But it is possible to add something to show that is an executable and not Windows into a file?

IMO that isn't necessary. With a solid background these follow the same style as snaps and Flatpaks. Do you have an idea for the gear?

Jupiter007-43:
Yes indeed, but Snaps and Flatpacks are both packages while Windows is an OS (sorry to insist like that 😇).
I cannot try now but I will test something with a gear shortly later...

madsrh:
We could also adopt (or Yaru style adapt) the Windows icon for exe files that have no icon. A quick mockup:

image

Jupiter007-43:
> We could also adopt (or Yaru style adapt) the Windows icon for exe files that have no icon.

Good idea 👍
But it really need to be adapted, because it is a bit ugly and will looks inconsistent (even in Windows 😅).

madsrh:
image

Feichtmeier:
For Linux binary runnable files we have that gear icon. I think using something totally different for .exe files would be maybe not fitting. Even though the idea is great

madsrh:
> using something totally different for .exe files would be maybe not fitting. Even though the idea is great

I 100% agree, but I had to try :)

> ...Yes indeed, but Snaps and Flatpacks are both packages while Windows is an OS 

I don't agree that the gear is fitting here, but perhaps you can convince me with a great icon.

One could argue that these are "packages" for Wine.
.RPM files are from a different OS too (although not binary).

Feichtmeier:
grafik

That's application-x-executable
It's a generic icon, for something very generic - any binary, runnable. 
What if we just put a little windows emblem somewhere there?

Jupiter007-43:
> image

The icon looks good, but as @Feichtmeier said it doesn't fitting. And also, people who are not from Windows might not understand the meaning. Myself, who hasn't used .exe on Windows since I was three years old, I find it confusing...
But thanks for trying @madsrh and sorry. 🙂

> That's application-x-executable
> It's a generic icon, for something very generic - any binary, runnable.
> What if we just put a little windows emblem somewhere there?

That's exactly what I wanted to do, but I didn't have time yesterday. I'll do it this morning.

Jupiter007-43:
Some ideas (very quickly realized):

Capture d’écran du 2020-07-24 12-29-30
Capture d’écran du 2020-07-24 13-13-41
Capture d’écran du 2020-07-24 13-14-34
Capture d’écran du 2020-07-24 13-31-11
Capture d’écran du 2020-07-24 13-34-03

Jupiter007-43:
But if we don't have right to use Windows logo, I think that the first image will be the better choice because it show that is an executable which is not like the others from the rest of the system (and it'll be easier to shrink than the others).

🤔 Is wine installed by default on Ubuntu?

Feichtmeier:
If we could use the logo, I'd go with the last icon. Otherwise the first :D

Jupiter007-43:
I'm of the same opinion as you @Feichtmeier.

> Edit: no, wine is not installed by default

As said @madsrh we could have eventually considering that .exe are Wine packages, but if Wine isn't by installed default I think it is not a good idea.

Jupiter007-43:
And what about adding "EXE" text below the gear?

Capture d’écran du 2020-07-24 15-14-24

clobrano:
Smart idea! :smile:

Feichtmeier:
Good idea. It looks a little unbalanced on the square baseplate, but the idea is nice

Muqtxdir:
I think **most** mime types use the vertical-oblong rectangle shape as baseplate
based on that my quick mockup (I tried 😅)
exe

Jupiter007-43:
The icon looks very good @Muqtxdir, but I think we should avoid to use anything that might remind a Microsoft branding. 😓
In fact I used the square baseplate (and the gear) to fit with **application-x-executable** icon because an .exe is a binary executable and not a package like .deb.
But this is just my opinion... 😉

Muqtxdir:
Yes, the gear-icon is the way to go ( sorry I didn't see the whole conversation and posted the mock-up )

And I agree with your opinion

Issue #2271, opened 2020-07-22
Guidelines for mimetype icons for packages, archives and more...


madsrh:
**Hi**
Related to the discussion in #2251 

The outcome might just be to leave everything as it is, but if you have any thought please leave them below @ubuntujaggers @Feichtmeier @clobrano @Jupiter007-43 @avil25

My questions on the current icons:
- The Flatpak pictogram is monochrome. Should they all be this way? Or should we add color to the Flatpak icon?
- Is that a Windows logo? Why not change it like Eaglersdeveloper [suggested](https://github.com/snwh/suru-icon-theme/issues/65#issuecomment-399678431)?
- Because the Window files are binary, should they have a solid background too or...?
- Should all of these icons really be packages?
- Or just leave everything as it is 👌 

I have to admit that I'm in a bit over my head here with all the file types, so please be patient 😅 

![image](https://user-images.githubusercontent.com/3986894/88183034-3ac9f800-cc31-11ea-9c27-a088a7166b05.png)

This looks pretty good, but didn't we already add text for zip, rar, 7z? 🤔 
(doh, a package icon got included here) 😄 
![image](https://user-images.githubusercontent.com/3986894/88183082-4e755e80-cc31-11ea-939c-6aec7ccc2809.png)

Images files. Should these have a background color? 
![image](https://user-images.githubusercontent.com/3986894/88183168-6947d300-cc31-11ea-9f8d-bee4bcf34480.png)

There's a few more files we don't have yet which is for VirtualBox. Here are the icons from Suru++
![image](https://user-images.githubusercontent.com/3986894/88185447-3bb05900-cc34-11ea-98e9-adb560552c88.png)

clobrano:
It seems a big work indeed 😀

It seems hard to draw a precise line between mimetypes right? Some are just archive, but others are both archive and packages.

Jupiter007-43:
My opinion:
- It should be great to have color pictogram for Flatpack, to fit with the others.
- The actual W$ logo is very confusing. The Eaglersdeveloper suggestion looks better (maybe add something to show that .exe are executable, otherwise it looks like that file was just... Windows into a file 😈), but do we have rights to put this branding logo here?
- Images files look good, the disk pictogram is very visible and it's fine like that. 👍
- As a Virtualbox user I say yes to add these mimetype icons (reworked for Yaru of course).

I hope this will be useful for you.
Good luck 🍀, if you need help don't hesitated. 🙂

[...]

PR #2275, opened 2020-07-23
Replace disk image symbolic pictograms by fullcolor ones

Jupiter007-43:
Related to #2271

Capture d’écran du 2020-07-23 10-43-09


madsrh:
Look good to me 👍

ubuntujaggers:
Interesting idea :+1: I personally would reduce the strength of the disc outline on the smaller icons, we don't normally use edging that strong inside the icon itself (although we sometimes do around the icon edge).  What do others think?

Feichtmeier:
After you've said that this icon is the first mimetype that doesn't use the watermark I actually agree that we should revert the look to a more watermark icon. Though I don't know how we could apply this

madsrh:
By "watermark" you mean monochrome?

Feichtmeier:
No, less 3d. Like the other mimetypes :)

Jupiter007-43:
I reduce small size shadows:

Capture d’écran du 2020-07-28 20-32-28

Jupiter007-43:
Also remove the big size shadow:

Capture d’écran du 2020-07-28 20-39-19

madsrh:
I wonder if flatter would be better?

image

Feichtmeier:
Yes and then add the fold back I think. Then we'll get closer to the other mimetypes

Jupiter007-43:
I did a flattest version:

Capture d’écran du 2020-07-29 10-41-30

I think we should keep reflections to avoid any confusion with other things.

Feichtmeier:
That's cool. Could you try to add the folds back?

Jupiter007-43:
With the folds:

Capture d’écran du 2020-07-30 10-37-07

madsrh:
Let's merge the one without the folds, right @Feichtmeier ?

Jupiter007-43:
Maybe just add the corner:

PR #2277, opened 2020-07-24
Rework render-bitmaps script

clobrano:


clobrano:
Hey @Jupiter007-43, this should fix the render-bitmaps script. If you could test in your environment it would be great :)

Jupiter007-43:
It works with Inkscape 1.0 from snap. Thanks very much @clobrano !

clobrano:
I think we can merge this. If there's a need for work with inkscape version < 1.0 I can add a switch to use the old code.

PR #2273, opened 2020-07-22
Added mimetype icon for dart files

madsrh:
Closes #2271 

Currently **.dart** isn't recognized, but if support is added the icon is there - just [like we saw with the .nim icon](https://github.com/ubuntu/yaru/pull/1888#issuecomment-587597448).

image

This PR contains:
image

PR #2280, opened 2020-07-24
Added mimetype icon for Docker

madsrh:
Like the mimetype icon for Dart in #2273, support for this file type is still missing in Ubuntu. Hopefully @didrocks can help here :angel: 

image


Feichtmeier:
Looks really good. Just one question, is this for docker files, that describe where to download the docker image and so on, or is this for the actual container images?

madsrh:
Ahh, I see where you're going. I'm thinking this is the container image, so it should probably look like the other image icons, right?

Feichtmeier:
We still don't have a guideline for container images, but maybe adapt this one to your new snap mimetype, which does not havea white background

ubuntujaggers:
Brand for this one does seem to be blue on white:

image

So might be fair?

ubuntujaggers:
Are we happy to merge this? 👍

madsrh:
> Are we happy to merge this? 👍

I am 😄 We can just merge this now althought we haven't heard back from @didrocks

PR #2289, opened 2020-07-27
Add GPX mimetype

### LINKED ISSUE #2288: Missing mimetype icon for GPX
Capture d’écran du 2020-07-27 18-41-04

> GPX, or GPS Exchange Format, is an XML schema designed as a common GPS data format for software applications. It can be used to describe waypoints, tracks, and routes.

I think we should just symlink to **application-geo+json**:

Capture d’écran du 2020-07-27 18-41-25

Jupiter007-43:
Just link application-geo+json to application-gpx+xml

Capture d’écran du 2020-07-27 18-45-25

Closes #2288

### DIFF
--- a/icons/Suru/16x16/mimetypes/application-gpx+xml.png
+++ b/icons/Suru/16x16/mimetypes/application-gpx+xml.png
@@ -1,0 +1,1 @@
+application-geo+json.png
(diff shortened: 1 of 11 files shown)


Jupiter007-43:
> Nice spot.

In fact this file is in my folder since several month without icon and I just notice it!

PR #2291, opened 2020-07-27
Add pgp-signature mimetype

### LINKED ISSUE #2290: Missing mimetype icon for Pgp-signature
Capture d’écran du 2020-07-27 18-52-10

I think we should just symlink to **application-pgp-keys** icon.

Jupiter007-43:
Just link application-pgp-keys to application-pgp-signature.

Capture d’écran du 2020-07-27 19-00-13

Closes #2290


madsrh:
This is fine but I wonder if we should use the symbolic key icon here  (still with the text ofc) 🙄 Just a thought

Jupiter007-43:
> This is fine but I wonder if we should use the symbolic key icon here (still with the text ofc) roll_eyes Just a thought

That's also that I think now, as the name is **signature** 😥
As an excuse, I wanted to hurry up and go for a bike ride 😇

Jupiter007-43:
PGP signature icon on wikipedia:

pgp-signature

Jupiter007-43:
What do you think of something like this (sorry for the very bad signature I will rework it):

Capture d’écran du 2020-07-28 15-02-53

madsrh:
> PGP signature icon on wikipedia:

Okay, [that's](https://github.com/ubuntu/yaru/pull/2291#issuecomment-664599790) totally wrong IMHO. A shield would work better as a metaphor for encrypted IMO.

This discussion really shouldn't be had in a PR but since we're already started, it's probably best to keep everything in the same place 😄 

We really need:

- **signature**
- **keys**
- **encrypted** 

To be honest I'm not really a fan of the current lock 🔒  icon, so if you've got a better design idea there, that would be great. As mentioned, we do have a Yaru style key in the symbolic icons, but it's not great (just my opinion)
image



Other icon themes use this one for _signature_ but it would be better if all three could be drawn in the same style:
image

More random images from the internet as inspiration 😆 
image
image
image


image

image

madsrh:
> What do you think of something like this (sorry for the very bad signature I will rework it):

Yes, that could totally work. They should all just be drawn in a similar style

Jupiter007-43:
> They should all just be drawn in a similar style

You mean that all PGP icons should use the same style, am I right?

madsrh:
> You mean that all PGP icons should use the same style, am I right?

Exactly, your signature suggestion is different from the current [symbolic pgp icon](https://user-images.githubusercontent.com/36476595/88570190-f2994400-d03b-11ea-9eba-ba3556536c66.png)

Jupiter007-43:
🤔 Do you prefer a symbolic pen icon or a fullcolor lock icon?

madsrh:
> Do you prefer a symbolic pen icon or a fullcolor lock icon

I'm not sure 🙈 I do prefer your full colored pen, but it could also be mistaken for a drawing app icon (like Krita, Gimp,...). Vague reply, sorry

Jupiter007-43:
A fullcolor lock:

Capture d’écran du 2020-07-28 15-45-39

> I'm not sure see_no_evil I do prefer your full colored pen, but it could also be mistaken for a drawing app icon (like Krita, Gimp,...). Vague reply, sorry

Since it's marked _pgp_ below, I think there's no problem 😉

Jupiter007-43:
In fact I don't know how to do a good signature. Actually it really looks like a draw software icon. 😅

madsrh:
I like the lock with colors. So simple 😃 

> In fact I don't know how to do a good signature. Actually it really looks like a draw software icon. 😅

You can do it 💪 
Perhaps use text-x-generic as a base and add a scribble signature and a pen. If it's your pen that looks too slick, you could try using the Yaru pencil from accessories-text-editor ✏️ 

There's [a ton of inspiration](shorturl.at/fGNRW) but you're on the right track
image

Jupiter007-43:
I reworked the signature:

Capture d’écran du 2020-07-28 21-05-06


> If it's your pen that looks too slick, you could try using the Yaru pencil from accessories-text-editor pencil2

Thanks, but I don't think we can sign with a paper pencil 😉

Another proposal:

Capture d’écran du 2020-07-28 21-10-37

By the way @Feichtmeier @ubuntujaggers can you tell me if the pen and the lock icon are Yaruish enough?

ubuntujaggers:
> By the way @Feichtmeier @ubuntujaggers can you tell me if the pen and the lock icon are Yaruish enough?

Absolutely  :+1: 

I like the pen concept - it just depends on the execution working at all sizes.  It might require some changes to shine in the smaller icons (e.g., making the pen a bit fatter; making it 45 degrees; making the signature a less complex path).  But definitely a nice idea to start with :)

Very minor point: because the pen is a floating object I would prefer the folds to go _below_ it.  Having the folds on top of the pen makes it look a bit like a printout of a picture, rather than a physical pen?

Jupiter007-43:
And what about just a signature without the pen:

Capture d’écran du 2020-07-30 21-11-13

Jupiter007-43:
A screenshot of application-pgp-keys icons:

Capture d’écran du 2020-07-30 22-09-43

Feichtmeier:
I think the last one with the lock is the one that users can understand very fast + it looks nice!

Jupiter007-43:
And a screenshot of application-pgp-signature icons:

Capture d’écran du 2020-07-30 23-36-54

Jupiter007-43:
> I think the last one with the lock is the one that users can understand very fast + it looks nice!

@Feichtmeier in fact the icon with a lock pictogram is for "application-pgp-keys". I rework it to rest consistent with the new "application-pgp-signature" icon. But I'm glad you like it 😌 (and what do you think of this one upper?).

ubuntujaggers:
Beautiful!! Is this ready to merge from your PoV @jupiter007-43?

madsrh:
Both icons look great @Jupiter007-43 
As mentioned we need three pgp icons (signature, keys and encrypted) so I would have loved if there was somehow visually connected (like having the PGP text on all icons) but I don't see that working here. Perhaps removing the PGP from the lock? 🤔 


What do you think about making the inside of the lock darker and the yellow darker so it has better contrast against the white background? 🤷‍♂️  Just a thought
image

Jupiter007-43:
> As mentioned we need three pgp icons (signature, keys and encrypted) so I would have loved if there was somehow visually connected (like having the PGP text on all icons) but I don't see that working here. Perhaps removing the PGP from the lock? thinking

Sorry @madsrh, with my first search on a mimetype list I didn't noticed pgp-encrypted 😓 (the list had to be truncated). So I will also add it. 👍

> I would have loved if there was somehow visually connected (like having the PGP text on all icons) but I don't see that working here. Perhaps removing the PGP from the lock? thinking

I really prefer keeping the text, so here is another proposal:

Capture d’écran du 2020-08-01 11-48-19

> What do you think about making the inside of the lock dark and the yellow darker so it has better contrast against the white background? man_shrugging Just a thought

Really good idea! The contrast is indeed bad with this light yellow.

madsrh:
That looks good @Jupiter007-43 👌

Jupiter007-43:
Great!
I will use the lock for the "pgp-encrypted" mimetype, and I'll create a key pictogram for the actual "pgp-keys" mimetype.

Jupiter007-43:
Proposal for pgp-keys:

Capture d’écran du 2020-08-01 14-20-03

madsrh:
Wow, nice drawing 😃 
I have two thoughts and just ignore this comment if you disagree, but the black ring/keychain jumps out at me, so maybe something lighter? Maybe different colors for the keys would be better at small sizes? 🤷‍♂️ 

If you push the keys icon, perhaps @ubuntujaggers can review these? Great work @Jupiter007-43 💪 

Not very well executed, but just a mockup:
image

ubuntujaggers:
Yep, push the latest keys and I'll review 👍

Jupiter007-43:
Et voila!

Capture d’écran du 2020-08-02 02-13-59

Jupiter007-43:
If you think that the colors aren't good, don't hesitate to tell me 😉

madsrh:
> If you think that the colors aren't good, don't hesitate to tell me

Perhaps they do look a little pale compared to the other icons 😅

Jupiter007-43:
It is better like this?

Capture d’écran du 2020-08-02 02-33-35

madsrh:
I was thinking something much more colorful BUT I'm totally happy with this one

Jupiter007-43:
I'm ok for adding more colors but I really don't know which one use 😅

Jupiter007-43:
🤔 Are you thinking of a flashier yellow?

Capture d’écran du 2020-08-02 12-44-03

I used the same color as the lock.

madsrh:
Damn, you're were too fast 😄 
I tried to take colors from some apps, but using the lock color looks great and makes perfect sense 👍 Could you push that one? 

image

Jupiter007-43:
The blue and purple looks pretty fun!

ubuntujaggers:
I like the classic key colours in the "flashier yellow" one 👍

PR #2293, opened 2020-07-28
Executable mimetypes based on upstream

ubuntujaggers:
Give this a whirl and see what you think...

**application-x-executable**

image

image

image

image

**Blue version for application-x-ms-dos-executable:**

image

image


Feichtmeier:
I think it looks great :+1: 
Especially when you realize the special case of this Mimetype that is also put in the dock and thus would be more of an app icon. So imo this is a good and well executed compromise between mimetype and app icon :+1:

ubuntujaggers:
LOL @ gear wars :rofl:

Jupiter007-43:
I love these icons 😍

madsrh:
This works great @ubuntujaggers 🎉 
My only concern is that it looks a bit "heavy" next to the regular document files. I wonder if something like the below could work where the background is white like document files 🤷‍♂️ What do you think?

exe

Feichtmeier:
That's what I meant with compromise. It's actually a mimetype. Yet it appears on the dock. So either optimize it for the dock or optimize it for the appearance of a mimetype

madsrh:
>  So either optimize it for the dock or optimize it for the appearance of a mimetype

I understand what you're saying @Feichtmeier 👍  I think the light version works fine in the dock too
image

Feichtmeier:
Okay!

just leaving info here, this is another place where the icon can appear:
grafik

ubuntujaggers:
Devil's advocate: would it not be just as heavy if you had a folder with a small number of text documents outnumbered by lots of these:

application-photoshop

Or these:

application-mspowerpoint

I think if you have a folder that's mostly all the same non-white mimetype then any white ones will look a bit faint, but we've (rightly I think) decided to not have all white ones?

ubuntujaggers:
Alternatively, how about a colour instead of grey?

madsrh:
> Devil's advocate...

😄 

I get your point. I'll be testing this for a while, but ATM I still prefer the light background.

> Alternatively, how about a colour instead of grey?

I really couldn't imagine that being better, but do you have any suggestion?

ubuntujaggers:
Hang on, let me try light, if everyone likes that it's an easy solution :+1:

madsrh:
>  if everyone likes that it's an easy solution 👍

Might be just me 😄

clobrano:
I do like the lighter version, but I wonder whether the "heavier" one just needs some time to accept it :)

ubuntujaggers:
image

image

image

image

Must confess I prefer the grey version - I'm quite wedded to this issue because I do live with it on my launcher and app grid - but the white version is also quite nice.

I do think we have quite a few mimetypes that would outshine the paper ones in a scenario like the above - especially video, which is almost black:

image

Of all the mimetypes, this one has the best excuse (because there are multiple scenarios where it's used as an app icon) so it seems unnecessary to make it balanced with a plain document icon - but happy to go with majority design team vote :+1:

ubuntujaggers:
> I do like the lighter version, but I wonder whether the "heavier" one just needs some time to accept it, and get used to it :)

Might be the same for me and the light one! I'll leave it on my launcher and see if I grow to love it.

ubuntujaggers:
I can live with white 👍

ubuntujaggers:
Okay, current includes the following:

image

image

image

image

Can you try it and let me know if you're happy, @madsrh @Feichtmeier @clobrano ?

clobrano:
Gorgeous icons for me :+1:

Feichtmeier:
Looks great! May the story of the gear wars come to an end :drum: :dancers:

Jupiter007-43:
Tested, they look very good!

ubuntujaggers:
Only @madsrh left to try his heavy folder :)

madsrh:
This is superb @ubuntujaggers 🎉🥰 


> Must confess I prefer the grey version - I'm quite wedded to this issue...
> ...but happy to go with majority design team vote 👍

So @ubuntujaggers are you still on the dark side and just going with the majority here? (towards the light)

The old icon is a sore thumb in the beautiful icon theme, so someone please merge this!!

madsrh:
I'm totally loving this new look 🤩 
...but looks like we've missed the **application-x-msi** files, right?


image

ubuntujaggers:
In a word, yes ☺️ What is it, symlink or another colour?

madsrh:
> What is it, symlink or another colour?

That was what I meant to ask 😄 
IMO just a symlink. The user can see that these are Microsoft files, that should be sufficient I think.

PR #2297, opened 2020-07-29
Rounding some more symbolic icons

### LINKED ISSUE #2296: Rounding some more symbolic icons
I propose to rounding the corners of these icons to stay consistent with the other icons from #2283, like this:

mail-drafts and mail-drafts-rtl:

Capture d’écran du 2020-07-29 15-25-18

insert-image:

Capture d’écran du 2020-07-29 15-26-01

document-new:

Capture d’écran du 2020-07-29 15-26-21

and document-page-setup:

Capture d’écran du 2020-07-29 15-27-04

What do you think of @ubuntujaggers?

Jupiter007-43:
Closes #2296


ubuntujaggers:
Beautiful 🥰

PR #2300, opened 2020-07-30
Add a triangle to single and join display icons to improve readability

Jupiter007-43:
Link to #2260

Capture d’écran du 2020-07-30 16-02-18


ubuntujaggers:
You're a magician!  Why is it a draft...?

Jupiter007-43:
Thanks!

> Why is it a draft...?

Because I would like check something before and I forgot to mark it as Ready for review.

PR #2304, opened 2020-08-02
Add fullcolor pictogram on application-x-gameboy-rom

Jupiter007-43:
A proposal for application-x-gameboy-rom:

Capture d’écran du 2020-08-02 15-49-27

I kept the white background, just put the controller icon without the shadow and added a cable.

What do you think of @madsrh?


madsrh:
🤔 Yes, maybe. I don't know. I'll test this PR locally for a while and see how it feels.

Perhaps emphasize that this is a rom? Some icon themes just symlink to x-executable so I was thinking something like this?
image

Maybe they are too similar or maybe it's dumb that there's no reference to the game part 🤷‍♂️  I really don't know what I think atm, so I better sleep on it.

Jupiter007-43:
Some more ideas:

Capture d’écran du 2020-08-03 12-37-42
Capture d’écran du 2020-08-03 13-25-30
Capture d’écran du 2020-08-03 13-32-52

I find my first proposal a little boring, but I didn't have more ideas.

Feichtmeier:
@Jupiter007-43 @madsrh 

I don't think that the runnable file metaphor is right here.
This rom is more like an ISO/Image :)

Jupiter007-43:
> I don't think that the runnable file metaphor is right here.
> This rom is more like an ISO/Image :)

Yes indeed, it's an image of a cartridge!

Feichtmeier:
Maybe symbolise some those big game sticks you stick into the console? Like

Feichtmeier:
Or if the icon is only for game boy ROMs...

madsrh:
I don't think we need target every console, because there's many as you can see here: https://github.com/hlechner/hl-numix-mime/issues/2

Keep in mind that there's also files for devices that doesn't use cartridge. Like c64, Amiga...

Feichtmeier:
Ah, but the mimetype is named application-x-gameboy-rom, so I thought it would be specifically for game boy roms?
Maybe add only dedicated mimetypes for very popular roms like nes, snes, gameboy, etc and for all the other

a generic icon maybe only add the control cross and the 4 buttons on top of a white paper?

Jupiter007-43:
The gameboy icon is actually symlinked to every others roms icons, so it is actually like a generic icon.
If you want, I can create icons for popular roms. But if we do that, I think we should firstly create a real generic rom icon and link it to the others, then create dedicated roms icons.

The choice is yours 😉

> _plzdontkillme_

🤔 I'll think about it 😈

Jupiter007-43:
IMO this kind one should be a good generic icon (I mean just the buttons on the white background like you said @Feichtmeier):

Capture d’écran du 2020-08-03 13-32-52

But which shape should be used for all rom icons, square or vertical oblong?

Feichtmeier:
To the generic one: I would not cut the edges and rather put the cross and the buttons just in the middle of the baseplate, like most of the mimetypes.

Jupiter007-43:
Just like this?

Capture d’écran du 2020-08-03 20-49-24

Feichtmeier:
I don't know :shrug: Just :brain: :partly_sunny: xD 
Looks a bit empty  I guess even with the folds?, what if you make the whole thing look like a game plug? With the bottom with some chip leaking out? :see_no_evil: 
grafik

Jupiter007-43:
> Looks a bit empty I guess even with the folds?

I didn't noticed I hidden the folds layer, it wasn't on purpose sorry. 😓

> Looks a bit empty I guess even with the folds?, what if you make the whole thing look like a game plug? With the bottom with some chip leaking out?

👍 Good idea, I'll try, I think this will be less boring and empty than that I did until now...

Jupiter007-43:
Cartridge icon (just ignore the triforce logo 😉):

Capture d’écran du 2020-08-04 15-32-07

Feichtmeier:
I think this is the best metaphor, but let's see if the others also like it. And thanks by the way for all the cool work and for trying out my brain storming!

madsrh:
I'm sorry, but I don't think this is the right path. Nothing wrong with the artwork by Paul or the metaphor, but I'm not 100% sold on any of these fine suggestions. Just my opinion.

image

Feichtmeier:
My problem with using the cutted gears as the new to go base plate is that all the other mimetypes look completely different. We never cut stuff at the edges. It's always a clear symbol upon a baseplate.
Imo just taking random icons from upstream doesn't work.

What if we put the cartridge on a baseplate instead. On the oblong baseplate?

Jupiter007-43:
😋 Cartridge on white background:

Capture d’écran du 2020-08-04 19-18-02

Jupiter007-43:
I did a version with a dark grey background:

Capture d’écran du 2020-08-05 22-01-50

And should we or should we not add text (I prefer without):

Capture d’écran du 2020-08-05 22-04-46

By the way I look at the symlink list of roms and every console are using cartridge so this pictogram is IMO the best way.

Feichtmeier:
If it's for the generic icon, I'd say without text.
Then for the dedicated ones for gameboy, snes and whatever you have time to add, you could add either a different cartridge (time consuming) or just add a different text (easy)

Jupiter007-43:
Firstly I'll link this icon to the others, then I'll try to create some specific pictogram to see (if it looks not good I'll add text below).

Jupiter007-43:
How should I name this generic icon just rom.png?

Feichtmeier:
No idea :shrug:  :D I bet you find a cool name!

ubuntujaggers:
Just a thought: could the cartridge be the whole icon instead of pictogram on baseplate?

Jupiter007-43:
Like the previous version?

> Cartridge icon (just ignore the triforce logo):
> 
> Capture d’écran du 2020-08-04 15-32-07

ubuntujaggers:
My apologies, I managed to miss that!!  I'm on my phone and accidentally zipped through most of the thread.

I wonder if it would still be recognisable in the oblong shape? I think application-x-executable is a special case because it's used as an app icon too, but apart from that, multiplying the number of non-oblong mimetypes isn't going to be great for consistency.

I certainly don't dislike the cartridge on the baseplate, so if everyone is happy with that, let's have that ☺️

ubuntujaggers:
Thinking about it (and sorry I've been fairly absent these last few weeks, I'll do better and make some time - given that we have more contributors now) I think we need to lock down the principles of mimetypes asap. We've got lovely mimetypes now, but the longer we continue to expand the set without clear principles, the more we risk getting in a mess later. How about a conference call this weekend?

Jupiter007-43:
I resized the cartridge on the oblong baseplate yesterday:

Capture d’écran du 2020-08-07 11-27-13

Do you think it looks good?

Jupiter007-43:
Master system cartridge:

Capture d’écran du 2020-08-07 13-47-27

Feichtmeier:
Can we merge this? I think this looks pretty cool:
grafik

@madsrh @ubuntujaggers ?

ubuntujaggers:
Nice pixel work @Jupiter007-43 :+1:

PR #2312, opened 2020-08-05
Lighten dark theme success_color

Jupiter007-43:
### Adwaita

Capture d’écran de 2020-08-04 21-14-10

### Before

Capture d’écran de 2020-08-04 21-16-18

### After

Capture d’écran de 2020-08-05 12-45-01

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -19,7 +19,7 @@ $dark_fill: mix($borders_color, $bg_color, 50%);
 
 $warning_color: $yellow;
 $error_color: $red;
-$success_color: if($variant == 'light', lighten($green, 5%), darken($green, 5%));
+$success_color: lighten($green, 5%);
 $destructive_color: if($variant == 'light', $red, darken($red, 10%));
 
 $osd_fg_color: #eeeeec;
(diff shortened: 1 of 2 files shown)


Feichtmeier:
I always thought that is is too dark :shrug: 


could you check
-  how it looks in the shell? for example the dock badges?
I like to keep both color files as much in sync as possible
- does this change anything else in the gtk theme?

Jupiter007-43:
>  for example the dock badges?

Badges are red on my computer:

Capture d’écran du 2020-08-05 13-04-09

This affect the levelbar:

Capture d’écran du 2020-08-05 13-11-51

Capture d’écran du 2020-08-05 13-10-18

But the contrast is also better here.

Feichtmeier:
I think the level bar change is good, but wait for @clobrano and/or @madsrh opinion here.

Ah yeah... currently you only see the green badges in the ubuntu session which you dont change with the yaru meson build : (

clobrano:
It looks good to me 👍

madsrh:
Much better 👌

Jupiter007-43:
Great!
I tested some apps which use this color and I think it's okay!

Feichtmeier:
Cool, glad we have suggested_bg_color detached from success_color so I think it should be okay!
Stuff that is dangerous is when it is connected to borders. Like progressbgcolor and suggestedbgcolor

PR #2316, opened 2020-08-07
Suggestion: font mimetypes consistent with the categories icon

ubuntujaggers:
We inherited a nice (if subtle) font mimetype from Suru:

image

The reason I say "subtle" is because the guidelines only appear on the full size icon, which is rarely seen, and they're barely visible.  The icon makes me think of Rich Text Format or some other kind of text file.

We have a different symbol for "fonts" that appears as preferences-desktop-font:

image

...so this suggested PR makes the mimetypes consistent with that:

image

What do you think?


madsrh:
Yes! You're spot on again @ubuntujaggers 💪 

It makes sense totally to "sync" the two icons. However, I think the **preferences-desktop-font** icon can be improved. IMO it makes no sense when comparing fonts, to use different letters. 
~~Most~~ Some other icon themes and OS'es use the letter **A** with either two capital letters in different sizes **A**A  or one capital letter followed by a small one **A**a

IMHO we should adopt this metaphor (and sync the two icons ofc). Preferably with one letter as [Serif](https://en.wikipedia.org/wiki/Serif), like the example below from our friends in Redmond:

image

Feichtmeier:
I can't say it looks bad, but isn't it good that the mimetype icon like different to this very specific gnome app to open fonts, which is not the only app that can process fonts?

madsrh:
> ...isn't it good that the mimetype icon looks different than this very specific gnome app to open fonts, which is not the only app that can process fonts?

It's a good point, but couldn't this be fixed be making the gnome app icon look more like an app? Currently it's just sitting on a white baseplate like the document. So perhaps if it had a bit of color but used the same "pictogram" that might be different enough.

If not, then I vote for changing the mimetype to something like the above example 😄

ubuntujaggers:
Okay, we can add this to the list of "mimetypes issues" :) it's not a high pri but I thought it would be a quick win if it was uncontroversial, we can explore options :+1:

Feichtmeier:
Maybe make them blue, to look a bit different from the app? Other than that I understand that three letters invoke the association to font mimetype more than one letter ^^ so it's prbly a good change :thinking: 
How much fonts are in one font file?

Jupiter007-43:
> How much fonts are in one font file?

Just one I think.

Feichtmeier:
Ah I think it's good :+1: 
Do you want to change anything @ubuntujaggers or can we merge it?

ubuntujaggers:
The point about multiple apps viewing font files is a good one :+1:  Maybe the principle should be to only align proprietary file formats with the associated app.

Feichtmeier:
No idea @ubuntujaggers :shrug: 
It's up to you - since the others like it, and I think it looks Yaru ^^, even though the objections I posted, you can merge it if you like.

Feichtmeier:
@ubuntujaggers ping on this one :) What's happening with this?

PR #2317, opened 2020-08-07
Making image mimetypes landscape

ubuntujaggers:
The generic image mimetype is landscape, but the mimetypes for svg, xcf and kra (all images) are portrait.  I've asked whether they should also be landscape.  This is a draft PR so you can try them in landscape and see if you like them, to help us make a decision.

image

image


ubuntujaggers:
Oh, and the Krita files are also landscaped:

image

Feichtmeier:
Really cool idea. I think this is very intuitive!

madsrh:
This is a good idea. It makes totally sense, but it still takes some getting used to.

I tried the opposite - rotating the generic image to portrait, but just feels odd.

I'm going to test this PR for a while a see how it feels, but for now I'm definitely  +1 😃

ubuntujaggers:
Magic - I'll look at the other mimetypes, and if anyone can think of any others that are images or videos, name them here and I'll add them to the branch.

I'll also make the video icon have the same ratio as the image mimetypes. Then we'll have one mimetypes "rule" made consistent 👍

ubuntujaggers:
(You will likely have to turn off thumbnails in Nautilus to test.)

madsrh:
> if anyone can think of any others that are images or videos

3D files (stl or blender) comes to mind, but I think that's a different category.

> I'll also make the video icon have the same ratio as the image mimetypes. Then we'll have one mimetypes "rule" made consistent

Nice 🎉

Jupiter007-43:
Really good idea 👍
It looks very cool and consistent!

ubuntujaggers:
I added the Blender and Karbon mimetypes 👍

ubuntujaggers:
Anyone object to me merging?

PR #2325, opened 2020-08-16
Add gnome-disks-state-standby-symbolic icon

### LINKED ISSUE #2324: Icon Request: Disk Standby Icon for gnome-disk-utility
In Ubuntu 18.04, the icon for a sleeping disk is a sort of "ZZZ" icon (/usr/share/icons/hicolor/scalable/apps/gnome-disks-state-standby-symbolic.svg)
In 19.10 and 20.04, the sleeping disk is now accompanied by the main application icon (/usr/share/icons/Yaru/scalable/apps/org.gnome.DiskUtility-symbolic.svg, or possibly one of a handful of similarly named but identical SVGs).

See also this [Ask Ubuntu question](https://askubuntu.com/questions/1226490/what-does-this-icon-in-disk-utility-mean) and this [gnome-disk-utility issue](https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/issues/180).

The Ask Ubuntu thread provides examples of both icons, but I can probably find and attach the SVGs if helpful.

(Apologies if this issue is posted in the wrong place or lacking crucial information.)

Jupiter007-43:
I reused the same metaphor as upstream (zzz), adding a light gradient of opacity:

Closes #2324

### DIFF
--- a/icons/Suru/scalable/apps/gnome-disks-state-standby-symbolic.svg
+++ b/icons/Suru/scalable/apps/gnome-disks-state-standby-symbolic.svg
@@ -1,0 +1,151 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   id="svg7384"
+   sodipodi:docname="gnome-disks-state-standby-symbolic.svg"
+   version="1.1"
+   height="16"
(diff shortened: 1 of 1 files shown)


madsrh:
That was a fast response @Jupiter007-43 👍

How does this look in the light theme? Still visible

Jupiter007-43:
It looks good:

Capture d’écran du 2020-08-16 15-31-14

ubuntujaggers:
I'm not 100% sure this quite matches the existing symbols - I think the use of three opacity levels is unprecedented (I'm aware of some icons using 50%), and the chain of Zs is quite dense.  Also, the Z in the Ubuntu font has narrower proportions :thinking: 

How about less dense and without multiple layers of opacity:

image

?

Jupiter007-43:
> I'm not 100% sure this quite matches the existing symbols - I think the use of three opacity levels is unprecedented

Sorry, I thought it was a good idea for symbolize the sleep 😥

About the size of the Z, I compared with the Ubuntu font and I used the good size (5px width and 6px height). But the density is indeed too much. Fixed version:

Capture d’écran du 2020-08-18 20-55-05

ubuntujaggers:
That looks lovely!  Thank you :+1:

Feichtmeier:
Sorry for being late to the show, but this indicator is also used as a progress indicator, I am currently extracting an img file from a bootable USB Stick:
grafik

You think the ZzZz metaphor is right then? :thinking:

Jupiter007-43:
I used the same metaphor as upstream.
Personally I think it's ok:

Capture d’écran du 2020-08-19 11-51-47

PR #2326, opened 2020-08-16
New issue template workflow

Jupiter007-43:
I think it should be great to use the new issue template workflow into Yaru: https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates

You can look at the LO icons repository to have an example:

Capture d’écran du 2020-08-16 23-00-25

https://github.com/Jupiter007-43/libreoffice-style-yaru-fullcolor/issues/new/choose

I've only split the old issue template for the moment, but if you think this is a good idea, please tell me if you think that a generic issue need a template.

Thanks!


madsrh:
This looks like a good idea! The current template isn't working optimally IMO (no offense to who ever created it).

Speaking of improvements, could we somehow add people need to check if the issue also occurs in Adwaita? That would be great.

Jupiter007-43:
> This looks like a good idea! The current template isn't working optimally IMO (no offense to who ever created it).
> 
> Speaking of improvements, could we somehow add people need to check if the issue also occurs in Adwaita? That would be great.

Really good idea!

So I added this section to `bug_report`:

```markdown
**Upstream check**

Please also check if the problem also occurs with the Adwaita upstream theme:

- If the bug is related to an application, change the Gtk theme by running the following command:

    `gsettings set org.gnome.desktop.interface gtk-theme Adwaita`

- If the bug is about the icon theme, switch to Adwaita icon theme with this one:

    `gsettings set org.gnome.desktop.interface icon-theme Adwaita`

- If the bug concerns Gnome-shell, log-out then reconnect you using the Gnome session (click on the gear for switch).

Then check if the problem is still there.
```

Should the user still post the issue (in the case of an upstream bug) so we can redirect him to the correct repository?

Another thought, should be a good idea to split the `bug_report` template into a `gnome-shell_bug_report`, `application_bug_report` and `icon_bug_report`, for adding just the specifics instructions and clarify all this?

clobrano:
This looks great :+1:

I think that the issue template should be as simple as possible, to avoid scaring reporters :smile:

Feichtmeier:
Really good work!

One tiny bit of information is missing: to use the gnome session one needs to 
```
sudo apt install gnome-session
```

:)

Jupiter007-43:
> One tiny bit of information is missing: to use the gnome session one needs to
> 
> ```
> sudo apt install gnome-session
> ```

Indeed, I forgot that Gnome-session isn't installed by default 😅

No one has any other suggestions?

clobrano:
We could write it in stages. Like

- stage 1: Thank you for report, these are the minimum info we need
- stage 2: If you really want to help us, do also the following things
- stage 3: mind propose a PR? :smile:

.github/ISSUE_TEMPLATE/icon_problem_report.md
```diff
@@ -0,0 +1,60 @@
...
+
+* If you just use the default Ubuntu session (since Ubuntu 18.10)
+
+    `$ apt show yaru-theme-icon`
+
+* If you installed Yaru via Snap (Ubuntu 18.04 only)
+
+    `$ snap info communitheme`
+
+* I you installed from the sources, go the Yaru folder and copy the output of the following terminal command instead
```
vitalkanev, line 58: Typo. `I` => `If`

Jupiter007-43: Fixed! Thanks very much!

Feichtmeier:
@Jupiter007-43 I think we could also add a button for both gtk and gnome-shell problems (no feature requests)

Jupiter007-43:
> @Jupiter007-43 I think we could also add a button for both gtk and gnome-shell problems (no feature requests)

What do you mean exactly (actually I created 4 templates: gtk, shell and icon problem and missing icon request)?
By the way I'll convert this PR to draft as it's still work in progress.

Feichtmeier:
Oh? I just looked at the picture.
Those "gtk, shell and icon problem and missing icon request" are fine :+1: sorry then

Jupiter007-43:
> Oh? I just looked at the picture.

Ok, in fact the screenshot is the issue page of LO icons, just here for illustrate 😉

Jupiter007-43:
New PR here: ##2490

PR #2332, opened 2020-08-28
Rework color palette

clobrano:
To better support the incoming new Flavours, the `ubuntu_color.scss` which contains Ubuntu color palette is renamed into `palette.scss`. Each Flavour can override its `palette.scss` file without changing other scss files.

Moreover, this PR adds semantic colors to `palette.scss`, and using those in colors.scss would permit new flavours to just add their own palette file and adjust these semantic color variables to get their own style.

Added
- primary_accent_bg/fg_color
- secondary_accent_bg/fg_color

Linked to selection and progress colors


Feichtmeier:
Thanks :+1: 
Do you plan to sync the shell to this?

clobrano:
Yes, it makes sense

PR #2346, opened 2020-09-03
Upstream shell sync for groovy + keep yaru design changes

Feichtmeier:
Upstream shell sync + keep yaru design changes

- keep all visible yaru design differences
- add comments on every change to make it easier for other maintainers to keep our style
- add upstream fixes where it does not change our design

Edit: there will probably come more changes from upstream so let's keep this opened for a while - hope this is okay for your @3v1n0 ?


3v1n0 (approved):
I've been going through this, and looks good to me.

There are few missing changes that I'm syncing in a followup commit, and we're missing some resources, but *thanks a lot* for the _amazing job_ in commenting all this!

Will help for sure in future

PR #2374, opened 2020-09-30
symlink text-x-r-source to text-x-r

Muqtxdir:
hi,
 this will symlink **text-x-r-source** to **text-x-r**

## before:
r-before

## after:
r-after

Thanks!

### DIFF
--- a/icons/Suru/16x16/mimetypes/text-x-r-source.png
+++ b/icons/Suru/16x16/mimetypes/text-x-r-source.png
@@ -1,0 +1,1 @@
+text-x-r.png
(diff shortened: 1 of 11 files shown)


madsrh:
Thanks @Muqtxdir We love pull requests 😃 🎉 

Where does the _Rproj_ icon come from? @ubuntujaggers @Jupiter007-43 What do you think about making the two icons in the screenshot above more similar?

Muqtxdir:
>Thanks @Muqtxdir We love pull requests smiley tada

thanks @madsrh 

>Where does the Rproj icon come from? @ubuntujaggers @Jupiter007-43 What do you think about making the two icons in the screenshot above more similar? 

the second icon-mimetype  in the screenshot is for **appllication-x-r-project**

Issue #2164, opened 2020-04-25
In the gnome-shell calendar, days of a different month with events are colored differently from those without


Raytar:
![image](https://user-images.githubusercontent.com/10762082/80280572-24310500-8705-11ea-9627-3eaee8fd0383.png)

##  Expected Behavior
Any day of a different month should be colored differently from the days of the current month. 

## Actual Behavior
Days of a different month that have events are colored the same as days of the current (selected) month (January 1. and 8. in the image above have the wrong color).

## Steps to Reproduce the Problem

  1. Add events to the calendar in the beginning or end of a month
  2. Open the calendar and look at the colors of the dates

## Software that presents the issue

- Name: gnome-shell
- Version: 3.36.1-5ubuntu1

## Yaru version

```
$ apt show yaru-theme-gnome-shell 
Package: yaru-theme-gnome-shell
Version: 20.04.6
```

clobrano:
Hey @Raytar, thanks for taking time to contribute, and good catch

clobrano:
This might be an upstream issue, but I have to wait until my new modem arrives to have connection enough to download an up to date fedora for testing :)

clobrano:
Alright, no Fedora32 doesn't have this specific problem, but they have what we have fixed already :D, so, let's try to fix it locally, then in a next upstream sync will see what to do

clobrano:
Analyzing this issue I found two problems:

1. non-workdays with appointments have a wrong color (the same as the workdays, instead than paler).
2. next-month-days with appointments have a wrong color (the same as the workdays, instead than paler): this is the bug this ticket was opened for.

Focal 20.04 only has (2), while groovy 20.10 only has (1), which we had fixed in 20.04. So the plan is

- to look for the fix in Focal and check why isn't working in groovy
- check the fix in Groovy and backport it in Focal

PR #2380, opened 2020-10-01
calendar-other-month-day color shall not change if the day has an event.

clobrano:
"calendar-other-month-day" with event has a different color than
"calendar-other-month-day" without an event.

"calendar-other-month-day" is a special day, like non-working day, and
should not have a different color, or it would be mistaken for an
in-current-month working day.

Prevent this changing the color of a day with an event only if not a
"calendar-other-month-day".

Closes: #2164


clobrano:
This is probably shared with upstream. I will check and if confirmed, I will propose the same patch there as well.

Issue #2216, opened 2020-06-22
Nautilus column text in backdrop is gray


madsrh:
Nautilus column text in backdrop is gray, which can't be read on the selected item:

![image](https://user-images.githubusercontent.com/3986894/85334623-d4d83c80-b4db-11ea-8766-6146944b32a0.png)

madsrh:
I don't understand this 🤯 Did something change in master? Because now, it seems that the backdrop works fine. The labels are dimmed but not gray. 
This PR keeps everything white 🤔

clobrano:
Something in master or in Nautilus. The code I removed was explicitly to fix this very same bug, but now I can only keep it white :(

madsrh:
I'm fine with the current master, but sorry for wasting your time on this 😬

Feichtmeier:
Maybe something changed in nautilus? I mean in the gtk code?

clobrano:
I did reproduce the error when working on this, soo I am a bit suprised

Feichtmeier:
I can reproduce this with the current yaru version in focal. If we did not change _apps.scss since 20.04.0, I don't think anything changed here, and we need indeed the fix from carlos PR.

PR #2389, opened 2020-10-04
Backdrop text improvements

### LINKED ISSUE #2388: Nautilus sidebar looks different in backdrop
Looks like the gray sidebar in backdrop isn't the same in every app.

image

Feichtmeier:
- align all sidebars to have the same backdrop color, like the rest of the window
- fix default theme headerbar to have the intended, but never correctly set, backdrop color

Closes #2388 

Before:
image

After:
Dark theme:
grafik
grafik

Default theme:
grafik
grafik

Light theme:
grafik
grafik

@madsrh please test and see if you see stuff I've missed

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -80,15 +80,6 @@ $ambiance: null !default;
       &.dnd { border-style: none; }
     }
 
-    .nautilus-list-view .view:selected {
-      &:backdrop {
-        &, &:dir(ltr), &:dir(rtl) {
-          &, image, label, row { color: if($variant=='light', darken($backdrop_fg_color, 7%), lighten($backdrop_fg_color, 6%)); }
-          background-color: if($variant=='light', darken($bg_color, 7%), lighten($bg_color, 6%));
-        }  
-      }
-    }
-
     notebook > header {
       background: $bg_color;
     }
@@ -141,8 +132,8 @@ $ambiance: null !default;
       }
 
       &:backdrop {
-        background-color: transparentize($fg_color, 0.8);
-        color: $backdrop_text_color;
+        background-color: $selected_bg_color;
+        color: $backdrop_selected_fg_color;
       }
     }
 }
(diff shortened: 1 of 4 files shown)


madsrh:
@Feichtmeier could you try to add one more change? To me, changing this line to 80% in `_colors.scss` looks good.

`$backdrop_fg_color: mix($fg_color, $backdrop_bg_color, 80%);`

What do you think? :thinking:

Muqtxdir:
tried the new backdrop PR, and noticed the toggle switches are different in backdrop

### before:


before

### after:
after

Feichtmeier:
@madsrh @Muqtxdir please re-check

Apart from the fixes for gray rows I've added:
- default theme: pathbar -> align border
- nautilus header in treeview: correct bg color for all themes 
- switch border improvements in backdrop

Muqtxdir:
the pathbar is not clearly distinguishable to me in the default theme (minor nitpick)

### default theme:


### dark theme:


 
can the pathbar color  in default theme be same as the pathbar color in dark theme?


thanks!

Feichtmeier:
@Muqtxdir okay, please re-check

Muqtxdir:
yes, definitely better

Feichtmeier:
Okay okay. Here is what I think even if I made so much noise about it:

if we want to get this into groovy, we should revert the part of this PR that touches the rows (orange->gray).
But then also revert the Icon view becoming gray in nautilus. (because this is inconsistent currently)

The problem is, as @madsrh  pointed out, that there is always something orange leaking out of the backdrop, unless we change selected_bg_color from orange to gray and selected_fg_color from white to fg_color and then explicitly color things orange we want to be orange, like the close button or the tab underlines.

Maybe we can talk about good solutions for the next cycle in a dedicated issue.
If no one screams a big veto, I am going to remove the changes for the gray rows (but keep the rest, because I've fixed some more stuff here)

Feichtmeier:
This is fixed now: https://github.com/ubuntu/yaru/issues/2216

PR #2401, opened 2020-10-09
Add missing view-sidebar symbolic icon

### LINKED ISSUE #2400: Missing evince icon: view-sidebar-symbolic
grafik

https://gitlab.gnome.org/GNOME/evince/-/blob/master/data/icons/scalable/actions/view-sidebar-symbolic.svg

Jupiter007-43:
Capture d’écran du 2020-10-09 13-39-57

Closes #2400

### DIFF
--- a/icons/Suru/scalable/actions/view-sidebar-symbolic.svg
+++ b/icons/Suru/scalable/actions/view-sidebar-symbolic.svg
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Thank you @Jupiter007-43 !

I think you need to add it to source-symbolic first and then export the symbolic :smiley_cat:

Jupiter007-43:
In fact @ubuntujaggers and me have agreed to don't use the source-symbolic file and direct put svgs into Suru folder.
But if ofc if it's needed to use it there's no problem 😌

Feichtmeier:
Oh?
I missed this then!
So why does it still exist then?

Jupiter007-43:
Just in case we want to reuse it, but I don't know if that decision is still relevant today. 🤔

PR #2407, opened 2020-10-11
Lp bug tracker

clobrano:
This is an initial deploy for the resources needed to automatically
track LP issues on Yaru-theme package.

This commit contains:
- a python script that uses launchpadlib to get a list of yaru-theme
  bugs on Launchpad.
- a new GitHub action yaml file to configure the automation.
- the first copy of the json file containing the LP issues.

The automation is initially supposed to:
1. download the list of LP issues for Yaru and store them in a json file.
2. compare the above list with the one under launchpad/yaru_lp_bugs.json.
3. if the two files differs, the new one is stored and a PR is created.

A maintainer is then supposed to take the list, which contains ID, Title,
and web link, and manually check the issues, eventually adding or removing
bugs.

The python script is able to extract which bugs are actually new, and it is
possible that a future configuration will create the bugs automatically,
but this is left for future development, once this first step is working fine.

### DIFF
--- a/.github/lpbugtracker.py
+++ b/.github/lpbugtracker.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# vi: set ft=python :
+"""
+LP Bug Tracker uses launchpadlib to get the Yaru-theme bugs.
+The first time, the full list is saved in a json file.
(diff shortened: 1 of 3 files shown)


clobrano:
And it's already failing :rofl:

PR #2419, opened 2020-10-15
Replace single symbolic file with one file for each symbol

clobrano:
Since working in parallel on the same svg file can cause many
synchronization problems, this change is intended to split the
source-symbolic file into its components.

The copies - optimized - of the contents of source-symbolic is already
in icons/Suru/scalable folder, then this change copies these folders inside
icons/src/scalable.
Finally, this change provides a bash script to optimize the SVG files
using svgo, as the previous ruby script did.

render-symbolic-icons.sh can work on a single file, on all the files
into a group (e.g. actions), or on the full list of groups, using the
flags --file, --group, and --all.
See  render-symbolic-icons.sh --help for more info.

The next step of this change would be to remove the, now unnecessary,
resources like the ruby script and rewrite README.


Jupiter007-43:
I tested the new organization and it's very very much better than before 😃
Everything seems to work correctly 👍

Jupiter007-43:
Looking at the Suru folder I seen that some svgs aren't optimized, could you re-render all icons into this PR with your brand new script @clobrano ?

clobrano:
@Jupiter007-43, the only part of the original script I skipped, is trimming the image to 16x16 pixels, which was done with ruby code. 
Do you think it was important?

Jupiter007-43:
I don't think so, as all symbolic icons use a 16x16 canvas.
But honestly I don't know if it's an important feature to crop objects that could overflow.
Maybe @ubuntujaggers could answer better than me to that question.

Feichtmeier:
Thanks again for the awesome work @clobrano :clap: 

Could the icon crew check all open pull requests if they conflict with this change here? @Jupiter007-43 @ubuntujaggers @Muqtxdir ? thanks in advance

clobrano:
> 
> 
> I don't think so, as all symbolic icons use a 16x16 canvas.
> But honestly I don't know if it's an important feature to crop objects that could overflow.
> Maybe @ubuntujaggers could answer better than me to that question.

Ok, let's wait for @ubuntujagger opinion

clobrano:
Thinking twice, better do the re-rendering in a different PR, so that we can easily revert it if needed

Issue #2451, opened 2020-10-20
Cursor icons missing


seb128:
The bug has been reported initially on https://bugs.launchpad.net/ubuntu/+source/chromium-browser/+bug/1892085

On ubuntu focal (or current serie)

Install the chromium snap and try the MDN cursor demos at https://developer.mozilla.org/en-US/docs/Web/CSS/cursor for cursor: wait and cursor: grab. (Click the "cursor: wait;" or "cursor: grab;" button, then move the mouse over the blue box on the right to see the cursor set by that CSS option.), the CSS-set wait spinner and grab hand cursors are using the Adwaita theme

it seems because the 'wait' and 'grab' cursor are missing from yaru (firefox handle those by fallbacking e.g 'wait' to 'watch'), one option would be to symlink or copy those to the other names?

The following testcase is identical using gtk
$ python3 -c "from gi.repository import Gtk,Gdk,GObject; window = Gtk.Window(); window.show_all(); window.get_window().set_cursor(Gdk.Cursor.new_from_name(Gdk.Display.get_default(), 'wait')); Gtk.main()"

Feichtmeier:
Thanks @seb128 for the report

We had similar issues in the last with gnome  baobab for example. The wait cursor always reverted to adwaita.
In other apps it was always working right. Then in baobab it was fixed with a recent gtk release. Maybe the gtk snap connected to the chromium snap is outdated?

seb128:
The python command still shows the issue, also it's clear that there is no 'wait' file in the cursor directory for yaru, it should be easy to cp watch wait (and same for the .ani), which from local testing fixes the issue in the testcase at least

seb128:
@Feichtmeier trying baobab it's still bugging using gtk 3.24.23-1ubuntu1 which is current, could you give some details on what version you tested?
the source is https://gitlab.gnome.org/GNOME/baobab/-/blob/master/src/baobab-window.vala#L123

Feichtmeier:
> The python command still shows the issue, also it's clear that there is no 'wait' file in the cursor directory for yaru, it should be easy to cp watch wait (and same for the .ani), which from local testing fixes the issue in the testcase at least

Oh? Okay... I thought I remember baob working in gnome 3.38 but I can be wrong ofc
If this fixes it, can you create a PR?

seb128:
>  If this fixes it, can you create a PR?

Could you maybe providing some guidance on what is needed? In adwaita they did symlinks in the cursors dir
[...]

PR #2453, opened 2020-10-21
Install grab and wait cursors

seb128:
Closes https://github.com/ubuntu/yaru/issues/2451

### DIFF
--- a/icons/src/cursors/x11-make.sh
+++ b/icons/src/cursors/x11-make.sh
@@ -42,6 +42,7 @@ xcursorgen grabbing.in $DIR/$THEME/cursors/fleur
 xcursorgen grabbing.in $DIR/$THEME/cursors/grabbing
 xcursorgen grabbing.in $DIR/$THEME/cursors/size_all
 xcursorgen hand1.in $DIR/$THEME/cursors/hand1
+xcursorgen hand1.in $DIR/$THEME/cursors/grab
 xcursorgen hand2.in $DIR/$THEME/cursors/hand2
 xcursorgen hand2.in $DIR/$THEME/cursors/pointer
 xcursorgen left_ptr.in $DIR/$THEME/cursors/left_ptr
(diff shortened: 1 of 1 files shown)


seb128:
Tentative fix for #2451 , I don't really understand the build system for cursors though so it's just guess work. Does the win32 build need to be updated the same way (and how does one test the result on windows?)

seb128:
hum, also if not symlinking the watch.ani probably does need to be copied to wait.ani but that isn't in the same file?

Feichtmeier:
Thank you very much @seb128 

CC @ubuntujaggers @Jupiter007-43 please help @seb128  :smiley_cat:

Jupiter007-43:
I think @ubuntujaggers will be much more helpful than me. I had a hard time with my last cursors PR. 😥

ubuntujaggers:
Argh, cursors 😁

I will look at this as soon as I can 👍

Feichtmeier:
Hi @seb128 

Trying this command you've posted

```python
python3 -c "from gi.repository import Gtk,Gdk,GObject; window = Gtk.Window(); window.show_all(); window.get_window().set_cursor(Gdk.Cursor.new_from_name(Gdk.Display.get_default(), 'wait')); Gtk.main()"
```
which pops out the window and the wait cursor. however, with yaru master the wait cursor shows up in 20.10 with yaru master.
when I do the same on my 20.04 host the adwaita cursor shows up.

Trying chromium snap currently has no theming at all. It has that fuzzy x cursor and no yaru theme :| 

Any other idea how I could test this?

seb128:
> with yaru master the wait cursor shows up in 20.10 with yaru master.

Hum, maybe the issue is fixed in master? I'm trying on 20.10 ... is there any package from master I could try (or any other way to opt in for a trunk build?)

Feichtmeier:
you could

```
sudo apt install libgtk-3-dev git meson sassc
git clone https://github.com/ubuntu/yaru.git
cd yaru
meson build
cd build
sudo ninja install
```

seb128:
Ok, I don't understand it, the testcase works also in a VM for me, but also on a focal xubuntu live session when I just tried... need some more investigation

seb128:
seems like gtk has some mapping table 
https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/x11/gdkcursor-x11.c#L623

the extra files might still be required for other toolkit or chromium though?

Feichtmeier:
Sure thing, we can merge it anyways thanks for the work 
Maybe they changed the name in newer versions :) If you tested this yourself and it resolved the missing cursor let's merge it

Feichtmeier:
Okay, let's merge it, eventually we could include this in the yaru SRU to backport it for focal @seb128 ?

seb128:
Sorry for not replying earlier, including in one of the next SRUs should be fine yes, at the same time the main complain we received was with chromium which uses themes from the snap so that's probably low priority

PR #2454, opened 2020-10-21
(gnome-shell): Add widget files to upstream sync script

isantop:
GNOME Shell upstream has broken out widget styling into separate inherited files. Using the sync script does not fetch these files by default. This will add the new files to the list of files to fetch so that changes will be pulled in from upstream when necessary.


clobrano (approved):
It should work, good job!

PR #2518, opened 2020-12-23
Re-Install grab and wait cursors

### LINKED ISSUE #2507: Drag cursor is not consistent
Yaru cursor dragging icons are not consistent I don't know whether that is a feature or a bug though, when I drag a window I get this nice cursor:

Screenshot from 2020-12-17 10-54-30

But if I try to drag a web element, something inside a web app or even in some other desktop applications I get this:

Video:

https://streamable.com/2fc0nl

That is when cursor is hovering OVER the element, if I try to click and drag, it briefly turns to that white hand from first image, and then goes back to this dark cursor image again..

To reproduce this, it's easiest to try and drag elements in scratch editor, which I did in the second picture https://scratch.mit.edu/projects/editor/?tutorial=getStarted.

I used Chromium browser on Ubuntu 20.10

Jupiter007-43:
I found the problem: (#2453) was linking hand1 to grab but the make shell script wasn't been run.
So this PR just run the script (also link **hand1** to **grab** and **watch** to **wait** into w32-make.sh).

Closes #2507


Feichtmeier:
Great find. Sorry, I guess I should have checked the PR better. Do we have any tool for debugging all cursor names? :thinking:

Jupiter007-43:
> Great find. Sorry, I guess I should have checked the PR better. Do we have any tool for debugging all cursor names? thinking

Personally I use the MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Amr-Ibra:
> Personally I use the MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Just hover over the "CSS Value" column in the "Keyword values" table, and you'll notice the missing cursors.

I can see that Yaru is missing some cursors already:
- move (the same as grabbing in Yaru)
- vertical-text
- grab
- all-scroll (the same as grabbing in Yaru)

Jupiter007-43:
Looking at Yaru cursors folder, I can see the **move**, **vertical-text** and **all-scroll** cursors (and this PR add the **grab** one). I think it is caused by a bug with Firefox.

Btw I updated the csscursors.html test file: #2521

Amr-Ibra:
> I think it is caused by a bug with Firefox.

It's strange that Firefox can show most of the cursors listed in the table on MDN and some not. I can see that by hovering the mouse pointer on the table and see the cursor changes according to the CSS value.

Jupiter007-43:
It's indeed strange, I tried Gnome-web (Chromium is also Webkit based but as it's a snap we cannot add new cursors without update gtk-common-themes) and all cursors are present (using this PR ofc).

Jupiter007-43:
I have tested a lot this PR and I think there's no more cursor problem (except with Firefox but it's not related to Yaru).

PR #2528, opened 2021-01-11
Making network-server-symbolic more like the Yaru style

ubuntujaggers:
The current server symbol is very angular and not really in keeping with the current Yaru style:

image

This suggested change makes it a) more like the other symbols and b) more like the full colour server icon:

image

Just an idea so don't worry if you don't like it :+1:

### DIFF
--- a/icons/Suru/scalable/places/network-server-symbolic.svg
+++ b/icons/Suru/scalable/places/network-server-symbolic.svg
@@ -21,7 +21,7 @@
         <dc:format>image/svg+xml</dc:format>
         <dc:type
            rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
-        <dc:title></dc:title>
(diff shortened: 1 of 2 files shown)


madsrh:
IMHO I prefer the current. I agree that the sharp edges are a bit out of Yaru style, but that's the way a network-server looks 😄

Jupiter007-43:
I agree with @madsrh, the current pictogram is much understandable (but your proposal looks better IMO). Could you try to just add round corner on it?

ubuntujaggers:
How about this:

image

?

madsrh:
> How about this:

IMHO that is perfect

Jupiter007-43:
> IMHO that is perfect

+1, looks perfect 👍

ubuntujaggers:
Pushed :+1:

PR #2535, opened 2021-01-14
Add MS Publisher and QEMU disk image mimetype icons

Dragon8oy:
image
image


Jupiter007-43:
Thanks @Dragon8oy !
Could you try to reduce a bit the pictogram size of 48, 32, 24px versions, IMO it is too high actually.

Jupiter007-43:
@Feichtmeier Should we create two issue for each icons?

Feichtmeier:
Yeah why not :)

Dragon8oy:
image
Is this better? 48px might need to be a little smaller

Jupiter007-43:
> 48px might need to be a little smaller

+1 indeed, and it will be fine after imo.

Dragon8oy:
That should be okay, let me know if it needs more tweaking

Jupiter007-43:
Everything looks perfect, I don't have any more suggestion.
Thanks @Dragon8oy !

Jupiter007-43:
Is everyone Ok for merging?

Issue #2473, opened 2020-11-04
Gedit text is unreadable under Yaru dark theme


lekegf:
## Expected Behavior

See the text you're writing in Gedit in the current line.

## Actual Behavior

There's a white highlight in the current line (where the font is white as well), making it unreadable. 

## Steps to Reproduce the Problem

  1. Set Yaru theme to dark
  2. Open Gedit (Text Editor)
  3. Start typing 

## Software that presents the issue

- Name: gedit
- Version: 3.36.2

## Yaru version

The default in Ubuntu 20.04 up to the date 2020 November 3th
Important note: The same issue is present in the Adwaita dark as well.

![Captura de tela de 2020-11-03 23-03-21](https://user-images.githubusercontent.com/62917388/98061598-fd2d5b80-1e2a-11eb-9cf8-805627958c8c.png)

madsrh:
Hi. Thanks for reporting this issue

> Important note: The same issue is present in the Adwaita dark as well.

Since Yaru is based on Adwaita, this bug should be reported there (and it already is), so the fix is upstream.

https://gitlab.gnome.org/GNOME/gtksourceview/-/issues/29 you can see that it is marked as low level, or non-issue upstream. 

Closing the issue here

clobrano:
Also duplicate of #2090 and #2444

lekegf:
Since probably there's not going to be a upstream fix anytime soon, wouldn't it make sense disable the option "Highlight current line" in gedit by default in Ubuntu?

clobrano:
As yaru project, we have no way to decide gedit defaults. I think that the best option is just to select a different dark gedit colorscheme.

Issue #2140, opened 2020-04-09
Color Problem in Gedit with Yaru Dark Theme


srueegger:
## Expected Behavior

First of all I want to apologize. My English is not very good.

The only thing you have to do is open a text file in Gedit. It is important that the Yaru Dark Theme is activated.

## Actual Behavior

The line that is in focus is then no longer legible

## Steps to Reproduce the Problem

  1. Activate Yaru Dark Theme
  2. Open a Text File in Gedit
  3. Focus a line

## Software that presents the issue

- Name: Gedit
- Version: 3.36.1

## Yaru version
I'm not sure. Is the Yaru Theme in the actually Ubuntu 20.04 Beta Repositorys..
![Bildschirmfoto vom 2020-04-09 14-40-42](https://user-images.githubusercontent.com/11472914/78896567-f2673f80-7a70-11ea-9892-2d367e0f8606.png)

clobrano:
Hey @srueegger, thanks for taking time to help us.

This is a duplicate of #2090, we can't do much about it, I can suggest you to use a dark-friendly gedit's color scheme under <kbd>Preference</kbd>-><kbd>Font & Colors</kbd>

Issue #2090, opened 2020-03-19
Gedit: text barely visible on Yaru-dark with "Highlight current line" enabled


mrprobot:
## Expected Behavior

Readable text when "Highlight current line" is enabled.

## Actual Behavior

Text is barely visible because of the highlight color.
![Bildschirmfoto vom 2020-03-19 08-10-27](https://user-images.githubusercontent.com/23498825/77040925-3d81bb80-69b9-11ea-854c-7e8b8e1a44e1.png)

## Steps to Reproduce the Problem

  1. set Yaru-dark theme
  2. open gedit
  3. enable "Highlight current line"

## Software that presents the issue

- Name: gedit
- Version: 3.36.0

## Yaru version

apt show yaru-theme-gtk gives:

Package: yaru-theme-gtk
Version: 20.04.2
Priority: optional
Section: misc
Source: yaru-theme
Origin: Ubuntu

clobrano:
Hey @mrprobot, thanks for taking time to contribute. Could you tell me if it is reproducible with Adwaita as well?

mrprobot:
![Bildschirmfoto vom 2020-03-19 10-04-58](https://user-images.githubusercontent.com/23498825/77049911-32cf2280-69c9-11ea-9cac-b32cfc000a8d.png)

Yes, seems to be an upstream issue with the Tango color scheme.

clobrano:
This seems to be the related issue upstream https://gitlab.gnome.org/GNOME/gtksourceview/issues/29

and apparently the fix will appear in some next releases. Closing this one then, thanks!

Jupi007:
Hi @cociweb,
I can no more reproduce this issue using 22.04 and Yaru-dark scheme:

![Capture d’écran du 2022-10-02 10-24-59](https://user-images.githubusercontent.com/36476595/193445111-cbb35fd5-1c91-4450-bf93-a8618379176b.png)

Can you provide a screenshot, please?

cociweb:
Hello @Jupi007,
Here it is:
![image](https://user-images.githubusercontent.com/1935413/193470509-ec7dbf67-ecea-4e95-960d-be0c072bf0fe.png)
The version is upgraded from 20.04:
![image](https://user-images.githubusercontent.com/1935413/193470548-d898349e-a7dc-41d4-b660-ea79edc23191.png)

I can see some difference with your image: You have light window header&footer.

Jupi007:
> I can see some difference with your image: You have light window header&footer.

[...]

PR #2546, opened 2021-01-20
gtksourceview: Add dark and light styles

flexiondotorg:
This adds dark and light GtkSourceView styles that use a colour palette matching Yaru.


flexiondotorg:
These GtkSourceView styles, if applied by a gsettings override, would address #2090, #2140 and #2473

PR #2549, opened 2021-01-21
darker mate panel colors for default+dark

Feichtmeier:
Before:
grafik

After:
grafik
@flexiondotorg could you double check?

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -682,13 +682,14 @@ SheetStyleDialog.unity-force-quit { background-color: $bg_color; }
 /***********
 * Mate *
 ***********/
-
-$_base_bg_color: if($ambiance==true, $headerbar_bg_color, if($variant=='light', $bg_color, $headerbar_bg_color));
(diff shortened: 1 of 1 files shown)


flexiondotorg:
I like the proposed darker panels, since it is consistent with Yaru in GNOME Shell. Other than the re-colouring with Chelsea Cucumber, I'd like the Yaru MATE variations of the theme to be as close to Yaru as is feasible :slightly_smiling_face:

Feichtmeier:
yeah it matches better because its the same color :D 

the question is whether @flexiondotorg wants them to merge or keep them seperated

flexiondotorg:
> the question is whether @flexiondotorg wants them to merge or keep them seperated

Errrr, what is the question?

Feichtmeier:
hehe

Do you want the two elements "titlebar/headerbar" and "mate panel" to be visually separated from each other when the window is maximized or do you want them to be visually distinguishable ?

madsrh:
Wow, it only took five seconds for me to look like a total douchebag here 🤦‍♂️ 😆 I'm sure I made the same comment when we switched to the darker panels in the Yaru.

Sorry for the noise! 🤫

PR #2556, opened 2021-01-23
assets: Metacity (Marco) and Unity assets have had the colour retouched and refactored

flexiondotorg:
This pull request aligns the Metacity (Marco) window button colouring and behaviour to match how Yaru themed window buttons in CSD windows look and behave. The 3 variants are distinctive.

The Metacity (Marco) theme has been refactored to drop unsupported button types and remove rendering of shade and unshade buttons, since these are rarely rendered. Shade and unshade behaviour is preserved via shortcut keys or mouse button bindings.

With the exception of menu buttons, all window buttons are now .svg files, transformed from the .svg files available from gtk/src/default/unity

Slider .pngs have been rendered using the MATE accent colours.


Feichtmeier:
Couldnt find any problems here, so I guess this is working as intended :D

Feichtmeier (approved):

PR #2559, opened 2021-01-24
icons: Initial integration of Suru icons in Yaru MATE

flexiondotorg:
This pull request lands the initial integration of Suru icon into Yaru MATE.

Ambiant-MATE and Radiant-MATE are used as a fallback, particularly for AppIndicator support. In addition symlinks have been added for MATE components where a suitable Suru icon exists.


Feichtmeier:
Thank you very much Martin!

As those are so many commits, do you want them to be all visible in the history or are you okay with squash and merge, which would leave them as a list in the commit body?

PR #2564, opened 2021-01-25
metacity: Add unmaximize buttons

flexiondotorg:
When a window is maximized, the maximize button will change to reflect that the window can only be unmaximized.


flexiondotorg:
Ooops. Sorry, I already filed this PR and it is merged.

Issue #2517, opened 2020-12-23
Missing icon for Timeshift


tosziro:
There are missing icons for Timeshift and Htop. Maybe you could guys in spare time add Yaru versions of them? Right now they look like from other world. :-)

/usr/share/applications/timeshift-gtk.desktop
/usr/share/applications/htop.desktop

Im running Ubuntu 20.04 fully updated.

![missing](https://user-images.githubusercontent.com/41963712/102991814-44c29080-451a-11eb-8304-72c811ffb727.png)

Feichtmeier:
Hi @tosziro 
For both: if the creators of those apps give their ack, we can see if our icon designers have time to make something that fits
For htop: a symlink to system monitor should be enough

Feichtmeier:
@tosziro htop is merged thanks to @vitalkanev 

If you want this Timeshift icon, you need to ask those developers first, as we do not put icons into yaru that overwrite app icons against the will of their creators

Muqtxdir:
Hi @AaronTechnic 

> I believe the Timeshift developer is not against it. Linux Mint comes with timeshift preinstalled, and the dev knows that, and uses a custom icon.

could you share a screenshot of icon that is used in Linux just to confirm it. 

I think a symlink to [`backups-app`](https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/apps/backups-app.svg) would be the easy-way if we want to provide an icon for timeshift

AaronTechnic:
Hi, sorry for my late response, I have not used GitHub for a while, but i will send the screenshots soon.

Batwam:
I just installed Ubuntu 22.04 and it looks like the Backup icons is being used for the Yaru theme (HiColor theme has correct red icon). This wasn't the case in 21.10. Is this regression normal?

![image](https://user-images.githubusercontent.com/56710655/164149303-5183d345-ab2d-4b1e-b538-f2f3b78f2879.png)

PR #2565, opened 2021-01-25
Add symbolic link for 'htop'

vitalkanev:
Supplemental pull request for https://github.com/ubuntu/yaru/issues/2517

/cc @Feichtmeier because he suggested symlink https://github.com/ubuntu/yaru/issues/2517#issuecomment-750221951 (and possibly cherry-pick to Focal?)

### DIFF
--- a/icons/Suru/16x16/apps/htop.png
+++ b/icons/Suru/16x16/apps/htop.png
@@ -1,0 +1,1 @@
+system-monitor-app.png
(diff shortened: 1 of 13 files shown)


Feichtmeier (changes requested):
Looking good!
Eventually, symlink the symbolic, too? :thinking:

vitalkanev:
> 
> 
> Looking good!
> Eventually, symlink the symbolic, too? 🤔

Don't think so...

Feichtmeier:
Yeah, but if you want to symlink the app icon, you also need to symlink the symbolic icon.
Otherwise you'll end up with the new shiny app icon and the old symbolic

Feichtmeier (approved):

PR #2567, opened 2021-01-25
icons: Render all legacy icons

flexiondotorg:
Legacy icons include some glossy fullcolor versions in the rendered bitmaps. The glossy icons have been deleted and flat legacy icons rendered in their place.


Feichtmeier:
I don't get it, we had icons for this but they weren't rendered?

PR #2568, opened 2021-01-25
Adding start-here icons to Ubuntu Mate Yaru theme

ubuntujaggers:
This branch is currently just for testing, to see if the Yaru theme picks up start-here icons when added.  We can improve them when we know they're in the right places and correctly named.


flexiondotorg:
This works a treat :smiley: Thanks for working on in @ubuntujaggers

PR #2581, opened 2021-01-30
Update Unity window buttons and add Metacity themes

flexiondotorg:
This pull request updates the Unity window button svg files to provide pixel-perfect replicas of the window buttons rendered by CSD windows in GNOME Shell. Menu button assets have been added to Unity, that while not used by Unity, serve are source origin assets for the Metacity window buttons.

A script to transform the authoritative Unity window buttons into the 3 variants of Metacity theme is included. The script was used to populate the Metacity window button assets.

A template `metacity-theme-1.xml.in` is included which perfectly emulates window border rendering from GNOME Shell and CSD windows, including window button spacing, off-sets, titles and colours.

Finally, the meson build system configuration is updated to install the Metacity themes. Versioned symlinks of the `metacity-theme-x.xml` are also catered for, to support Metacity, Marco, Compiz, etc.

### DIFF
--- a/gtk/src/default/unity-to-metacity.sh
+++ b/gtk/src/default/unity-to-metacity.sh
@@ -1,0 +1,135 @@
+#!/usr/bin/env bash
+
+for FLAVOUR in default dark light; do
+    echo "${FLAVOUR}"
+    METACITY_DIR="../../../metacity/src/${FLAVOUR}/metacity-1"
+    mkdir -p "${METACITY_DIR}"
+
+    case ${FLAVOUR} in
+        default)
+            close_unfocused="636363"
+            close_unfocused_glyphs="F7F7F7"
+            focused_glyphs="F7F7F7"
+            focused_normal="323030"
+            focused_prelight="4F4F4F"
+            focused_pressed="5C5C5C"
+            unfocused_glyphs="A3A2A2"
+            unfocused_normal="3F3C3C"
+            unfocused_prelight="4F4F4F"
+            unfocused_pressed="5C5C5C"
+            ;;
+        dark)
+            close_unfocused="575757"
+            close_unfocused_glyphs="D2D2D2"
+            focused_glyphs="F7F7F7"
+            focused_normal="2C2C2C"
+            focused_prelight="4A4A4A"
+            focused_pressed="5E5E5E"
+            unfocused_glyphs="D2D2D2"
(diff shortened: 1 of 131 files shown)


Feichtmeier:
Thanks @flexiondotorg 
Since it includes changes to the build, would be cool if carlo could double check

flexiondotorg:
> Hello! could you change the position of the menu that is 11 to 3?

If you do that then the window button alignment won't match CSD windows. The intention here is to have windows rendered by Metacity, Marco, Compiz, etc looks the same as those rendered by GNOME Shell.

gtk/src/default/unity-to-metacity.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
+for FLAVOUR in default dark light; do
```
clobrano, line 3: I am not sure if I got this script completely, but I wonder if this is the right place for such script. If it applies to default, light and dark, is there a reason to keep it inside the default folder only?

metacity/meson.build
```diff
@@ -0,0 +1,11 @@
+flavours = []
+foreach flavour: ['default', 'dark', 'light']
```
clobrano, line 2: It would be good if we can find a way to define this "flavour" list in one place only (not sure it is possible), otherwise keep a note about update-it/keep-it-in-sync if we want to add a new GTK Flavour (the other place where this list exists is gtk/meson.build)

gtk/src/default/unity-to-metacity.sh
flexiondotorg: The `unity-to-metacity.sh` script is not intended to be run routinely. So, not something that should be executed every build, but only when the unity window assets are modified and the metacity assets require refreshing.

metacity/meson.build
flexiondotorg: Yeah, I've just copied the pattern I've seen used elsewhere.

gtk/src/default/unity-to-metacity.sh
flexiondotorg: @clobrano Anything more you need from me?

clobrano: I am sorry @flexiondotorg, I just missed this message. All ok for me, thank you!

clobrano (approved):

PR #2582, opened 2021-01-30
Pixel perfect indeterminate checkbox

Jupiter007-43:
**Before:**

Capture d’écran du 2021-01-30 13-23-23

**After:**

Capture d’écran du 2021-01-30 13-21-56


Feichtmeier:
Woah this is indeed much better. I think this was one of the first assets I've rushed last year ^^
Ty!

Feichtmeier (approved):

PR #2589, opened 2021-01-31
Move _palette to root

Jupiter007-43:
This PR move _palette.scss into **yaru/colors/** and symlink it to all subfolders.

@Feichtmeier is that you was thinking of?


Feichtmeier:
Yes would be nice. Let's see what @clobrano thinks about it

clobrano:
Don't we lose the ability to build variants with different palette this way?

Jupiter007-43:
> Don't we lose the ability to build variants with different palette this way?

It depends. How do that will be done?

clobrano:
I didn't write the process, but basically to have a new variant we just copy the gtk default folder with all the content. So the palette file can be changed without affecting the other variants.

With this change we shall remember to change not the shared palette under yaru/color, but to delete the symlink and replace  it with a dedicated file

Jupiter007-43:
Could it be possible to add several palettes into yaru/colors?

clobrano:
Yes, but the point of yaru/color isn't to share the same file?

Jupiter007-43:
Indeed the goal is to share the same file, but into all component (gtk3, gtk4, ...) of the same color variant (less it's my vision of things).
IMO we should create one _palette into /colors for each variants then symlink it into each components

clobrano:
This wouldn't be a problem, but if we like to have different accent variants, this would result in a bit of a mess, with some folders that use the symlink, and some others that have the custom file.

Feichtmeier:
maybe, just maybe have a $flavour variable that could have the values 'ubuntu', 'mate', "flavourXYZ'
Then have one palette file and set the whites, blacks, reds, oranges and semantic colors like 

```
$green: #109a26;
@if $flavour=='mate' {
    $green: #87a556;
}
@if $flavour=='xfce' {
....
}

....
$primary_accent_bg_color: $orange;
@if $flavour=='mate' {
    $primary_accent_bg_color: $green;
}
@if $flavour=='xfce' {
$primary_accent_bg_color: $blue;
}

...
```

?

clobrano:
This seems a bit overcomplicated to me. Way easier one file with the appropriate palette.

Feichtmeier:
You mean and then load a different _flavourxyz_palette.scss in  gtk.scss depending on the $flavour?

clobrano:
No, I meant having each gtk/flavour folder with a complete set of necessary files which the meson build picks without the need to know the flavour name.

Feichtmeier:
Hm, but then it is much harder to have all the code changes in all flavours.

You could have a $flavour var in gtk.scss and then load then palette files depending on this var, I  think this would be the easiest

clobrano:
Why harder? My understanding is that we only have 2 palette files right know. The gtk flavours that use the same palette already have symlinks

Jupiter007-43:
Are we also going to provide Shell color variants?
If yes, put all palette (the default one, the Mate one, ...) into a root folder is IMO the best way. But if we only provide Gtk color variants, symlink the most recent Gtk palette to all oldest Gtk version would be preferable.
One more question: will there be changes into scss code across color variants?

clobrano:
The original project kept separated shell and gtk, so to keep this separation, which simplifies support for variants without shell, I think it's better to have two palette files, even if similar. Consider that right now they look the same, but gtk and shell widgets are different, so also colours might differ. However I didn't have a look at gtk4 yet, so this separation might not apply anymore.

Jupiter007-43:
> However I didn't have a look at gtk4 yet

Gtk4 really looks like Gtk3 (it is more like a Gtk3++).

Feichtmeier:
Yes the _colors.scss file can and actually is already different.
But the palette shouldn't. That's the whole point of the extracting the palette into this file.
Also without this change flavours can only exist in a different branch which makes it even more annoying to maintain. 
Seriously if we want flavours to be a thing, we need the global palette and a flavour variant imo. No need to rush it in this pr but as soon as we resync the mate flavour to master this already makes sense (to me)

clobrano:
The palette shouldn't have semantic colours then

clobrano:
Flavours can exist in this project by the way. Just clone the default folder and add the proper flags in meson build.

clobrano:
What you call the flavour variable is already implemented in our "meson_option.txt"

each of the meson build options "default", "light", "dark", etc.

https://github.com/ubuntu/yaru/blob/9b47c2069c098881b6cac70d904b5affb719df72/meson_options.txt#L10

is a flavour flag, which is build by default (since the boolean value is True), but can be enabled/disabled in command line (`$meson build  -Dlight=false` for example disable build for light variant, see for example https://github.com/flathub/org.gtk.Gtk3theme.Yaru-light/blob/master/org.gtk.Gtk3theme.Yaru-light.json)

Feichtmeier:
Actually those accent colours are not semantic colours, if if the comment says so. They are accent colours and help to make different flavours possible. Semantic colours are "selected_bg_colour", "succes_color" etc

I think if we want to provide multiple flavours in the future, in one branch we shouldn't copy more code around. I'm really not a fan of code duplication

Jupiter007-43:
Code duplication is indeed very bad.
Could be possible to have one code base (to rule them all) and many enterpoints (could be light, dark and all color variants) which import they own _palette and snippet of scss?

PR #2607, opened 2021-02-05
GtkTheme: do not change tab fg color on hover

### LINKED ISSUE #2588: Revisit tabs styling
@clobrano @Feichtmeier I know that we've tweaked the tabs a ton of times 🙈 I don't recall why we landed with the current design (or maybe something changed), but having the label go darker on hover seems very unintuitive IMHO. 
Could we revisit or was there perhaps a larger issue blocking?
I remember that there was an issue with the black underline in this issue: https://github.com/ubuntu/yaru/pull/1983

Peek 2021-01-30 16-54

Peek 2021-01-30 16-55

Altso Mate:
alt text

Feichtmeier:
ya
Peek 2021-02-05 14-11

Fixes #2588

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -270,7 +270,6 @@ notebook > header {
                   border-color: transparent;
                 }
               }
-              &:hover:not(:checked):not(:backdrop) { color: $insensitive_fg_color; }
             }
           }
         }


madsrh:
Yes 👍

Jupiter007-43:
Nice @Feichtmeier !
As we also support Gtk4, this should be also updated there I think.

madsrh:
@Feichtmeier Hmm, I might have done something wrong, but here it looks like this is still broken in Mate :man_shrugging: 

light

dark

Feichtmeier:
Yes it's not in the mate branch yet

PR #2609, opened 2021-02-05
Add in/decrease symbolics icons

Jupiter007-43:
Capture d’écran du 2021-02-05 18-07-43


Feichtmeier:
So the answer is yes! Thank you :dancers:

PR #2617, opened 2021-02-06
Use different branch for each Github Action upstream projects

### LINKED ISSUE #2611: Github upstream actions architecture problem
There's a problem with the actual architecture of Github upstream actions:
it is using the same branch for two parts (ex: gnome-shell and gnome-shell resources) and so the first one is overwritten by the second. And there's the same problem with gtk (gtk3 and gtk4 so).
If you look at the gnome-shell auto PR (https://github.com/ubuntu/yaru/pull/2510) there's only changes about resources.
And looking at gtk auto PR (https://github.com/ubuntu/yaru/pull/2512) gtk3 changes has been overwritten since I added Gtk4 action.

Jupiter007-43:
Closes #2611

### DIFF
--- a/.github/workflows/open-pr-on-changes.yaml
+++ b/.github/workflows/open-pr-on-changes.yaml
@@ -13,21 +13,61 @@ jobs:
     strategy:
       matrix:
         buildenv: [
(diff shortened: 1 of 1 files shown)


Amr-Ibra:
A tiny cosmetic note: I noticed that the destination folder is named gtk+4.0, however, upstream dropped the "+" from the name, it's now GTK 4.

PR #2634, opened 2021-02-09
gnome-shell: Use $osd_fg_color for all the login messages

3v1n0:
We accidentally reverted this change (commit ad9d2b18) in commit 49b71e3
causing the PAM info messages (as the one for fingerprint) being drawn
with the wrong color.


Feichtmeier (approved):
LGTM :)
Still a bit time to UI freeze
Shell theme review is on my todo list after I finished my IRL stuff this week

PR #2646, opened 2021-02-11
ShellTheme: make the dark shell darker

Feichtmeier:
Independent of the decision which theme to default to (light or dark) this commit includes the following changes:

- for popups, notifications and dialogs the dark theme is now much darker, having a transparent border. The background however is *not* transparent, unlike in previous iterations, to prevent visual glitches
- border is also slightly darker than the OSD border
- the slider knob is adapted to the OSD chroma in gtk, the slider trough sharpness is improved by using a solid darkened white
- **_all these changes do not affect the light (currently default) theme_**

dark_shell_desktop
dark_shell_overiew
Screenshot from 2021-02-11 09-33-48
Screenshot from 2021-02-11 09-41-34

Edit: sorry missed this:

the rationale behind this is that the current very light "dark" colors are not blending so well over the gtk windows in my opinion. This dark/jet look features the current ubuntu style better, in my opinion.
So this is exclusively a style change and there are no technical problem with the dark shell currently in master

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -2,14 +2,14 @@
 // it gets @if ed depending on $variant
 @import 'palette';
 
-$base_color: if($variant == 'light', #ffffff, lighten(desaturate($jet, 20%), 10%));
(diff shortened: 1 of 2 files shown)


madsrh:
@Feichtmeier I am a huge fan of the light border/separator :tada:

PR #2650, opened 2021-02-14
Remove some duplicated symbolic icons

Jupi007:
Link:
- **value-increase-symbolic** to **list-add-symbolic**
- **value-decrease-symbolic** to **list-remove-symbolic**
- **mail-sent-symbolic**_(-rtl)_ to **mail-send-symbolic**_(-rtl)_

Also fix twisted path on **list-add-symbolic**


Feichtmeier:
grafik

:thinking: not sure if this is intended? =)

Jupi007:
> thinking not sure if this is intended? =)

Indeed 😓

Jupi007:
I checked everything and I don't noticed any problem.
I indeed removed **value-decrease** and I created **mail-send-rtl** but Github seems to think I renamed the first to the second.

ubuntujaggers:
Shouldn't be a problem - the icon is the right design and has the right name, so GitHub might think that one replaced the other, but it wouldn't make any difference to the theme :shrug:

Jupi007:
I fixed conflict problems by merging master into this.
I also canceled Github Actions (as there is some change of GA into master it triggered them) to avoid mails ^^

Can I merge?

PR #2654, opened 2021-02-16
Use scour instead of svgo for SVGs optimization

### LINKED ISSUE #2630: Corrupted symbolics in GtkFileChoser
### Happens when:

Spawned from vscode classic snap
With yaru build locally from master

grafik

### Does not happen when:

Spawned from Gedit deb
and/or
With yaru deb in focal
grafik

Edit: does not happen with the Adwaita icon theme, not in groovy not in focal

Jupi007:
As svgo render broken SVGs (see: https://github.com/svg/svgo/issues/1317) this PR propose to use scour instead (scour is used by Inkscape for saving optimized SVG).

**Before:**

**After:**

Closes #2630


Jupi007:
@Feichtmeier VSCode save dialog is fixed:
Capture d’écran du 2021-02-16 18-52-25

Muqtxdir:
This works great!

Issue #2408, opened 2020-10-11
Suggestion: Less orange (in backdrop)


madsrh:
This is a continuation of the gray sidebar discussion started by @Feichtmeier, where the goal was to reduce the amount of orange.

As you can see in [this example](https://user-images.githubusercontent.com/3986894/95690372-921bab00-0c17-11eb-965b-4b567c7b1d09.png), there's a lot of attention grabbing orange in the window that's actually in backdrop. Only the ❌ is colored in the active window. This is a bit misleading. So my suggestion is that **there's no orange in backdrop** (obviously nothing changes in the shell). There's some quick mockups below. If someone could make a testing PR that would be great, because this is something you need to feel and judging from a few screenshots, really isn't fair.

I think, this is also what @Feichtmeier was saying here - he just uses them fancy word 😜 

> The problem is, as madsrh pointed out, that there is always something orange leaking out of the backdrop, unless we change selected_bg_color from orange to gray and selected_fg_color from white to fg_color and then explicitly color things orange we want to be orange, like the close button or the tab underlines.

If anyone thinks that this is TOO MUCH GRAY, I would suggest reverting https://github.com/ubuntu/yaru/pull/2376 but still applying the "orange -> gray backdrop" 🙈 [Like this...](https://user-images.githubusercontent.com/3986894/95690656-86c97f00-0c19-11eb-86dc-03a52498a6a4.jpg)

![1](https://user-images.githubusercontent.com/3986894/95690732-06574e00-0c1a-11eb-8e9a-428d8f17fd10.jpg)

![3](https://user-images.githubusercontent.com/3986894/95690735-0b1c0200-0c1a-11eb-91d0-d60d08854f6b.jpg)

![2](https://user-images.githubusercontent.com/3986894/95690738-0eaf8900-0c1a-11eb-89f8-e3d23b92f36f.jpg)

Feichtmeier:
It looks less annoying in the backdrop with gray!

Problem: upstream doesn't have a`$backdrop_selected_bg_color` they only have a `$backdrop_selected_fg_color`
After fiddling around with this hecticly in the last PRs, I think the easiest would be to add this new color, and change %selected_items in `_common.scss` to include a backdrop section. Could be maybe only some lines of diff. Plus that one line for nautilus.
But one thing could always happen, since Adwaita doesn't have it, some app might have their new widget colored in the backdrop 🤷 
Let's see

clobrano:
[...]

PR #2655, opened 2021-02-16
Gtk theme: darker dark theme

### LINKED ISSUE #2227: Gtk3: Different font colors in backdrop in listbox/sidebar rows
White text should be "dimmed" equally in backdrop:

gif

The combobox text almost looks disabled in backdrop 🤷‍♂️

Feichtmeier:
- base the whole gtk dark theme on $jet `#181818` instead of $inkstone `#3b3b3b`
- as a result everything is now a bit darker
- another result is that the dark gtk theme is now closer to the shell popups/dialogs/notifications
- the gtk theme however does not use half transparent borders to avoid glitches
- gtk4 is adapted to this change (should be the same result ideally)

WIP: up-to-date screenshots can be followed through in the conversation below

Everyone please have a look and share your feedback.

Closes #2227

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -23,8 +23,15 @@ $ambiance: null !default;
     // keep details box visible
     // details box looks ugly when many items are selected and it floats above the orange
     background-image: none;
-    background-color: if($variant=='light', white, $base_color);
-    &:backdrop { background-color: $backdrop_base_color }
+    background-color: if($variant=='light', white, $bg_color);
+    &:backdrop { background-color: if($variant=='light',  $backdrop_base_color, $backdrop_bg_color); }
+
+    placessidebar {
+      background-color: if($variant=='light', $bg_color, $base_color);
+      &:backdrop {
+        background-color: if($variant=='light', $backdrop_bg_color, $backdrop_base_color);
+      }
+    }
 
     // Makes icons less bright in backdrop
     @at-root %dim_icons_in_backdrop,
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -29,7 +29,7 @@ $ambiance: null !default;
     placessidebar {
       background-color: if($variant=='light', $bg_color, $base_color);
       &:backdrop {
-        background-color: if($variant=='light', $backdrop_bg_color, $backdrop_base_color);
+        background-color: $backdrop_bg_color;
       }
     }
 
--- a/gtk/src/default/gtk-3.20/_colors.scss
+++ b/gtk/src/default/gtk-3.20/_colors.scss
@@ -55,9 +55,9 @@ $insensitive_borders_color: $borders_color;
 
 //colors for the backdrop state, derived from the main colors.
 $backdrop_base_color: if($variant == 'light', darken($base_color, 1%), lighten($base_color, 3%));
-$backdrop_text_color: if($variant == 'light', mix($text_color, $backdrop_base_color, 80%), mix($text_color, $backdrop_base_color, 60%));
+$backdrop_text_color: if($variant == 'light', mix($text_color, $backdrop_base_color, 80%), transparentize(white, 0.45));
 $backdrop_bg_color: if($variant == 'light', $bg_color, lighten($bg_color, 3%));
-$backdrop_fg_color: if($variant == 'light', mix($fg_color, $backdrop_bg_color, 80%), mix($fg_color, $backdrop_bg_color, 60%));
+$backdrop_fg_color: if($variant == 'light', mix($fg_color, $backdrop_bg_color, 80%), transparentize(white, 0.35));
 $backdrop_insensitive_color: if($variant == 'light', darken($backdrop_bg_color, 15%), lighten($backdrop_bg_color, 15%));
 $backdrop_selected_fg_color: if($variant == 'light', $backdrop_base_color, $backdrop_text_color);
 $backdrop_borders_color: mix($borders_color, $bg_color, 80%);
--- a/gtk/src/default/gtk-3.20/_colors.scss
+++ b/gtk/src/default/gtk-3.20/_colors.scss
@@ -55,9 +55,9 @@ $insensitive_borders_color: $borders_color;
 
 //colors for the backdrop state, derived from the main colors.
 $backdrop_base_color: if($variant == 'light', darken($base_color, 1%), lighten($base_color, 3%));
-$backdrop_text_color: mix($text_color, $backdrop_base_color, 80%);
+$backdrop_text_color: if($variant == 'light', mix($text_color, $backdrop_base_color, 80%), mix($text_color, $backdrop_base_color, 60%));
 $backdrop_bg_color: if($variant == 'light', $bg_color, lighten($bg_color, 3%));
-$backdrop_fg_color: mix($fg_color, $backdrop_bg_color, 80%);
+$backdrop_fg_color: if($variant == 'light', mix($fg_color, $backdrop_bg_color, 80%), mix($fg_color, $backdrop_bg_color, 60%));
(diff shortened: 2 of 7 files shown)


madsrh (commented):
This looks good @Feichtmeier 
I wonder if the backdrop also should have the sidebar background:

image

Muqtxdir:
hi, 

this darker-theme looks nice and beautiful. however, for me the backdrop is where this is getting tricky.
In some apps, there is no problem in backdrops but a few have minor-nitpicks. so far I've only used this for couple of hours and **gnome-web(epiphany-browser)** sticks out here (i'll add more of the backdrop-issue cases when I come across but please do keep in mind this are very very minor-nitpicks):

### current-dark theme:
before

### darker-dark theme:
after

Feichtmeier:
Thanks @Muqtxdir 
Considering that we went much lighter with yaru dark than Adwaita dark in 20.04 I wonder how Adwaita dark themes the contrast in the backdrop? I am not at the computer right now but I think that yaru light and Adwaita don't use much contrast in the backdrop by design. This doesn't mean than we can and should do what we think is correct though

elioqoshi:
Thanks @Feichtmeier !

This looks good to me and I'd consider this an improvement on the first sight. However what @Muqtxdir is a good point. We should probably not sacrifice eligibility here. Is there a way to have the best of both worlds?

> I wonder if the backdrop also should have the sidebar background color:

I'd think it's good the way it is as it contrasts with the background and sets foreground and background apart better.

P.S: Is it fine if I add myself to reviewers? Just so I can cover Yaru during my work time.

Feichtmeier:
> Thanks @Feichtmeier !
> 
> This looks good to me and I'd consider this an improvement on the first sight. However what @Muqtxdir is a good point. We should probably not sacrifice eligibility here. Is there a way to have the best of both worlds?
> 
> > I wonder if the backdrop also should have the sidebar background color:
> 
> I'd think it's good the way it is as it contrasts with the background and sets foreground and background apart better.
> 
> P.S: Is it fine if I add myself to reviewers? Just so I can cover Yaru during my work time. Let me know if you prefer a process.

Sure thing - very much appreciated to have you reviewing!

Best of both world should be possible.
The upstream theme we rely on has various vars connected to the base colors and then darkening / lightening it appropriately in the backdrop/active/and so on. But those functions have their limits. Or maybe we could say that they are not "linear".
Eventually I need to fix the backdrop colors, but if we do this, it needs to be in _colors.scss or _drawing.scss and not special casing more stuff in _tweaks.scss :)

Feichtmeier:
Okay, so let's take a look at Adwaita light first - the idea here is to have the backdrop fade out and only show a minimal difference between element states:

adwaitalight

We follow with yaru-light 100% here and I think this works great:
adwaitalight

Adwaita dark also has a minimal difference in backdrop between the states:
adwaitadark

Current Yaru-dark actually is pretty similar to Adwaita dark
yarudarkcurrent

Proposed Yaru-dark is not very far away from it 
proposedyarudark

So I think those issues happen specifically in the apps you've shown @Muqtxdir 

It's the PDF viewer and the web browser right?
The webbrowser uses libhandy styling though. So maybe this is an app specific issue after all

Muqtxdir:
Thank-you @Feichtmeier, for pointing out the comparision between yaru-backdrop and adwaita-backdrop here

> It's the PDF viewer and the web browser right?

yes those two, I use them regularly so it caught my eye but  can't say about other apps

> The webbrowser uses libhandy styling though. So maybe this is an app specific issue after all

does the work done by @Jupi007 for adding libhandy-styling to gtk not overwrite the default libhandy-styling?

Feichtmeier:
I think we can get closer to what we had by just lightening up the backdrop overall a little bit, so the button functions don't overreact so much:

Here everything is a little bit lighter plus the popovers rely on the regular backdrop border color @madsrh 
grafik

Muqtxdir:
pulled in the latest commit changes and the backdrop definitely improves, lightening up the colors in backdop improves the eligibility on the app's I've tried 👍
### before darker-theme backdrop:
darker-screenshot
### latest darker-theme backdrop:
new-dareker-screenshot

thanks !

Feichtmeier:
Thanks @Muqtxdir glad it works

A little improvement to backdrop insensitive buttons in the backdrop, now they change bg color accordingly 
grafik

madsrh:
Sorry @Feichtmeier we've misunderstood each other again 😄💬 What I was trying to say was, that these two have the same color in focus, so shouldn't they also have that in backdrop?

image

Feichtmeier:
> Sorry @Feichtmeier we've misunderstood each other again smilespeech_balloon What I was trying to say was, that these two have the same color in focus, so shouldn't they also have that in backdrop?
> 
> image

hehe np

Is this the case only in my change? Because I didn't touch any text / forground color vars :thinking: 
Or do you mean in general? If yes then we should maybe set up an extra issue for this

Or do you mean the same background color?
If yes, then I would rather adjust the background color of the sidebar of nautilus in backdrop

elioqoshi:
@Feichtmeier Looking solid! I am wondering that maybe it can be a bit clearer which window is in the background by slightly making the icon and text colour darker on the backdrop window. It could help emphasize which window is currently focused, what do you think?

Feichtmeier:
> @Feichtmeier Looking solid! I am wondering that maybe it can be a bit clearer which window is in the background by slightly making the icon and text colour darker on the backdrop window. It could help emphasize which window is currently focused, what do you think?

Do you mean the color of the text and the color of symbolic icons in the backdrop in the sidebar of nautilus and gnome control center?

elioqoshi:
> Do you mean the color of the text and the color of symbolic icons in the backdrop in the sidebar of nautilus and gnome control center?

Yep

Feichtmeier:
Yes you are right, the light theme actually fades out much more
grafik
grafik

Will fix this later - thanks! :+1:

Jupi007:
I tested this PR the whole day and I didn't noticed any problem.
I think it looks very good!
Good work @Feichtmeier 💪

Feichtmeier:
@elioqoshi is this okay or too much?

grafik
grafik

But that gray in the orange rows is not so sexy.
Maybe we can fix #2408 with this PR too


#2227 is now fixed @madsrh :)

madsrh:
> #2227 is now fixed @madsrh :)

🎉 

> But that gray in the orange rows is not so sexy.

+1 maybe just use white with 50% opacity

elioqoshi:
> @elioqoshi is this okay or too much?

I think this is too little in fact.

> +1 maybe just use white with 50% opacity

How about 30% white? Disabled states of web elements usually use 30% of the default state

madsrh:
> How about 30% white? Disabled states of web elements usually use 30% of the default state

I'm good with anything but gray 😄 ...but the "button" isn't actually disabled - it's just not in focus 😉 

image

elioqoshi:
> but the "button" isn't actually disabled - it's just not in focus 😉

You might be right, although functionally there is not much of a difference of something currently being disabled and not being in focus. In both case you can't interact with it. Interacting with an unfocused window brings it into focus but doesn't register any interaction on that window in itself.

Feichtmeier:
There is even :backdrop:disabled and crazy states like this in gtk ^^ 

Let's try transparent whites for the dark themes backdrop color 
grafik
grafik

WDYT @elioqoshi @madsrh ?

elioqoshi:
@Feichtmeier I do think that looks good. Is that 50% white? My first intuition would be to make it even more transparent, but I might be biased as I checkeed on macOS and ttheyy apprroximately fade out 50% of unfocused UI elements too.

Feichtmeier:
> @Feichtmeier I do think that looks good. Is that 50% white? My first intuition would be to make it even more transparent, but I might be biased as I checkeed on macOS and ttheyy apprroximately fade out 50% of unfocused UI elements too.

It's 35% and 45%
Depending if it is drawn on $backdrop_base_color or $backdrop_bg_color

Feichtmeier:
Anything else or can I merge @elioqoshi @madsrh ?

madsrh:
This looks good 👍

elioqoshi:
Looking good, thanks!

PR #2672, opened 2021-02-19
Add Build CI test

Jupi007:
Add very simple Build CI test using Github Action.

### DIFF
--- a/.github/workflows/build-ci.yaml
+++ b/.github/workflows/build-ci.yaml
@@ -1,6 +1,6 @@
 name: Build CI test
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Looking good!
Eventually start gtk3-widget-factory and see if it outputs any errors, which it shouldn't? But this is optional

Jupi007:
> Eventually start gtk3-widget-factory and see if it outputs any errors, which it shouldn't? But this is optional

That could be a good idea, but sadly, as there's no display server, the widget-factory cannot run.

didrocks:
This looks good and the test on this PR passes :)

didrocks (approved):

PR #2675, opened 2021-02-22
GtkTheme: tone down selected rows

### LINKED ISSUE #2408: Suggestion: Less orange (in backdrop)

### LINKED ISSUE #2666: The color of the sidebar in Files is orange instead of gray
**Expected Behavior**

To see the gray color on sidebar in Files, following the Adwaita approach.

**Actual Behavior**

There's the orange color instead, in both active and non-active windows.

1. Open nautilus
2. Spot at the sidebar to see the orange color
3. Change the theme to Adwaita to see the difference 

- Files
- OS version: Ubuntu 20.04.2

**Yaru version**

- Version: 20.04.10.1

**Screenshots**
Captura de tela de 2021-02-18 01-32-46
Captura de tela de 2021-02-18 01-32-05

Feichtmeier:
Due to the ubuntu orange being quiet bright
we change the sidebar selection from a full orange
row to a thin orange stripe plus a soft gray selection
This change does not affect treeview rows, only .nautilus-list-view treeview

Before:
grafik
grafik

grafik
grafik

After:
grafik
grafik

grafik
grafik

CC @madsrh @elioqoshi @waynecrosby

Closes #2666 
Closes #2408

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -89,6 +89,21 @@ $ambiance: null !default;
 
       // Hide superfluous treeview drop target indication
       &.dnd { border-style: none; }
+
+      $_selection_bg: if($variant=='light', darken($bg_color, 5%), lighten($base_color, 5%));
+      treeview {
+        &:selected {
+          background-color: $_selection_bg;
+          color: $fg_color;
+          &:hover {
+            background-color: if($variant=='light', darken($_selection_bg, 5%), lighten($_selection_bg, 5%));
+          }
+          &:backdrop {
+            color: $backdrop_fg_color;
+            background-color: $_selection_bg;
+          }
+        }
+      }
     }
 
     notebook > header {
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -391,3 +391,23 @@ switch {
     }
   }
 }
+
+// Due to the ubuntu orange beeing quiet bright
+// we change the sidebar selection from a full orange
+// row to a thin orange stripe plus a soft gray selection
+$_selection_bg: if($variant=='light', darken($bg_color, 5%), lighten($base_color, 5%));
+list row, placessidebar row, sidebar row, .sidebar row {
+  &:selected {
+    &.activatable, & {
+      &, &:hover {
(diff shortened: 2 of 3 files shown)


Jupi007:
IMO, text color shouldn't change when selected:



It actually result a bad contrast.

madsrh:
> IMO, text color shouldn't change when selected:

+1

Otherwise it looks good 👌

elioqoshi:
Thanks @Feichtmeier !

> IMO, text color shouldn't change when selected:

Same. selected text and icon colour should stay the same. It should be good to merge otherwise.

waynecrosby:
@Feichtmeier I think everyone is agreed to just use the highlight and not change the text in both light and dark versions, then looks good to merge :)

Feichtmeier:
@waynecrosby @madsrh @elioqoshi @Jupi007 
I agree very much. I think now it should be the same
grafik
grafik

elioqoshi (approved):
Looking good, thanks!

Jupi007:
Tested. It's perfect 👍
Thanks @Feichtmeier !

real-amano:
And there goes Bob, the Big Orange Blob ;)

PR #2685, opened 2021-02-28
Adding full colour category icons for Ubuntu Mate

ubuntujaggers:
This PR just reuses existing designs, but they're a good match for the categories, and IMO better than not having Yaru icons - we can always add new designs as we have them :+1:


flexiondotorg:
I agree @ubuntujaggers :+1: Thanks for creating the symlinks, that looks much better.

@Feichtmeier Please merge this to `hirsute/mate`

PR #2687, opened 2021-03-02
Shell and gtk themes: do not transparentize border

Feichtmeier:
all dark themes: slightly brighten up the borders
shell theme: do not use a transparent border, as it becomes blurry on certain backgrounds

Before:
before

After:
after

CC @elioqoshi @madsrh


elioqoshi:
Great small improvement, thanks :)

PR #2688, opened 2021-03-04
GtkTheme: align dark themes separators to shell

Feichtmeier:
This aligns gtk dark themes menus and popovers to shell themes popup separators 

Before:
popovers_before
menus_before

After:
popovers_after
menu_after

CC @madsrh @elioqoshi

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -376,11 +376,21 @@ decoration {
                 0 0 0 1px if($variant=='light', transparentize(black, 0.77), $_dark_theme_outer_menu_border);
   }
 }
-popover.background {  
+@if $variant=='dark' {
+  menu separator {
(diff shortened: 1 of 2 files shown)


elioqoshi:
@Feichtmeier Good improvement, thanks, I'm all for it.

Kind of unrelated, how come the seperators on thee application popup are not full width and middle-aligned?

image


Should we make these full width with some padding instead maybe? Like here:

image

Feichtmeier:
> @Feichtmeier Good improvement, thanks, I'm all for it.
> 
> Kind of unrelated, how come the seperators on thee application popup are not full width and middle-aligned? Is this a Gnome Shell thing?
> 
> image
> 
> 
> Should we make these full width with some padding instead maybe? Like here:
> 
> image
> 

Yes that's a gnome shell thing :D kinda puzzles us many times over the years how inconsistent gnome shell and gtk looks. :/ Can't do much here sadly

Feichtmeier:
Thanks for the thumbs, merging then :)

PR #2713, opened 2021-03-09
GtkThemes: orange close only on :hover / :active

Feichtmeier:
This removes the orange close circle and only shows it on :hover or :active or :backdrop:hover
No screenshots provided since this needs real testing.

@ubuntu/yaru-team please test if you :+1: or :-1: this

Edit: inspired by https://github.com/canonical/ubuntu-desktop-installer/issues/43:

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -110,7 +110,7 @@ button.titlebutton:not(.appmenu) {
   .titlebar &,
   headerbar.selection-mode &,
   & {
-    &.maximize, &.minimize {
+    &.maximize, &.minimize, &.close {
 
       &, &:backdrop {
         &:hover {
@@ -124,21 +124,19 @@ button.titlebutton:not(.appmenu) {
     }
 
     &.close {
-      color: $selected_fg_color;
-      @include draw_circle($selected_bg_color);
-
       &:hover {
+        color: $selected_fg_color;
         @include draw_circle($selected_bg_color);
         @include draw_circle(lighten($selected_bg_color, 5%));
       }
 
       &:active {
+        color: $selected_fg_color;
         @include draw_circle($selected_bg_color);
         @include draw_circle(darken($selected_bg_color, 5%));
       }
 
       &:backdrop {
-        @include draw_circle(lighten($inkstone, if($variant=='light', if($ambiance==true, 15%, 40%), 10%)));
         &:hover { @include draw_circle(lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%))); }
       }
     }
--- a/gtk/src/default/gtk-4.0/_tweaks.scss
+++ b/gtk/src/default/gtk-4.0/_tweaks.scss
@@ -78,7 +78,7 @@ windowcontrols {
         box-shadow: none;
         background: transparent;
 
-        &.maximize, &.minimize {
+        &.maximize, &.minimize, &.close {
             &, &:backdrop {
                 background: transparent;
                 &:hover {
@@ -92,8 +92,6 @@ windowcontrols {
         }
 
         &.close {
-            background: $selected_bg_color;
-
             &:hover {
                 background: lighten($selected_bg_color, 5%);
             }
@@ -103,7 +101,6 @@ windowcontrols {
             }
 
             &:backdrop {
-                background: lighten($inkstone, if($variant=='light', if($ambiance==true, 15%, 40%), 10%));
                 &:hover { background: lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%)); }
             }
 


Jupi007:
Sorry @Feichtmeier but I Really Really don't like this 😖
I tested this during all yesterday evening and this morning ... and I just can't...
The orange close button is part of the Yaru/Ubuntu look and drop it make the headerbar very sad.

Off topic, but this is the first time I put a 👎  😅

Feichtmeier:
That's the this pr is for. I don't know how a headerbar can look sad, but I get your point. Let the others have a chance to test it and then we see :)

elioqoshi:
I have concerns about this as well mostly from a brand perspective. As @Jupi007 said, the orange close button has a lot of Ubuntu DNA to it.

HOWEVER, I do think it's not exactly right to keep it that way. After all the close button colour is a semantic signal colour (red) which means closing/removing something. However we are using a brand colour (orange) to fulfill the purpose of a signal colour which is weird in current Ubuntu as well.

I'd suggest to change the orange to a contrast-rich red which is always visible. For what it's worth I am also not sure how final the design decisions in Flutter are to keep it as a reference for production right now ( @Long-Chung might know more)

Feichtmeier:
@elioqoshi I agree that using the brand color or a "dangerous action" is not good.

I noticed (after I posted the "reference") that the close button is also a lot bigger there, which again reduces the "sad"-effect (tm), @Jupi007 mentioned.
Maybe, we could postpone this discussion until stuff like client-side-decorations without using the gtk headerbar works in flutter in Ubuntu and then draw our own title/headerbar which then has a much bigger close button and does not need the perma-signal-colored disk below?

elioqoshi:
I am in favour of digesting this a bit more and rechecking once we have more reference :)

long-chung:
The decision to remove the orange in the close button was more to align with what we get out of the box with flutter and material design.

Feichtmeier:
Hi @long-chung - thanks for diving in :)

Even though yaru.dart and yaru gtk are two individual GitHub projects, the close button somehow connects both to each other. Since from what I tried while working on yaru.dart, it is currently not possible to draw a flutter window without the gtk headerbar on top. So, if I understand everything correctly here, it is currently not possible to implement your mock-ups' close button, without changing yaru gtk? Except, we would receive a style class for the flutter windows and remove the orange disk only for this window (guess not really wanted for consistency)

Initu:
And I think it is more modern to remove the orange color as it was done in the sidebar, the less colorful the theme looks more professional.

elioqoshi:
> Personally I don't think that using a brand color for the the close button is a bad idea.
> In fact by **sad™** (😉) I mean that the close button add a bit of color into this very monochrome headerbar. Drop the color make it sad.
> Using the brand color allow us to use different colors like with the MATE variant.
> IMO this orange doesn't mean "danger": It is just a visual indicator to easily match the close button, as there is a strong contrast with the background.
> 
> And as I said upper, this orange button it is part of the Ubuntu look (just like the dark headerbar - even if it is bugged).
> When I see a dark header with an orange close button on any computer, I can see right away that this computer use Ubuntu.
> It's IMO important to have a strong visual look.

I have to disagree completely. It's less of an opinion that brand colours and semantic colours serve different purposes. Closing a window is a destructive action (that's exactly the word in UX research) and the X symbol is a semantic symbol representing a destructive action as well. Using a brand colour in conjunction with these doesn't make sense. It only looked like it made sense as the orange was close to a semantic red and served both purposes. A colour shouldn't serve both brand and semantic purposes though.

Jupi007:
Now I understand more your point of view, and I totally agree with you @elioqoshi . Brand colors should be used for "dangerous" actions buttons like the close one.

@Feichtmeier So maybe should we use `$error_color` instead of  `$primary_accent_bg_color`?

Jupi007:
I tried to play a bit with window buttons. Firstly I didn't found a way to increase the size of the icons (in fact so, using `-gtk-icon-transform: scale(2)`, but that looks blurry).

And this is how it looks using `$error_color`:

Capture d’écran du 2021-03-11 18-45-37

Feichtmeier:
I don't know. This looks even worse than orange. 
Maybe we should just postone this discussion until we have more flutter apps around. I can live with the orange close button, I just think that the big close button in the mock-up looks great without.

Jupi007:
I also don't like that red 😅

madsrh:
I think all the points raised are valid. I prefer the red over the initial purposed naked X, but my favorite option would be to leave the orange - at least for now.

Just remembered that the Redmond camp had a red button in Win7, but in Win10 it's only red on hover 🖱️

Feichtmeier:
Closing this then, let's remember this pull request when the time comes :brain:

clobrano:
>  It's less of an opinion that brand colours and semantic colours serve different purposes. Closing a window is a destructive action (that's exactly the word in UX research) and the X symbol is a semantic symbol representing a destructive action as well.

That's one of our oldest discussions :smile: 
Indeed, to associate a destructive action to a brand color is wrong in principle, however this is a particular case where the brand object is the "orange close window". It became an Ubuntu brand symbol starting with Ambiance, and many people expressed their preference over this special customization. This doesn't mean we have to keep it, in fact [Kill Your darlings](https://www.masterclass.com/articles/what-does-it-mean-to-kill-your-darlings) might be necessary and also okay if there are other elements supporting the brand (like the current use of aubergine).

PR #2721, opened 2021-03-13
removed text-x-plain.svg and text-x-cobol.svg

Muqtxdir:
Hi,

This PR removes **text-x-plain.svg** (*symlink already exists in mimetypes.list*) and **text-x-cobol.svg** (*added symlink to **text-x-script.svg** in mimetypes.list*)

#### screenshots:

Thanks!


Muqtxdir:
just noticed that, ***text-x-script*** was not rendered with the new fold, so rendered it again

Jupi007:
Nice catch for text-x-plain and the unrendered script icon @Muqtxdir, but I don't think that a good idea to drop the cobol icon by symlink it.
Cobol has a very special syntax and it's really not related to scripts.

Muqtxdir:
sorry, I thought they were supposed to look the same. I'll close this branch and repush it

Jupi007:
Don't be sorry, I think it's normal you don't know what is Cobol 😉

PR #2728, opened 2021-03-17
Gnome 3.38 resync

3v1n0:
Resync stuff from upstream 3.38 branch.

@Feichtmeier there are some changes we were keeping that are not upstream anymore, mostly the switcher values and the polkit dialog icon (that should not change though), so please ensure that this keeps following what we expected.

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_app-grid.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_app-grid.scss
@@ -29,9 +29,9 @@ $app_grid_fg_color: #fff;
 }
 
 /* App Folders */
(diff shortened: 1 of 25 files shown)


Feichtmeier:
Hey @3v1n0 👋 
Thanks for this. looks good from the changes but I will test this on my VM in a sec, good thing shell didn't change so I can just apply this branch without your PPA magic :)

Feichtmeier:
My gnome version is 3.38.3 in my VM
Looking good with your branch
Also the fix for the alt+tab is nice and I think we had an launchpad issue for this

Feichtmeier (approved):

PR #2751, opened 2021-03-28
rename sessions according to ubuntu 21.04 sessions

Muqtxdir:
Hi,

Since wayland will be the default-session starting ubuntu-21.04, this PR renames and aligns the session names for both wayland and Xorg same as Ubuntu session. 

#### screenshot: 
Screenshot from 2021-03-28 13-58-28

*Note: Screenshot taken in vm of hirsute-daily builds*

### DIFF
--- a/sessions/THEMENAME-xorg.desktop.in
+++ b/sessions/THEMENAME-xorg.desktop.in
@@ -1,6 +1,6 @@
 [Desktop Entry]
-Name=@ThemeName@ session on wayland
-Comment=This session logs you with @ThemeName@ on wayland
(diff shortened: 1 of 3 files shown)


Feichtmeier:
Looks good, has anyone tested what will happen to current sessions? Will this override them or add those on top?

Muqtxdir:
Hi @Feichtmeier,

I've used this PR and it does not overwrite the previous existing session instead creates a new session under xsessions and wayland-sessions

#### screenshot:


*Note: was unable to open/start hirsute-vm so took a screenshot on ubuntu 20.04 at the moment*

PR #2782, opened 2021-04-09
Update render-cursors script for inkscape 1.0

clobrano:
With inkscape 1.0 something changed in the way inkscape runs in background, and it seems than pre-running the script, creating a inkscape thread pool, communicating later the command to execute is not working properly.
As "Dirty Fix"® this commit replaces the inkscape thread pool with a direct call to inkscape with the full command for each rendering.
**Very likely this is slower, a lot**, but it should allow users to continue working with our cursors during a deeper refactory of this script.

### DIFF
--- a/icons/src/cursors/render-cursors.py
+++ b/icons/src/cursors/render-cursors.py
@@ -190,11 +190,12 @@ class SVGRect:
             if os.path.exists (output):
                 skipped[output] = True
                 return
-            command = '-w {size} -h {size} --export-id="{export_id}" --export-png="{export_png}" {svg}\n'.format (size=size, export_id=self.name, export_png=output, svg=svgFName)
-            dbg("Command: {}".format (command))
 
-            inkscape = subprocess.Popen (['inkscape', '--batch-process', '--shell'], stdin=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-            inkscape_instances[roundrobin[0]][0].stdin.write(command.encode())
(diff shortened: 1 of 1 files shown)


Jupi007:
Thanks a lot @clobrano for working on this ♥

Sadly it seems doesn't working on my computer.
I run `./render-cursors.py source-cursors.svg` and it just open a blank window of Inkscape.

clobrano:
Thanks for testing!
Did you first removed some PNG from bitmaps folders?

Jupi007:
> Did you first remove some PNG from bitmaps folders?

Yes, I removed one.

clobrano:
Weird :thinking: 
Which version of inkscape, and how did you install it?

Jupi007:
Inkscape 1.0.1 (3bc2e813f5, 2020-09-07)
I'm using the deb version.

clobrano:
I tested both snap and deb versions, but they are both 1.0.2, which is the package that goes into 21.04

Jupi007:
I did a `apt full-upgrade`. Inkscape version is still 1.0.1. I'm using 20.10.

clobrano:
The snap version should be 1.0.2, if you could test that we can find out whether is a matter of version or script

Jupi007:
Now it works with the deb (also tried snap) 🎉

Sorry for the merge commit, I clicked on the VS code button for fetching and it does that by itself.

clobrano:
Great 😊

It doesn't seem merged to me (from mobile)

Jupi007:
> It doesn't seem merged to me (from mobile)

No no I misspoke, I doesn't mean that one. It has merged this branch on itself :)

clobrano:
Oh, I see now. No problems 👍

Jupi007:
Thanks a lot @clobrano 💪

PR #2785, opened 2021-04-11
Rework render cursor script

clobrano:
Rework render cursor script to

- re-enable parallel rendering using multiple instances of inkscape (v1.0)
- use proper logging
- improve code legibility


Jupi007:
Just tested. Awesome work @clobrano 
It works perfectly, and it is just super fast 😯

clobrano:
Let's merge it then

PR #2792, opened 2021-04-14
Sync libhandy with upstream

Jupi007:
Sync our libhandy theme with upstream master branch.

The most important change add the "**Tab View**" support:

**Before:**

Capture d’écran du 2021-04-14 12-04-20

**After:**

Capture d’écran du 2021-04-14 12-04-02

**Question:** It still needs to remove all ugly shadows. But do we want to mimic normal Gtk tabs look (I mean the orange bottom border)?

### DIFF
--- a/gtk/src/default/gtk-3.20/libhandy/_Adwaita-base.scss
+++ b/gtk/src/default/gtk-3.20/libhandy/_Adwaita-base.scss
@@ -2,6 +2,9 @@
 @import 'fallback-base';
 @import 'shared-base';
 
+$tab_bg: darken($bg_color, if($variant == 'dark', 6%, 12%));
+$tab_bg_backdrop: darken($backdrop_bg_color, 6%);
+
 // HdyComboRow
 
 popover.combo {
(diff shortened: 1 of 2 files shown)


Feichtmeier:
I would say for now just take everything they do
Until not many libhandy apps are in ubuntu 
If more and more apps drop it we can think about design changes
At least that would be my opinion
@clobrano @madsrh ?

madsrh:
How does this look in Adwaita? I mean, is that their intended look? In the _before_ image you can't see which tab is selected (or where they are separated) 🤷‍♂️ Sorry if I sound stupid - I haven't tried libhandy handy yet 😁

Jupi007:
> How does this look in Adwaita? I mean, is that their intended look? In the _before_ image you can't see which tab is selected (or where they are separated) man_shrugging Sorry if I sound stupid

The after image is the default Adwaita look imported into Yaru (on the before one it is just how it looks now in Yaru with the missing styles).

Here is a better screenshot:

Capture d’écran du 2021-04-14 13-44-41

Left to right: unselected, hovered and current tab.

Feichtmeier (approved):

PR #2736, opened 2021-03-20
Initial support of GS40

Jupi007:
- [x] Import scss code from upstream.
- [x] Reimport Yaru changes.
- [x] Finish light variant.
- [x] Reinsert transparent backgrounds on panel-buttons.
- [x] Fix app-folder page preview

See screenshots in comments below 👇

Closes #2667

PR #2803, opened 2021-04-24
Fix app grid folder border-radius

### LINKED ISSUE #2801: hovering on ubuntu dock apps icons, names have uneven bordering issue
**Expected Behavior**
bordering must be on all four sides

**Actual Behavior**
bordering is only on the right & bottom part of the rounded corner rectangle 

first set wallpaper with the dark background if u r using ubuntu standard appearance because borders are of light colour 
hover on ubuntu dock icons so that u see names of apps & look at borders around naming which is a black rectangle

**System information**

**Yaru version**
yaru used for ubuntu 21.04

**Screenshots**

**Upstream check**

Jupi007:
Here is a "backport" of the fix I did for GS40 (https://github.com/ubuntu/yaru/pull/2736/commits/bce61903168e28f4bc89c8642f2c8f1525b6422f).

Closes #2801

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_app-grid.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_app-grid.scss
@@ -31,7 +31,7 @@ $app_grid_fg_color: #fff;
 /* App Folders */
 .app-well-app.app-folder {
   background-color: transparentize($osd_bg_color, 0.8);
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Thanks @Jupi007 

maybe we could create some sort of tags? This should def. go into small yaru update for 21.04, can we get a fresh master in @3v1n0 ?

PR #2807, opened 2021-04-25
WIP: Gtk theme window control experiments

### LINKED ISSUE #2798: When wearing glasses with high correction, the closing button moves within the red circle
**Explanation/Summary**

When using the Yaru theme, the colour of the X within the close button differs from the background around the red circle. For people using glasses with high correction, this causes the X to move within the red circle based on the viewing angle. To resolve this issue, the colour of the X needs to be the same as the title bar background. Since this is the case when using yaru-light, that's a valid "workaround". If at all possible, without affecting your style and branding, I would like to be able to change the colour of the X within the red circle.

**Expected Behavior**

The closing button (X) remains in the centre of the closing button.

**Actual Behavior**

When wearing glasses, the closing button (X) moves within the red cricle.

This is going to seem silly:

1. Wear glasses
2. Use the default or dark yaru theme
3. look at the closing button and move your head from left to right

All windows that use the yaru theme (and window borders)

**Yaru version**

* `20.04.10.1`
* `21.04.1`

Though, I'm fairly certain all version are affected.

**Screenshots**

Affected:

Screenshot from 2021-04-23 18-38-04

Not affected:

Screenshot from 2021-04-23 18-38-11

**Upstream check**

This issue does not affect Adwaita, as Adwaita doesn't use a circle around the closing button and the colour of the X and background differ.

Feichtmeier:
Unfinished screenshots, this will probably not go live

grafik

grafik

grafik

Peek 2021-04-25 23-58

Eventually...

Closes #2798

### DIFF
--- a/gtk/src/default/gtk-3.20/_tweaks.scss
+++ b/gtk/src/default/gtk-3.20/_tweaks.scss
@@ -110,10 +110,13 @@ button.titlebutton:not(.appmenu) {
   .titlebar &,
   headerbar.selection-mode &,
   & {
+    color: transparent;
     &.maximize, &.minimize {
 
+      @include draw_circle(if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color));
       &, &:backdrop {
         &:hover {
+        color: if($ambiance==true, lighten($headerbar_fg_color, 7%), $fg_color);
           @include draw_circle(if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color));
         }
 
@@ -124,22 +127,26 @@ button.titlebutton:not(.appmenu) {
     }
 
     &.close {
-      color: $selected_fg_color;
       @include draw_circle($selected_bg_color);
 
       &:hover {
+        color: $selected_fg_color;
         @include draw_circle($selected_bg_color);
         @include draw_circle(lighten($selected_bg_color, 5%));
       }
 
       &:active {
+        color: $selected_fg_color;
         @include draw_circle($selected_bg_color);
         @include draw_circle(darken($selected_bg_color, 5%));
       }
 
       &:backdrop {
         @include draw_circle(lighten($inkstone, if($variant=='light', if($ambiance==true, 15%, 40%), 10%)));
-        &:hover { @include draw_circle(lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%))); }
+        &:hover {
+          color: $selected_fg_color;
+          @include draw_circle(lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%)));
+        }
       }
     }
   }
--- a/gtk/src/default/gtk-4.0/_tweaks.scss
+++ b/gtk/src/default/gtk-4.0/_tweaks.scss
@@ -80,7 +80,7 @@ windowcontrols {
         color: transparent;
 
         &.maximize, &.minimize {
-            @include draw_circle(if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color));
+            background: if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color);
             &, &:backdrop {
                 background: transparent;
                 &:hover {
--- a/gtk/src/default/gtk-4.0/_tweaks.scss
+++ b/gtk/src/default/gtk-4.0/_tweaks.scss
@@ -77,11 +77,14 @@ windowcontrols {
         border: none;
         box-shadow: none;
         background: transparent;
+        color: transparent;
 
         &.maximize, &.minimize {
+            @include draw_circle(if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color));
             &, &:backdrop {
                 background: transparent;
                 &:hover {
+                    color: if($ambiance==true, lighten($headerbar_fg_color, 7%), $fg_color);
                     background: if($ambiance==true, lighten($inkstone, 7%), $_base_hover_color);
                 }
 
@@ -95,16 +98,21 @@ windowcontrols {
             background: $selected_bg_color;
 
             &:hover {
+                color: $selected_fg_color;
                 background: lighten($selected_bg_color, 5%);
             }
 
             &:active {
+                color: $selected_fg_color;
                 background: darken($selected_bg_color, 5%);
             }
 
             &:backdrop {
                 background: lighten($inkstone, if($variant=='light', if($ambiance==true, 15%, 40%), 10%));
-                &:hover { background: lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%)); }
+                &:hover {
+                    color: $selected_fg_color;
+                    background: lighten($inkstone, if($variant=='light', if($ambiance==true, 20%, 45%), 15%));
+                }
             }
 
             image {


reC0D3D:
This looks superb! I'm totally diggin' this new design. However, I have a few comments and concerns.

- On focused windows, the icons fade in yet they don't fade out. This isn't the case on unfocused windows. You can see this issue demonstrated in your attached gif. Upstream Adwaita unfortunately has the same problem. Is it possible to fix this?
- I feel like the minimize and maximize buttons should light up just a tad, similar to how the close button lights up on hover while revealing its icon.
- Speaking of "lighting up" or highlighting effects, the close button darkens when it's clicked, but the minimize and maximize buttons lighten. Something needs to be done to have all three buttons consistent.
- Maybe the buttons should be closer together? Right now I feel as if they take up too much space in the title bar. I'd like to hear other's thoughts on this idea.
- It's arguable that this new design may cause confusion since minimize and maximize look identical. MacOS solves this issue by introducing three colors instead of two. Should the maximize icon use the purple accent color or a lighter gray? Is this even something to consider? Lots of open-ended questions here.
- Lastly, when testing, I noticed that Firefox doesn't reveal the icons when hovering, unlike every other app and Chromium-based browsers. How will this be approached, seeing as Firefox is the default browser in Ubuntu?

elioqoshi:
Very good points @reC0D3D , thanks!

> Maybe the buttons should be closer together? Right now I feel as if they take up too much space in the title bar. I'd like to hear other's thoughts on this idea.

I feel the same but I cannot back it up with research on why. One could say that the touch area of each button should have less non-touch gutters between them. I suspect the reason they are also smaller on macOS is because it needs to minimize the risk (no pun intended) of misclicking them. I'd like to see 30-50% less spacing between them, but I can't really justify it why.

> It's arguable that this new design may cause confusion since minimize and maximize look identical.

Very true, I had the same concern. The only way to convey the information of what these buttons do is by memory (I remember the maximize button is in the middle) and by hover actions. Both might not be sufficient. The issue is however that window controls should use semantic colours, not brand colours. Using aubergine purple for maximizing creates this artificial believe that this brand colour has a semantic function, which is incorrect. Having said that, using a green for maximizing would probably be best.

ubuntujaggers:
I think this is a cool concept but right now I think I prefer current... because we have the large Adwaita-style headers, the circles look really big and chunky rather than sleek. But I like the idea enough to think it's worth exploring further :)

I agree that two identical buttons is a problem. A green circle in such a prominent position would be a big change to look and feel. Personally, I think aubergine for maximise/restore wouldn't be too crazy. We currently use aubergine for:

- Sliders (which have a bigger/smaller dimension)
- Toggles (when you're flipping between two states)

...so it sort of makes sense to me that aubergine would be associated with a function where you toggle a window between a larger and smaller size.  I'm not saying it makes completely rigorous scientific sense :) but I don't think it would be confusing, and it might be really cool to unite both brand colours at the top right of each window.

Feichtmeier:
Alright let's stay in this PR for the discussion about IF this is a solution for the issue, there are other solutions to this issue (for example, no circle at all, black X... )

Just so everyone can be sure we do not want this, here is the color parade :clown_face: 
grafik

elioqoshi:
One alternative would be to have the icons show up all the time and the coloured circle background to appear on hover, basically like Windows does it. 

image

Muqtxdir:
Hi @ubuntu/yaru-team , @jbouter, @reC0D3D   👋

Sorry little-late to this.
The new changes to window-buttons are really fresh but little too much like macOS (which many people for **Yaru** may hate, saying from as I have experimented with a similiar change in the past for **Yaru-remix** and many people were not pleased with it being another macOS clone theme and so I had to revert it and branch it-out as **Yaru-macOS**)

From top of my mind a safer-approach to this, I suggest we bring-up orange-color when on **hover** or **active** in this way the accesibility issue will be rectified and branding will also be kept (_See below screen-record for demonstration_ )

https://user-images.githubusercontent.com/54065734/116055812-b82c0880-a69a-11eb-8c2f-a7c8c1cd102b.mp4


Thanks!

Muqtxdir:
> One alternative would be to have the icons show up all the time and the coloured circle background to appear on hover, basically like Windows does it.
> 
> image

@elioqoshi just looked at your suggestion now, omg we both suggested the same thing (_deja vu_  😊 )

Feichtmeier:
Since this got brought up, linking this discussion/PR for orange only on hover: https://github.com/ubuntu/yaru/pull/2713

elioqoshi:
@Muqtxdir exactly what I was referring to, this looks great.

> @elioqoshi just looked at your suggestion now, omg we both suggested the same thing

Hah indeed.

I'd still argue that brand shouldn't be part of the window control icons. The red should be a semantic red we use for destructive actions and not the Ubuntu orange.

Feichtmeier:
Kind of funny how we can see things from all perspectives :D 
Even though I argues in the PR differently, I am now feeling @clobrano argument a little bit more, that without the orange dot, it looks very much Adwaitaish in this corner of the window

grafik
:shrug: 

Maybe Adwaitaish or Macosish is not really a strong argument. If it works it works.

Feichtmeier:
> We could eventually also discuss the spacing between the 3 buttons as well. I am curious about the reason why historically they have so much spacing. Is this a Yaru or GNOME thing?

It's the upstream spacing

Muqtxdir:
> Since this got brought up, linking this discussion/PR for orange only on hover: #2713

🤔 Somehow I was not aware of this, otherwise it would have been a +1 from me

madsrh:
> ...without the orange dot, it looks very much Adwaitaish...

+1 We've been here before 😅 In the non orange version I still miss that Ambiance /Ubuntu feeling.

clobrano:
> Kind of funny how we can see things from all perspectives :D
> Even though I argued in the PR differently, I am now feeling @clobrano argument a little bit more, that without the orange dot, it looks very much Adwaitaish in this corner of the window
> 
> grafik
> shrug
> 
> Maybe Adwaitaish or Macosish is not really a strong argument. If it works it works.

that feeling is even stronger if you try the light Yaru. What about a red circle that is filled only on hover?

Muqtxdir:
> What about a red circle that is filled only on hover?

### Desrtuctive-color: red on window-close button

 (_based on @elioqoshi suggestion_)


https://user-images.githubusercontent.com/54065734/116062374-7488cd00-a6a1-11eb-9754-a47ac4eb7a7e.mp4

Feichtmeier:
:shrug: I don't know

I can't really say what it is but unless I know what it is I share carlos opion on adwaita. Maybe there is still too much in the theme going on that reminds of adwaita and the orange close button is the last barriert to see this? 
Maybe it would work better if we would modernize yaru even further, for example removing the strong headerbar bottom border, or making all headerbar buttons transparent (yaru 1.0 etc). 

So it is def. a solution to the issue, but it does not appeal visually. If the cross would be bigger maybe

Feichtmeier:
This thread brought me into yaru and I always return to this explosion of creativity (https://discourse.ubuntu.com/t/design-mockups-updates-designer-only/2053/31) 

What color does the X have @madsrh ?

madsrh:
> This thread brought me into yaru and I always return to this explosion of creativity

Ahh those were the days ☺️ I'm AFK so can't tell or remember, but it looks more red than orange 🤷‍♂️

jbouter:
Hi @Muqtxdir! :wave:

> From top of my mind a safer-approach to this , I suggest we bring-up orange-color when on hover or active in this way the accesibility issue will be rectified and branding will also be kept (See below screen-record for demonstration )

Bit late to the party here, but I do agree with you. I mentioned this in the issue by comparing it to how the Breeze theme handles it. I do really like this approach, and it definitely resolves the issue for me. But then again, I can't decide on yaru branding.

Jupi007:
Just tested this, sorry for my late answer:

- At first glance, it looks very good and fresh.
- But, I think it will be disturbing for many people, especially for computer novice.

IMO the best design could be the orange dot on hover, just like @Muqtxdir did on its screen recording.

wgx:
One quick thought on discoverability: ideally controls _should_ be immediately apparent without hover - you could argue that an operating system GUI is a special case: the users will be exposed to it all day, every day, and have a real opportunity to learn "hidden" affordances - which is why Apple can get away with hiding the icons on the window buttons, and nobody minds too much.

That said, I _do_ like the version with the icons always visible then the colour adding additional information when hovered - it feels more usable, and doesn't look overly cluttered to me.

Feichtmeier:
Yes totally, we are just iterating over several issues.
We decided to move out of this PR for discussion currently and will update as soon as we've settled down with different options. :) 
At this point it is rather unlikely that what you will see in this PR will go live

Feichtmeier:
Closing as the branch is horribly outdated and the design pushed won't happen.

PR #2809, opened 2021-04-26
Gtk inverted headerbar only ssd

Jupi007:
Capture d’écran du 2021-04-26 12-08-19

@Feichtmeier Your idea could be good, but I personally think that it is weird to have apps with a dark headerbar and other apps with a white headerbar.

@ubuntu/yaru-team What do you think of?

### DIFF
--- a/gtk/src/default/gtk-3.20/_apps.scss
+++ b/gtk/src/default/gtk-3.20/_apps.scss
@@ -123,11 +123,11 @@ $ambiance: null !default;
 
     .path-bar-box {
       border-radius: 5px;
-      border: 1px solid if($ambiance, darken($headerbar_bg_color, 9%), $borders_color);
-      &:backdrop { border-color: if($ambiance, $headerbar_bg_color, $backdrop_borders_color);}
-      background-color: if($ambiance, lighten($headerbar_bg_color, 6.5%), $bg_color);
+      border: 1px solid $borders_color;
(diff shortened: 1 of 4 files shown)


Feichtmeier:
I think this would be very good solution, and makes it even more obvious that a headerbar is not always a titlebar :) Which was imo a false assumption when Ubuntu switched to CSD windows 
Also tagging @waynecrosby for a branding opinion

clobrano:
Sorry, but I didn't get what this solution is for 😅

Feichtmeier:
> Sorry, but I didn't get what this solution is for 😅

https://github.com/ubuntu/yaru/discussions/2677

Problem 6

ubuntujaggers:
I'm open minded 🤔 Need to try. Will flutter apps have a header bar?

clobrano:
>Problem 6: the headerbar colors are not exported at the top of the stylesheet of gtk!

Technically you're right that titlebars are not headerbars, but from a user point of view I believe they are seen as the same thing. I don't really think mixing colors in the top window will look good. This will probably just backfire with bugs of windows not styled "correctly", either people complaining that they selected light and got dark, or vice versa.

Feichtmeier:
Yeah you are right @clobrano :) 
Thanks @Jupi007 for investigating though. I hope it did not take you too much time?

PR #2816, opened 2021-04-30
Update Gtk2 assets render scripts

Jupi007:
@Jannomag I found a solution for Inkscape 1.0:

- modify all **assets.svg** files: replace `inkscape:pageopacity="1"` with `inkscape:pageopacity="0"`
- edit all scripts: drop `--export-background-opacity=0` and also replace `--export-png=$ASSETS_DIR/$i.png` by `-o $ASSETS_DIR/$i.png`

That's it :)

Closes #2815

Issue #2815, opened 2021-04-29
No transparency when rendering gtk2 assets for default and light variant


Jannomag:
Hi,
I'm currently working on an update for my Yaru-Color theme.    
I noticed that the assets for each gtk2 variant are having transparent backgrounds.    

When I render them using the render-all-assets.sh script, the assets won't get rendered with transparent background if there is a background in assets.svg.     
This happes for light and default variant ONLY. Dark seems to be okay, even though there's a background behind those assets in the assets.svg.    

I attached two files to compare.    
    
This one is directly from yaru master in ./gtk/src/gtk-2.0/light/assets directory
![menu-pan-down](https://user-images.githubusercontent.com/31586674/116536741-98d4eb80-a8e5-11eb-92a5-76fddff469e0.png)

    
This one is rendered using render-all-assets.svg within the ./gtk/src/gtk-2.0/light directory. 
![pan-down](https://user-images.githubusercontent.com/31586674/116536079-cbcaaf80-a8e4-11eb-8164-4f940e4ccf16.png)    

Nothing was modified, I downloaded the master branch as zip and tried it with those fresh files.    

Steps to reproduce:     
- clone or download the repro
- head over to ./gtk/src/gtk-2.0/light/
- check the assets in ../assets, for example "pan-down.png", it's a transparent png
- execute the render-all-assets.sh script
- check the same asset file again, it's not transparent anymore.

Jannomag:
Reason found: with Inkscape 0.92.5 it's working.

Are you able to fix this?
Because it's needed to compile older Inkscape on >18.04 manually

Feichtmeier:
No idea at this department but @clobrano and @Jupi007 recently worked on the script

clobrano:
We did not work on `render-all-assets.sh`, that's why isn't working with inkscape 1.0. Weird that it works fine in dark variant though, it makes me wonder if there is a difference in the input asset.

Jannomag:
Okay. I played a bit with inkscape's cli but couldn't find any alternative commands. `--export-background-opacity` is still valid.

@clobrano @Jupi007 
But I've found another bug. I have absolutely no clue about github and pull requests and before I do something wrong I post it here.

In the assets.svg of dark gtk2.0 variant is the "scale-slider" wrong.
Missing "scale-slider" in dark assets.svg of gtk2 theme.

Fix:
Replace lines 2404 to 2408:
```
       id="g1967"
[...]

PR #2727, opened 2021-03-17
Sync gtk2 assets with Gtk3/4

Jupi007:
- Fix blurry assets
- Update all checkbox/radio
- Update slide handles and tracks

PR #2817, opened 2021-04-30
fixed assets.svg in gtk2 dark

Jannomag:
I fixed some wrong lines in assets.svg for dark gtk2 variant. This issue causes the scale-slider.png not being rendered.

### DIFF
--- a/gtk/src/dark/gtk-2.0/assets.svg
+++ b/gtk/src/dark/gtk-2.0/assets.svg
@@ -2402,10 +2402,9 @@
          style="opacity:1;fill:none;fill-opacity:1;stroke:#b56d54;stroke-width:1;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1" />
     </g>
     <g
-       id="g1967"
-       inkscape:export-filename="/home/jupiter007/yaru/gtk/src/dark/gtk-2.0/assets/scale-slider.png"
-       inkscape:export-xdpi="96"
-       inkscape:export-ydpi="96">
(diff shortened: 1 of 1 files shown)


Muqtxdir:
Hi @Jannomag 👋, 
Thanks for taking time to contribute to yaru with a PR

@Jupi007 could you take a look and review at this PR as you both are trying for the same issue #2815. I think this has a chance that this clashes with your PR #2816

Jupi007 (approved):
Thanks @Jannomag for this PR.
Everything looks in place.

But remember that all Gtk2 __assets.svg__ aren't uptodate with current rendered assets pngs. They need a sync.

Anyway, thanks again :)

Jannomag:
That means that the assets.svg files aren't containing the correct assets which are already pre rendered?

Jupi007:
> That means that the assets.svg files aren't containing the correct assets which are already pre rendered?

Yes, rendered pngs does not comes from the source file. Some assets are blurry inside it btw.
In fact, I fixed blurry images sometimes ago (#2727) but all __assets.svg__ files was already not synced, so I manually exported all pngs directly with the Insckape export dialog.
So all these __assets.svg__ files need a sync, but that represent a lot of work and Gtk2 is "EOL" so I never taken the time to do that task (and the manually export works great for that kind of little fix).

The problem is that you want to modify in a large way all assets to modify colors so that's a really problem for you.
If I have time, maybe I could have a look to the sync.
Can you open an issue to remind me, please?

Issue #2403, opened 2020-10-09
[wishlist] Snap symbolic icon


fthx:
Hi,

I was writing a snap manager extension for Gnome and I noticed that there's no snap symbolic icon. Could it be created even if it does not exist in Adwaita?

fthx:
Ok, I created a transitional one:
![image](https://user-images.githubusercontent.com/23138504/95677133-bb125080-0bc3-11eb-9c6d-e733fba3e37c.png)

I do what I can. I don't know how to make an SVG. And Inkscape snap isn't working. :-p

fthx:
Looks better in panel:
![image](https://user-images.githubusercontent.com/23138504/95677164-f280fd00-0bc3-11eb-8e92-f2136cfccd83.png)

Feichtmeier:
Awesome. I think we can add it. Could you make a PR? Thanks

fthx:
Sorry for my limited knowledge but I don't know how to to that.
What are the steps to achieve that? I tried but without success... "I do what I can." :-)

Muqtxdir:
hi @fthx,

sorry for being this late, how does this look?


![snap-symbolic](https://user-images.githubusercontent.com/54065734/95677433-2ddef400-0be3-11eb-85c6-8ad216bc70c5.png)


zip file containing the svg : [snap-symbolic.zip](https://github.com/ubuntu/yaru/files/5360991/snap-symbolic.zip)

fthx:
Looks great to me:
![image](https://user-images.githubusercontent.com/23138504/95677591-d3379f00-0bc6-11eb-87a5-0bca2e3903f4.png)
Better than mine, for sure! Happy to see a faster-than-light icon creation!

fthx:
You can find the new extension version including your icon here:
https://extensions.gnome.org/review/download/19853.shell-extension.zip

fthx:
I'm here for another issue but I see this one still open.
The icon from @Muqtxdir is very good, is it now included in Yaru icon theme?

fthx:
Bump?

Muqtxdir:
> The icon from @Muqtxdir is very good, is it now included in Yaru icon theme?

hey @fthx sorry for the late reply, could you please tell what should be the name given to the snap symbolic?

I was under the assumption that the snap symbolic was only for your extension, but it seems this needs to be included in yaru as well (am I right?)

fthx:
Hi,

[...]

PR #2835, opened 2021-05-14
Add symbolic Snap app and mimetype icons

elsiehupp:
Fixes https://github.com/ubuntu/yaru/issues/2403.

I based the icon name on the name of [the corresponding fullcolor icon](https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/mimetypes/application-vnd.snap.svg).

The primary change I made when scaling down the full-size icon was thickening the negative-space lines and aligning them to the pixel grid so that the icon will look sharper when rendered on screen. There are a number of side-by-side previews in the linked Issue.

Please let me know if there's anything else I need to do for the pull request. (I'm new here.) Thanks!

### DIFF
--- a/icons/Suru/scalable/apps/snap-symbolic.svg
+++ b/icons/Suru/scalable/apps/snap-symbolic.svg
@@ -1,0 +1,7 @@
+<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
+<svg fill="#808080" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
+ <path d="m0 0 8 3v5l-8-8z"/>
+ <path d="M8.3,9.7L2,16L6.2,7.6L8.3,9.7Z"/>
+ <path d="M9,3.946L9,9L12.675,5.325L9,3.946Z"/>
+ <path d="m9 3h6l1 2.625-7-2.625z"/>
+</svg>
--- a/icons/Suru/scalable/mimetypes/application-vnd.snap-symbolic.svg
+++ b/icons/Suru/scalable/mimetypes/application-vnd.snap-symbolic.svg
@@ -1,0 +1,8 @@
+<!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'>
+<svg clip-rule="evenodd" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round" stroke-miterlimit="1.5" version="1.1" viewBox="0 0 16 16" xml:space="preserve" xmlns="http://www.w3.org/2000/svg"><g fill="#808080" fill-rule="nonzero">
+    <path d="m3 3 5 2v3l-5-5z"/>
+    <path d="M7.8,9.2L4,13L6.533,7.933L7.8,9.2Z"/>
+    <path d="M9,6L9,8L10.429,6.571L9,6Z"/>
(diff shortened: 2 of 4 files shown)


Jupi007 (changes requested):
Hi @elsiehupp,
thanks a lot for your PR.

The SVG itself is great.

But there's a few more steps to do:
- That not very important, but we usually use a mid-grey `#808080` for our symbolic icons.
- Also you only added your icon into the `/src` folder. You must to use the render script which will export a minimized version into `/Suru`. Run it like this: `./render-symbolic-icons.sh -c mimetypes -f application-vnd.snap-symbolic.svg`

Thanks :)

elsiehupp:
Okay, cool! The version I initially posted just didn’t have color data to begin with, but it was easy enough for me to add it to the SVG.

Is it all good, now?

elsiehupp:
Actually, one thing: It looks like the norm is for mimetypes to be inside outlines, so this should probably be an application icon, not a mimetype icon.

elsiehupp:
I moved the existing icon into “apps” and added a separate mimetype icon based on the existing norm of roundrect outlines.

Jupi007 (approved):
Looks good for me, thanks for this @elsiehupp 👍

I have an off topic question: what do you use to edit your SVG?

Muqtxdir:
@elsiehupp had a look at this branch, the icons look really good and nicely aligned.

elsiehupp:
> Looks good for me, thanks for this @elsiehupp 👍
> 
> I have an off topic question: what do you use to edit your SVG?

I use [Affinity Designer](https://affinity.serif.com/designer/) on macOS. I used to use Illustrator, but Affinity had a 50% “work from home” sale at the beginning the pandemic, and Affinity has a pay-once business model, whereas Adobe has a monthly subscription business model.

I’ve been wading through lots of raw SVG lately, and Affinity seems to generate much cleaner output than Inkscape, Illustrator, or Vectornator. Inkscape uses a bunch of SVG extensions in order to use SVG as its primary format, but the result is that there’s a lot of extraneous stuff left over in the raw SVG. Affinity doesn’t support any of the more clever programmatic features of SVG (I think Inkscape might, but I’m not sure), but it’s still fairly good as a starting point.

As an aside, doing my illustration work on macOS is why I’m getting [pesky `.DS_Store` files](https://github.com/ubuntu/yaru/pull/2836) everywhere. It might be because I’m using [Netatalk](http://netatalk.sourceforge.net/) instead of Samba, idk.

> @elsiehupp had a look at this branch, the icons look really good and nicely aligned.

Thanks! 😊

Jupi007:
@elsiehupp Thanks for your answer :)
In fact I ask this because as you seems to want to contribute, you must use Inkscape because we use some specifics Inkscape's metadata and any other software will brake the workflow.
But don't worry, it is just for fullcolor icons, symbolics ones are not affected.

elsiehupp:
> @elsiehupp Thanks for your answer :)
> In fact I ask this because as you seems to want to contribute, you must use Inkscape because we use some specifics Inkscape's metadata and any other software will brake the workflow.
> But don't worry, it is just for fullcolor icons, symbolics ones are not affected.

You mean like [this](https://github.com/linuxmint/mint-y-icons/pull/285)? (I haven’t actually checked to see if you’re using the same Python script, but I’d be shocked if you weren’t.)

elsiehupp:
[It was, in fact, the same Python script.](https://github.com/ubuntu/yaru/pull/2839)

Jupi007:
@ubuntujaggers I think we can merge this, wdyt?

elsiehupp:
FYI one of the icons here [uses stroke width instead of fill](https://github.com/ubuntu/yaru/pull/2843#issuecomment-841855455), so idk if it will work properly.

Jupi007:
Sorry, I didn't tell you earlier that using stroke is problematic.
Can you make a patch PR?

PR #2836, opened 2021-05-14
Add macOS .DS_Store to .gitignore

elsiehupp:
macOS will automatically add [`.DS_Store`](https://en.wikipedia.org/wiki/.DS_Store) files when you browse to a location in the Finder. It even does this if you are browsing a remote server that isn't a Mac (which is admittedly annoying). Adding `.DS_Store` to `.gitignore` eliminates the annoyance of having to manually exclude these files from Git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ stage/
 prime/
 snap/.snapcraft/state
 *.snap
+
+# filesystem clutter
+.DS_Store


clobrano:
A better approach for this problem is to use a `global gitignore` file, where listing all environment files that are unnecessary to the project.

Usually this is done saving a gitignore file somewhere out of the project (e.g. `$HOME/.global_gitignore`) with all the files you want to ignore regardless the specific project and then configure git to look for it

```
git config --global core.excludesFiles /home/<name>/.global_gitignore
```

GIT will look first into this file and then in the project specific gitignore

PR #2837, opened 2021-05-14
Add VS Code workspace to .gitignore

elsiehupp:
Adding this lets individual developers save personalized VS Code workspace configuration files within the project directory (which is the norm).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ stage/
 prime/
 snap/.snapcraft/state
 *.snap
+
+# Developer config files
+*.code-workspace


clobrano:
Hey, thank you for the PR. I'd say VS code is pretty popular and it does not hurt to block list it, however I think that there might be a better option. GIT allows to setup a global gitignore file (litteraly `.gitignore_global`) under your `$HOME` directory which adds block list items without polluting the specific projects.

Feichtmeier:
Personally I also always include vscode stuff in gitignore https://www.toptal.com/developers/gitignore helps a lot
Often if I create templated projects in java and dart a .gitignore is auto created and already includes .vscode etc

clobrano:
Yeah, but with the global gitignore you can add vscode stuff only once for any project you're working on

Feichtmeier:
Sounds like the perfect solution if you know this exists. Maybe for people that don't, vscode in the yaru gitignore is the "safe" way to prevent people to commit this stuff =)

clobrano:
Yeah, it's for people that don't know this that I wrote it XD
vscode adds itself to gitignore because it doesn't know where the global gitignore is, but Ok if you want to add .vscode no big deal

Feichtmeier:
No hard feelings either :)
I think often vscode asks in a newly opened dir if you want to add the .vscode folder into the workspace or not

elsiehupp:
One reason to use the per-repository `.gitignore` for the VS Code workspace is that there are plausible reasons for certain repositories to want to include it. I’ve set up a global `.gitignore` previously, though, so I could look into adding this to it, as well.

Feichtmeier:
@clobrano what shell we do with this?

clobrano:
I don't think we need this, I prefer the global gitignore way

PR #2838, opened 2021-05-14
Add miscellaneous symbolic mimetype icons

elsiehupp:
After I got started with the [Snap symbolic icon](https://github.com/ubuntu/yaru/pull/2835) I decided to work my way down the list and make symbolic versions of a bunch of mimetype icons. Here is an initial batch that I feel is basically ready. Some of them could probably stand to be finessed.

One thing of note: I used portrait-orientation roundrect outlines for these icons in order to be consistent with the fullcolor mimetype icons. In some cases it could work to use a “document” outline instead, but I haven’t gotten to any of those yet.


elsiehupp:
I’m not sure how the shell script is set up, but at some point I could probably combine these into a single large SVG for the `src` folder, which would help in terms of viewing them all at once.

Jupi007:
> I’m not sure how the shell script is set up, but at some point I could probably combine these into a single large SVG for the `src` folder, which would help in terms of viewing them all at once.

You're right, that was the precedent organization of symbolics. But that was cause many and many git merge clash and that's why we exploded it into multiple files. (see #2419)

elsiehupp:
FWIW it would presumably be possible to automatically compile a “contact sheet” of the symbolic icons as part of `render-symbolic-icons`, but I don’t know Ruby.

Jupi007 (changes requested):
I took the time to look at your PR. First at all, big thanks, that's a huge work and it looks awesome!

Here is a few notes to polish your icons and make them a bit more Yaruish:

- We don't use gradient with our symbolics. You should remove them from `application-json-symbolic` and `application-pgp-signature-symbolic.svg`.
- When you want to make some part of the icon transparent, please use only 50% of opacity.
- We usually use square shape for stroke-linecap.
- The text lines from `application-pgp-signature-symbolic.svg ` looks horizontally misaligned.
- For the `application-vnd.flatpak-symbolic.svg` icon I think you should use the same emblem as the fullcolor one (which looks fully flat).
- `application-x-cd-image-symbolic.svg` looks strange. You could use the same pictogram as here: #2213
- The lock pictogram on `application-pgp-encrypted-symbolic.svg` looks like the Adwaita one. Can you try to use the same as `rotation-locked-symbolic.svg`?
- Also maybe add a transparent dot on the `application-geo+json-symbolic.svg`?

Just one more question: where these icons are used (I mean did you already seen them into an application)?

Thanks a lot :)

Jupi007:
@ubuntujaggers can you have a look at this, please?
Maybe you'll see things I haven't seen (or also maybe I said wrong things 😅).

elsiehupp:
> Just one more question: where these icons are used (I mean did you already seen them into an application)?

I was just working my way down the list and making symbolic versions of existing icons. It isn’t actually based on any particular use case.

A lot of the design quirks you’ve picked up actually come from maintaining fidelity to the non-symbolic icon the designs were based on. Improving consistency is definitely part of what I meant by “finessing”, though. I will try and make changes based on your feedback, though.

A few things worth noting:
* I’ve been using a pair of consistent outlines for the mimetype icons. It might make sense for me to submit the outlines as “template” files. I had used roundrects with arc corners for sharper rendering, but I could try using more “squircle”-shaped outlines for comparison.
* It would probably make sense to have a set of design guidelines for symbolic icons as a Markdown file in the `scalable` folder in the repository. I can throw something together based on your feedback and some of the more portable guidelines from [GNOME/Adwaita](https://developer.gnome.org/hig/stable/icon-design.html.en)

> **NOTE:** The proceeding two bullet points could go in a separate pull request.

* Specifically regarding stroke-linecap, from the standpoint of sharper rendering, it makes sense to use square caps only when the line termination is orthoganal to the pixel grid. Diagonal line terminations would probably render more sharply with round stroke-linecaps, but it’s possible to use béziers in some circumstances to force orthogonal terminations.

ubuntujaggers:
These are beautifully clear :) - but...  does Ubuntu actually use symbolic mimetypes anywhere? :thinking: 

In terms of stuff like gradients and different levels of transparency, symbolics are a bit unusual because they take the local text colour, apart from elements that have a xml property like "warning" set - which take other theme colours.  I know that full opacity and 50% opacity are preserved (and the current icon set uses both) but have no idea what other stuff will do.  Worth noting: symbolics are very often inverted because of the recolouring (i.e., white on black background) so it's worth testing more complex designs on both to check they look good both ways.

If symbolic mimetypes are used I say make Jupiter's amendments and then merge :+1:

elsiehupp:
@ubuntujaggers Again, this is entirely speculative work on my part (i.e. for my own amusement). However, in doing this work I’ve found that a number of Yaru’s full-color icons have very poorly pixel-hinted graphics in their smaller sizes, so it could be beneficial to back-port some of these designs to the full-color set, which actually does see significantly more use.

My understanding with symbolic icons is that GNOME discards color information (except for, as you mentioned, certain special colors), but keeps opacity. So *I think* gradients would work, but I haven’t actually tested them. (There are really only a handful of symbolic icons where gradients make sense.)

Jupi007:
I never seen an app (or even an icon theme) with mimetypes symbolic icons, expect a few very basic ones like `text-x-generic-symbolic` or `x-office-document-symbolic`. That's why I asked where they are used.
But I also think it can't hurt to add these into our repo (except the extra size), so why not? 🤷

About the gradients, as there's no symbolic icons into Yaru which use them, I think we should avoid to use them. And also, I don't think that will looks good enough at 16px especially with the dark theme.

ubuntujaggers:
> a number of Yaru’s full-color icons have very poorly pixel-hinted graphics in their smaller sizes

Yep, they're a real mixed bag - PRs for full colour icons with better executions in the smaller sizes would be very welcome indeed!! 🙂

Muqtxdir:
Hi @elsiehupp 

Just to repeat what others have already stated, GNOME does not use `symbolics` for mimetypes except the ones which we include with yaru. So I don't think there's any point in putting effort in creating `symbolics` which are never going to used.

However If you are interested in contributing to the Icon-theme you could help us in completing the missing icons from this issue: #3116  (_there's quite a lot_)

PR #2839, opened 2021-05-14
Simplify parsing to work with non-Inkscape SVGs

elsiehupp:
I copied the relevant bits from [my open pull request](https://github.com/linuxmint/mint-y-icons/pull/285) over at `mint-y-icons`. I tested it, and it seems to work.

(If you’re curious, the biggest difference with the version of `render-bitmaps.py` over at `mint-y-icons` is that they replaced all the `"` with `'`, which is the norm if you use `pylint`. Other than that, I think they added a few failsafes.)

### DIFF
--- a/icons/src/fullcolor/render-bitmaps-verbose.py
+++ b/icons/src/fullcolor/render-bitmaps-verbose.py
@@ -23,15 +23,15 @@ import xml.sax
 import subprocess
 import argparse
 
-INKSCAPE = '/usr/bin/inkscape'
-OPTIPNG = '/usr/bin/optipng'
-MAINDIR = '../usr/share/icons/Mint-Y'
-SOURCES = ['apps', 'categories']
+INKSCAPE_PATH = '/usr/bin/inkscape'
+OPTIPNG_PATH = '/usr/bin/optipng'
+OUTPUT_DIRECTORY = '../../Suru'
+CATEGORIES_TO_RENDER = ['apps', 'categories']
 
 # assert ('linux' in sys.platform), "This code runs on Linux only."
-assert (os.path.isfile(INKSCAPE)), "Expected to find Inkscape at /usr/bin/inkscape, but file does not exist."
-assert (os.path.isfile(OPTIPNG)), "Expected to find OptiPNG at /usr/bin/optipng, but file does not exist."
-assert (os.path.isdir(MAINDIR)), "Expected to find Mint-Y at ../usr/share/icons/Mint-Y, but directory does not exist."
+assert (os.path.isfile(INKSCAPE_PATH)), "Expected to find Inkscape at " + INKSCAPE_PATH + ", but file does not exist."
+assert (os.path.isfile(OPTIPNG_PATH)), "Expected to find OptiPNG at " + OPTIPNG_PATH + " but file does not exist."
+assert (os.path.isdir(OUTPUT_DIRECTORY)), "Expected to find output directory at " + OUTPUT_DIRECTORY + ", but directory does not exist."
 
 # the resolution that non-hi-dpi icons are rendered at (may be 90 or 96 depending on your inkscape build)
 inkscape_version = str(subprocess.check_output(['inkscape', '-V'])).split(' ')[1].split('.')
@@ -49,8 +49,8 @@ inkscape_process = None
 def main(args, SRC):
 
     def optimize_png(png_file):
-        if os.path.exists(OPTIPNG):
-            process = subprocess.Popen([OPTIPNG, '-quiet', '-o7', png_file])
+        if os.path.exists(OPTIPNG_PATH):
+            process = subprocess.Popen([OPTIPNG_PATH, '-quiet', '-o7', png_file])
             process.wait()
 
     def wait_for_prompt(process, command=None):
@@ -69,7 +69,7 @@ def main(args, SRC):
             output = output[1:]
 
     def start_inkscape():
-        process = subprocess.Popen([INKSCAPE, '--shell'], bufsize=0, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+        process = subprocess.Popen([INKSCAPE_PATH, '--shell'], bufsize=0, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
         wait_for_prompt(process)
         return process
 
@@ -79,9 +79,9 @@ def main(args, SRC):
             inkscape_process = start_inkscape()
 
         cmd = [icon_file,
-               '--export-dpi', str(dpi),
-               '-i', rect,
-               '-e', output_file]
+               '--export-dpi={}'.format(str(dpi)),
+               '--export-id={}'.format(str(rect)),
+               '--export-filename={}'.format(output_file)]
         wait_for_prompt(inkscape_process, ' '.join(cmd))
         optimize_png(output_file)
 
@@ -175,7 +175,7 @@ def main(args, SRC):
                         if dpi_factor != 1:
                             size_str += "@%sx" % dpi_factor
 
-                        dir = os.path.join(MAINDIR, self.context, size_str)
+                        dir = os.path.join(OUTPUT_DIRECTORY, self.context, size_str)
                         outfile = os.path.join(dir, self.icon_name+'.png')
                         if not os.path.exists(dir):
                             os.makedirs(dir)
@@ -184,7 +184,7 @@ def main(args, SRC):
                         if self.force or not os.path.exists(outfile):
                             inkscape_render_rect(self.path, id, dpi, outfile)
                             if args.verbose:
-                                print("├─ Rendered new \"".decode('utf-8') + outfile + "\"")
+                                print("├─ Rendered new \"" + outfile + "\"")
                             new_renders += 1
 
                         # If PNG exists, compare modify time to that of SVG
@@ -196,27 +196,27 @@ def main(args, SRC):
                             if stat_in.st_mtime > stat_out.st_mtime:
                                 inkscape_render_rect(self.path, id, dpi, outfile)
                                 if args.verbose:
-                                    print("├─ Rendered updated \"".decode('utf-8') + outfile + "\"")
+                                    print("├─ Rendered updated \"" + outfile + "\"")
                                 # print("Rendered updated " + outfile)
                                 updated_renders += 1
 
                             # If PNG is newer than SVG, leave PNG as is
                             else:
                                 if args.verbose:
-                                    print("├─ \"".decode('utf-8') + outfile + "\" is newer than SVG")
+                                    print("├─ \"" + outfile + "\" is newer than SVG")
                                 skipped_renders += 1
                 if args.verbose:
-                    print("├────────────────────────────────┤".decode('utf-8'))
+                    print("├────────────────────────────────┤")
                 if args.svg is None:
                     print("")
-                    print("┌────────────────────────────────┐".decode('utf-8'))
-                    print("│ Directory: ".decode('utf-8') + self.context)
-                    print("│ Icon Name: ".decode('utf-8') + self.icon_name)
-                    print("├────────────────────────────────┤".decode('utf-8'))
-                    print("├─ Rendered %d new PNGs".decode('utf-8') % new_renders)
-                    print("├─ Rendered %d updated PNGs".decode('utf-8') % updated_renders)
-                    print("├─ Skipped %d up-to-date PNGs".decode('utf-8') % skipped_renders)
-                    print("└────────────────────────────────┘".decode('utf-8'))
+                    print("┌────────────────────────────────┐")
+                    print("│ Directory: " + self.context)
+                    print("│ Icon Name: " + self.icon_name)
+                    print("├────────────────────────────────┤")
+                    print("├─ Rendered %d new PNGs" % new_renders)
+                    print("├─ Rendered %d updated PNGs" % updated_renders)
+                    print("├─ Skipped %d up-to-date PNGs" % skipped_renders)
+                    print("└────────────────────────────────┘")
                     print("")
 
         def characters(self, chars):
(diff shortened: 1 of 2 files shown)


Jupi007:
Wow, great.
Thanks a lot for this!

I'll check that tomorrow :)

Jupi007:
@clobrano I asked your review because I don't know python at all. But if you don't have the time there's no problem.

clobrano (approved):
From the point of view of Python code, this seems very good (and the old code was terrible :D)
I can't say nothing from the point of view of parsing SVG file however, because I am not familiar with it.

Jupi007:
> From the point of view of Python code, this seems very good (and the old code was terrible :D)
> I can't say nothing from the point of view of parsing SVG file however, because I am not familiar with it.

Thanks for the review, I'll do the functional test :)

elsiehupp:
> From the point of view of Python code, this seems very good (and the old code was terrible :D)
> I can't say nothing from the point of view of parsing SVG file however, because I am not familiar with it.

I deliberately avoided making any changes other than incorporating the looser SVG parsing code. I actually did get carried away making changes to the version over on Mint-y-Icons, though I specifically separated out the SVG parsing changes there, too, because I wanted to get them merged more quickly. (It’s been like three months, and they still haven’t been merged. \*shrugs\*)

FWIW I made the changes largely by trial and error because the original code is a bit impenetrable to me, too, and I’m not hugely familiar with Python beyond the basics.

Jupi007 (changes requested):
The script works (the icon is rendered as expected), but there's something wrong: it is very very long because it seems render too many things:

```console
$ ./render-bitmaps.py configure
Rendering SVG "configure.svg" in ./actions
actions configure
......................................................................
```
Each point is a render, and here is the normal script return:
```console
$ ./render-bitmaps.py configure
Rendering SVG "configure.svg" in ./actions
actions configure
..........
```

elsiehupp:
@Jupi007  Thank you for bringing this to my attention! If the script is otherwise working as it’s supposed to, it might be more likely the case that the console output is printing extra points that don’t correspond with actual renders. I can do some more testing myself, as well as digging in the code to see what’s up.

FWIW I actually had [an earlier pull request over at Mint-y-Icons](https://github.com/linuxmint/mint-y-icons/pull/284) that made the console output more explicit. I could try and port some of that over here, too, but the changes are a bit more substantial.

elsiehupp:
I just added the more verbose version of the script (completely unchanged from my Mint-y-Icons pull request), so you can test it if you want. Among other things it should allow you to more easily verify what the script is actually doing, since I added the verbose output in the process of developing the loosened XML parsing.

elsiehupp:
It’s also possible that the reason for the script apparently generating excess exports is that I messed up in trying to merge my changes without merging any of the underlying differences between the scripts, hence why I’ve provided the fully verbose version of the script for comparison.

Jupi007:
Looks like your new verbose script isn't compatible with Yaru:

```console
$ ./render-bitmaps-verbose.py configure
Traceback (most recent call last):
  File "/home/jupiter007/yaru/icons/src/fullcolor/./render-bitmaps-verbose.py", line 34, in <module>
    assert (os.path.isdir(MAINDIR)), "Expected to find Mint-Y at ../usr/share/icons/Mint-Y, but directory does not exist."
AssertionError: Expected to find Mint-Y at ../usr/share/icons/Mint-Y, but directory does not exist.
```

🤔 Maybe I'm wrong but it seems want to push icons directly into `/usr/share/icons/Mint-Y`.

elsiehupp:
Well, that should be fixable!

elsiehupp:
I did some tweaks, and it should work, in theory, except that it doesn’t work with Inkscape 1.0 or later, and I can’t quite figure out how to install Inscape 0.92. (I tried following the instructions [here](https://inkscape.org/release/inkscape-0.92.5/gnulinux/ubuntu/ppa/dl/), and it didn’t work, probably because I’m not on Ubuntu <20.04.)

Part of the problem seems to be that Inkscape's CLI interface apparently changed pretty dramatically from 0.92 to 1.0, though I can’t quite figure out what changes broke the script.

The script prints a stream of errors like this:
```
verbs_action: Invalid verb: ./apps/screenshot-app.svg --export-dpi=96 --export-id=rect24x24 --export-filename=../../Suru/apps/24/screenshot-app.png
```
…but if I just enter:
```bash
$ inkscape ./apps/screenshot-app.svg --export-dpi=96 --export-id=rect24x24 --export-filename=../../Suru/apps/24/screenshot-app.png
```
…it works perfectly fine.

The changes to the CLI options are described on the wiki [here](https://wiki.inkscape.org/wiki/index.php/Using_the_Command_Line#Changes_from_0.92), but I can’t figure out what’s wrong.

Jupi007:
That's indeed a problem because we all use Inkscape 1.0+.
Our current script support the 1.0, maybe you could have a look at it?

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -0,0 +1,297 @@
...
+#
+# Thanks to the GNOME icon developers for the original version of this script
+import os
+import sys
+import xml.sax
+import subprocess
+import argparse
+
+INKSCAPE_PATH = '/usr/bin/inkscape'
+OPTIPNG_PATH = '/usr/bin/optipng'
```
clobrano, line 27: Since we are reworking this, I would suggest to get rid of the full path and just look for the binary wherever it is installed in the system. This way we should also be able to use snap/flatpak install.

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -0,0 +1,297 @@
...
+OUTPUT_DIRECTORY = '../../Suru'
+CATEGORIES_TO_RENDER = ['apps', 'categories']
+
+# assert ('linux' in sys.platform), "This code runs on Linux only."
+assert (os.path.isfile(INKSCAPE_PATH)), "Expected to find Inkscape at " + INKSCAPE_PATH + ", but file does not exist."
+assert (os.path.isfile(OPTIPNG_PATH)), "Expected to find OptiPNG at " + OPTIPNG_PATH + " but file does not exist."
+assert (os.path.isdir(OUTPUT_DIRECTORY)), "Expected to find output directory at " + OUTPUT_DIRECTORY + ", but directory does not exist."
```
clobrano, line 34: I think this could be refactored in some kind of loop, avoiding to repeat the same code

elsiehupp:
Try `./render-bitmaps-verbose.py —help`

I could further remove the default variables and instead use an external shell script to load them, which would make the python script completely reusable.

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -238,60 +221,278 @@ def characters(self, chars):
...
+    print("│ - status                       │")
+    print("│ - wip                          │")
+    print("└────────────────────────────────┘")
+    print("")
+    return
+
+
+def build_parser():
+
+    parser = argparse.ArgumentParser(add_help=False)
```
clobrano, line 350: I love the effort and the help output is wonderful, however it seems a bit maintenance unfriendly :) Any change in the arguments will require to update the help message too. Was the default help message so bad?

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -238,60 +221,278 @@ def characters(self, chars):
...
-    print("\nSuccessfully processed \"" + args.svg + "\" in \"" + success_directory + "\".\n")
-elif success_directory == "" and args.svg is not None:
-    print("\nFailed to process \"" + args.svg + "\" in " + success_directory + ".\n")
-elif success_directory != "":
-    print("Successfully processed listed sources.\n")
-elif success_directory == "":
-    print("Failed to process listed sources.\n")
-else:
-    raise Exception("Conditional statement falls through.")
+# This is the main function
```
clobrano, line 497: If this is the main entry, you should add the usual

```python
if __name__ == '__main__':
````

File: icons/src/fullcolor/render-bitmaps-verbose.py [hunk 5]
```diff
@@ -238,60 +221,278 @@ def characters(self, chars):
...
+    parser.add_argument(
+        '--help',
+        action='store_true',
+        help="show this help message and exit"
+    )
+    parser.add_argument(
+        '--inkscape_path',
+        type=str,
+        nargs=1,
+        default = '/usr/bin/inkscape',
```
clobrano, line 392: I do not think there is really the need to use default paths for inkscape (same thing for optipng). Since Python 3.3 shutil library has a `which` implementation, that you can use to verify whether the program is installed or not.

https://docs.python.org/3.9/library/shutil.html#shutil.which

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -238,60 +221,278 @@ def characters(self, chars):
...
+                # xml.sax.parse() uses the various ContentHandler() methods behind the scenes
+                #
                 xml.sax.parse(open(file), handler)
         
         return True
 
-    
 
-parser = argparse.ArgumentParser(description='Render icons from SVG to PNG')
+def check_dependencies(args):
```
clobrano, line 244: using https://docs.python.org/3.9/library/shutil.html#shutil.which this really simplifies in something like

```python
for dependency in ["inkscape", "optipng"]:
    if not shutil.which(dependency):
        print(f"Please install {dependency}")
```

which in my system results in
`Please install optipng`

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -238,60 +221,278 @@ def characters(self, chars):
...
+
+    # assert ('linux' in sys.platform), "This code runs on Linux only."
+    assert (os.path.isfile(args.inkscape_path)), "Expected to find Inkscape at " + args.inkscape_path + ", but file does not exist."
+    assert (os.path.isfile(args.optipng_path)), "Expected to find OptiPNG at " + args.optipng_path + " but file does not exist."
+    assert (os.path.isdir(args.output_path)), "Expected to find output directory at " + args.output_path + ", but directory does not exist."
```
clobrano, line 249: At line 226 `args.output_path` is created if it does not exist. Is this assert correct?

icons/src/fullcolor/render-bitmaps-verbose.py
```diff
@@ -1,234 +1,217 @@
...
+            #
+            # ContentHandler() implements an interface for xml.sax.parse()
+            #
+            handler = ContentHandler(file, args, True)
+            #
+            # xml.sax.parse() uses the various ContentHandler() methods behind the scenes
+            #
             xml.sax.parse(open(file), handler)
             return True
         else:
```
clobrano, line 217: there is no need for an `else` branch after `return`

clobrano (commented):
Hey, thanks a lot for this work!

icons/src/fullcolor/render-bitmaps-verbose.py
elsiehupp: That’s probably a holdover from something else. (The assertions are from when I was initially debugging the version of the script over at `Mint-y-icons`.)

icons/src/fullcolor/render-bitmaps-verbose.py
elsiehupp: Looking at the code, the `else` branch happens if the user specifies the name of an icon name that doesn’t exist. Or something. Again, basically all the comments and the `print()` statements are from when I was debugging the initial version of the script, so there are a fair number of extra failsafes built in. I should probably try running this through `pylint`.

elsiehupp: Well, it looks like the `—categories` parser is broken so I wasn’t able to reach that `else` branch quite yet:
```bash
$ ./render-bitmaps-verbose.py --categories apps --individual_icons thisappdoesnotexist
┌────────────────────────────────┐
│ Rendering from command-line argument "thisappdoesnotexist"
├────────────────────────────────┤
Source path "./a" does not exist.
Source path "./p" does not exist.
Source path "./p" does not exist.
Source path "./s" does not exist.
└────────────────────────────────┘

Failed to process "thisappdoesnotexist" in .
```

clobrano: Considering the output above, even if good looking, the logs can be formatted without those borders. What do you think?

elsiehupp: You mean the ASCII “box” outlines? I specifically found they helped with visual parsing extremely long sections of output. Like, CMake has weird specific formatting, too, for the purpose of visually parsing the (extremely verbose) output.

The issue in the above is that it’s parsing “apps” as “a” “p” “p” “s”.

icons/src/fullcolor/render-bitmaps-verbose.py
elsiehupp: The default help message was kind of a mess specifically because it doesn’t like longer argument names. I could presumably make the formatting of the custom help message more programmatic, but it might take a little bit.

icons/src/fullcolor/render-bitmaps-verbose.py
elsiehupp: Thanks for the link. I’ll add in the `which`, but I will probably leave in the arguments as a backup for if `which` doesn’t work or the user otherwise wants to call a custom executable.

icons/src/fullcolor/render-bitmaps-verbose.py
clobrano: Yes, I agree they help a little, but as well it is easy to break like the following
```
┌────────────────────────────────┐
│ Rendering from command-line argument "thisappdoesnotexist"
├────────────────────────────────┤
```

which is not really nice to see.

I mean, there are python modules to handle such prints, but it seems overkill to me. A better option would be to move to a proper logger, but it is better to move it to a different PR

clobrano: > The issue in the above is that it’s parsing “apps” as “a” “p” “p” “s”.

sorry if you already detected it, but this is because `args.category` is defined as single str value

```python
parser.add_argument(
        '--categories',
        type=str,
        nargs='?',
```

`nargs` should be `*` to get a list of zero or more values

elsiehupp:
I’ve been working on [a version of this](https://github.com/elsiehupp/render-svg-icons) separately, and I got it [uploaded on PyPI](https://pypi.org/project/render-svg-icons/). It would be great if you could help test it out regardless, though!

I'm closing this pull request because I'm working on the code elsewhere.

Jupi007:
Thanks @elsiehupp 
I'll try it :)

PR #2843, opened 2021-05-16
Symbolic icon guidelines and templates

elsiehupp:
This is a draft pull request with two sets of template icons. Basically, uh, I’d like the maintainers to choose one over the other before this gets merged.

I see there are other shape templates in [`source-symbolic.svg`](https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/source-symbolic.svg), so I could work with that, too.

Also I’d like to flesh out the icon design guidelines in the [`README.md`](https://github.com/ubuntu/yaru/blob/master/icons/src/scalable/README.md).


elsiehupp:
One thing I’m noticing from the existing symbolic icons is that they seem somewhat optimistic about what’s actually visible in a 100% 16x16 render.

Anyway, here are the template images at 16x16:

template-document-arc-symbolic template-roundrect-arc-symbolic template-document-bezier-symbolic template-roundrect-bezier-symbolic

The bézier versions are indeed a bit fuzzy, but they’re also more consistent with the “Yaru style”.

One big difference between these and the similar versions in existing icons is that these use curves with a stroke width rather than outline curves with a fill. Using a stroke width helps keep the icons more easily editable, and I’m kind of suspicious that the outline curves with fill used elsewhere are the product of an infelicitous format conversion.

A possible consideration with the curves-with-stroke-width approach is potentially using 32x32 SVGs instead of 16x16 SVGs, because otherwise in order to fit within a pixel grid the curve points have to fall at half-pixels, which is more difficult to do with snapping in most illustration programs. From the perspective of the rendering engine, I’m fairly certain the nominal dimensions of the SVGs don’t actually matter.

elsiehupp:
Aaaaand it looks like I accidentally based this branch off of my other one. That’ll be fun to untangle!

ubuntujaggers:
Templates are a great idea, it's certainly nicer than robbing an existing symbolic and deleting stuff out of it :)

One thing we should do is make sure the folded corners are aligned (or vice-versa) with the current symbolic for folder-documents-symbolic, which currently has a filled corner.  So whichever is preferred, we can either fill the corner here or un-fill the corner on folder-documents-symbolic.

The bezier versions IMO are a bit too rounded at this size - as long the corners aren't actually sharp, I think it's round enough for Yaru!

Couple more thoughts:

_"One thing I’m noticing from the existing symbolic icons is that they seem somewhat optimistic about what’s actually visible in a 100% 16x16 render."_

Don't forget, Gnome sometimes uses the symbolics at larger sizes - e.g.:

image

image

_"Using a stroke width helps keep the icons more easily editable, and I’m kind of suspicious that the outline curves with fill used elsewhere are the product of an infelicitous format conversion."_

There are certainly a lot of cases where a "stroke" is actually a filled element, but that's just my bad practice - I don't like working with strokes, and for a long time it was mostly just me doing icons, so I didn't have to worry muh about other people editing them.  Definitely agree that strokes are better, so maybe we're due a cleanup.

elsiehupp:
The bézier outlines I used in the template drafts are specifically based on the actual bézier handles themselves, which actually meet at the corners. It’s clearer in this screenshot:

Bézier handles

Using only four points, one centered on each side, and having the bézier handles meet in the corners makes the shape extremely simple to define, at basically any aspect ratio.

By contrast, here’s the stroke I extracted from [an existing symbolic icon](https://github.com/ubuntu/yaru/blob/23c154fcf55d14ca802d7111febfcb70d90cf048/icons/src/scalable/apps/rhythmbox-symbolic.svg):

Existing symbolic

Whereas here’s the stroke I extracted from [one of the high color templates](https://github.com/ubuntu/yaru/blob/c7d93582d5671799f20a7abc5827a8166a1f255a/icons/src/fullcolor/Vertical%20Oblong%20App%20Icon%20Template.svg):

Existing highcolor

And here’s a version of the stroke from the high-color template where I’ve removed the unnecessary side points and snapped the remaining béziers to the pixel grid:

Highcolor rationalized

The main advantage I see with the version I used is that it can easily be adapted to pretty much any size or aspect ratio, whereas using a shape where only the corners are rounded requires some degree of discretion in terms of how to scale the size of the corners.

Regarding the filled folded page corner, it might on some grander level be worth shifting in the direction of consistently avoiding fill. Adwaita-symbolic really likes fill. Breeze-symbolic really likes outlines. Yaru-symbolic… seems kind of unsure about it? This is well beyond the scope here, though.

Regarding icon sizers, yes, symbolic icons can definitely be used at higher resolutions (never mind higher DPI). But the way the icon theme is built currently uses a single asset for all sizes of a given symbolic icon. I don’t actually know how XDG icon themes work, but it may be possible to do different asset for different sizes, the way full-color icons usually work.

If indeed symbolic icons end up having different assets for different sizes, it may be worth renaming the parent folder from “scalable” to “symbolic”. It may be worth doing that regardless, though, just for clarity. Again, that question is somewhat beyond the scope here.

Jupi007:
Remember that (and I know how it is annoying to working with 😓) gnome doesn't support stroke at all and we must convert them into path.

elsiehupp:
> Remember that (and I know how it is annoying to working with 😓) gnome doesn't support stroke at all and we must convert them into path.

Lovely. Would it at least be possible to automate that with the Ruby script?

ubuntujaggers:
> Remember that (and I know how it is annoying to working with 😓) gnome doesn't support stroke at all and we must convert them into path.

Well thank God for me not liking paths 🤣

Jupi007:
> Lovely. Would it at least be possible to automate that with the Ruby script?

I think so (just a note: the Ruby script is obsolete, we currently use the bash script), but I personally prefer see the result into Inkscape directly by converting with `stroke-to-path`.
But that is only my personal point of view :)

elsiehupp:
I’m really not familiar with Inkcape’s API. It looks like [someone has already made a utility using that function](https://github.com/leifgehrmann/svg-stroke-to-path), though, so perhaps we could borrow that into the shell script?

ubuntujaggers:
"Regarding the filled folded page corner, it might on some grander level be worth shifting in the direction of consistently avoiding fill. Adwaita-symbolic really likes fill. Breeze-symbolic really likes outlines. Yaru-symbolic… seems kind of unsure about it? This is well beyond the scope here, though."

I think this would be a big change because, if you look at the Nautilus sidebar, a lot of the symbols use fill. Now you've mentioned it, I'm wondering if the folder symbol should have a bit of fill on the top left... or as you say, vice-versa. The approach there would be to do a test branch (probably starting with the Nautilus symbols, which are nice and conspicuous) so the team could try it without fill.

In terms of this PR I think we can probably take the folded corners out of the equation, because I think we came to the conclusion that Ubuntu doesn't use symbolic mimetypes - so we wouldn't need a template for them?

elsiehupp:
@Muqtxdir yes, I kind of lost my momentum with this initiative anyway. At the moment, the aspect of this that would probably be my biggest priority is working on the rendering scripts, though I’m more working on the full color script than the symbolic one.

PR #2845, opened 2021-05-16
Convert stroke to fill in Snap symbolic mimetype

elsiehupp:
This is supposed to be a patch on https://github.com/ubuntu/yaru/pull/2835, but I don’t think I did this right.


Jupi007 (approved):
Looks good for me, thanks :)

Issue #2800, opened 2021-04-23
redesign files icon


kothekarswaraj:
**Expected Behavior**
files- icon must be like home icon on the desktop(with the house icon). It should have a rectangle foundation with a small upper left part coming out of the rectangle. It already has all those things but I am talking about that square background removal I think it would really blend more right? white suits the icon.

**Actual Behavior**

**Steps to Reproduce the Problem**

**Software that presents the issue**

**Yaru version**
yaru used in ubuntu hirsute hippo 21.04

**Screenshots**

Feichtmeier:
Only gonna happen if we ever find a true Suru style replacement. No adwaita hardcopies. No big blue file cabinets.
If you have an actual idea that still fits exactly in our icon theme guidelines, shoot

Feichtmeier:
Closing this issue in favour of https://github.com/ubuntu/yaru/issues/2964 where the discussion popped out again. Maybe we find a Suru/Yaru style icon everyone can agree on

Issue #1755, opened 2020-01-16
Proposal: filemanager icon


Feichtmeier:
I think I did not make a secret out of my opinion on the blue file cabinet icon
![image](https://user-images.githubusercontent.com/15329494/72531631-cd946f00-3869-11ea-8929-bf47fee28dcc.png)

I really think it does not fit into that whole origami style of all the other nice icons you've made @ubuntujaggers 

I spoilered the idea  on telegram but here - for the "rest" - this is the idea:

![image](https://user-images.githubusercontent.com/15329494/72531810-2bc15200-386a-11ea-9afb-328a2f3fd545.png)

Basically the "old" / Suru icon but on one of the new Yaru baseplates.
I am not sure about the brown though, but it's prbly a safe pick.

Or a white one:
![image](https://user-images.githubusercontent.com/15329494/72532151-c883ef80-386a-11ea-884b-4cc6cade02ca.png)

Or... *insert any other baseplate color here*

Opinions, onions and whatnot? @madsrh @ubuntujaggers @clobrano ?

Just for the history on why we changed it: we completely removed the baseplate back then and it was only a single white folder as an icon - then someone complained (and this was a correct complain) that it would only represent a directory and not a file manager)

clobrano:
All the "complains" still hold for me:
- the folder icon was common under Ubuntu, but someone pointed out that it looks like a real folder, which is not
- the cabinet icon was not common under Ubuntu, so someone -at first- did not notice nautilus was there, just with a different icon

I am okay with both really, my only note is to take one and be consistent, better not going back and forth every other release :)

Feichtmeier:
Or make the current one look more like the rest of the icons :man_shrugging: 
Most of them only have 1 or 2 basic symbol on them. Very simple. I think the current one is too much. Too many highlights too many elements (3 handles 3 drawers) which make it look more "glossy" than the other icons.
I am not saying this is the best fit - I just do not think the current one is one of the better ones.

@clobrano most suru icons give the "context" by putting a symbol on a baseplate. 
Baseplate: square
Symbol: folder

clobrano:
Really, my only point was to decide for one - old or new - if we all are happy and can live long with it, or keep the current otherwise, I am more than open to proposals.

Feichtmeier:
Ok great! Just to get it more in line with the suru/yaru icons and to not some ideas on how those icons are made:

- they clearly should remind of paper or concrete: origami
[...]

PR #2846, opened 2021-05-16
Redesign Files Icon

elsiehupp:
This is a quick mockup of a concept for maybe fixing https://github.com/ubuntu/yaru/issues/2800. Thoughts? Is this design concept worth pursuing?

### DIFF
--- a/icons/src/fullcolor/apps/filemanager-app-draft.svg
+++ b/icons/src/fullcolor/apps/filemanager-app-draft.svg
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg width="100%" height="100%" viewBox="0 0 256 256" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" xmlns:serif="http://www.serif.com/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;">
+    <g id="_256x256" serif:id="256x256">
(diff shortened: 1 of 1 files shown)


ubuntujaggers:
I think we got told off for changing the Nautilus icon too much 🙂 The grey square with the white folder possibly isn't our most popular icon, but it's probably settled for a little while. There are definitely some icons in the theme that could be improved though, I sort of hate the "characters" app icon with the smiley face on it...

elsiehupp:
FWIW the reason I did the stack of three folders is that it more closely brings to mind [the official GNOME Nautilus icon](https://commons.wikimedia.org/wiki/File:GNOME_Files_icon_2019.svg), which is a file cabinet with three drawers:



(I can’t find for the life of me the source repository that includes this icon, hence why I’m linking to the mirror on Wikipedia.)

For that matter, it wouldn’t be terribly difficult to do a mashup of the above icon and the [Yaru Vertical Oblong Icon Template](https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/Vertical%20Oblong%20App%20Icon%20Template.svg):



Would that be more palatable, do you think?

elsiehupp:
Actually, the Wikipedia page links to [the Nautilus Gitlab](https://gitlab.gnome.org/GNOME/nautilus/blob/master/data/icons/hicolor/scalable/apps/org.gnome.Nautilus.svg).

Muqtxdir:
Hi @elsiehupp 👋,

Little late to this 
the current-icon is based on the discussions in the [issue](https://github.com/ubuntu/yaru/issues/1755).

elsiehupp:
> Hi @elsiehupp 👋,
> 
> Little late to this
> the current-icon is based on the discussions in the [issue](https://github.com/ubuntu/yaru/issues/1755).

No worries, and thanks for the explanation! I mainly took the initiative here because there was the open issue on the subject, not because I have any particularly strong preference on the matter.

(My main critique of the existing icon is that it’s a bit ambiguous, but it doesn’t really bother me that much.)

Issue #2849, opened 2021-05-17
Switches in selected rows have the wrong styling


Feichtmeier:
![photo_2021-05-17_14-50-52](https://user-images.githubusercontent.com/15329494/118491428-48e88800-b71f-11eb-9da8-c9260bd6efe8.jpg)

https://github.com/ubuntu/yaru/blob/7fba4e50952a081932d844524f060d0b61ff6db1/gtk/src/default/gtk-3.20/_tweaks.scss#L435-L449

Jupi007:
Fixed by #2850

PR #2850, opened 2021-05-17
Fix row:selected buttons

### LINKED ISSUE #2847: Outline radius on .circular buttons is wrong
grafik

https://github.com/ubuntu/yaru/blob/e021a6b90dad596a8f3eaa2d2d7ffb9249a41694/gtk/src/default/gtk-3.20/_tweaks.scss#L223-L230

Jupi007:
Capture d’écran du 2021-05-17 15-51-54

I'll do the patch for GTK4 inside the Libadwaita PR to avoid merging clash.

Closes #2847

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -678,19 +678,19 @@ button {
   }
 
   @at-root %button_selected, & {
-    row:selected & {
-      @if $variant == 'light' { border-color: $selected_borders_color; }
-    }
+    // row:selected & { // Yaru change: due to grey row selection background we no more need to modify buttons border-color
+      // @if $variant == 'light' { border-color: $selected_borders_color; }
+    // }
 
     @at-root %button_selected_flat, &.flat {
-      row:selected & {
-        &:not(:active):not(:checked):not(:hover):not(disabled) {
-          color: $selected_fg_color;
-          border-color: transparent;
+      // row:selected & { // Yaru change: due to grey row selection background we no more need to modify border-color and color
+        // &:not(:active):not(:checked):not(:hover):not(disabled) {
+          // color: $selected_fg_color;
+          // border-color: transparent;
 
-          &:backdrop { color: if($variant=='light', $backdrop_base_color, $backdrop_fg_color); }
-        }
-      }
+          // &:backdrop { color: if($variant=='light', $backdrop_base_color, $backdrop_fg_color); }
+        // }
+      // }
     }
   }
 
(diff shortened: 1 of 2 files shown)


Feichtmeier (commented):
Hm, I rather thought about a :not(.circular) selector to only change the radius of all other buttons, shouldnt this be enough?

Jupi007:
Sorry I misunderstood your issue 😓
That I fixed here is the orange border and the dark icon on selected rows.
I didn't noticed the white outline...

In fact the buttons doesn't use the `.circular` selector but a `.btn_circular`, btw in the dev version, it looks like they moved to `.circular`: https://github.com/bottlesdevs/Bottles/commit/4a2572d51409c53ca473d9aabffc49e4707e546d#diff-e144420ba74b5637b04111e3de9b54f8c9451a2c626eac32b946b9218bff9a34
That fix this issue :)

Feichtmeier:
Oh okay 😊
It's still too much diff in common I think. Can't we somehow leave this in tweaks?

Jupi007:
> It's still too much diff in common I think. Can't we somehow leave this in tweaks?

The problem is that these lines overwrite the border of all widgets on `row:selected`. It is very simpler to comment them directly.
Move this to tweaks will need to re-overwrite the border to the default value. But I think it is the better solution if we want one day to drop some tweaks.

Well, stop complaining, I'll try to do this :)

Feichtmeier:
Right, okay. Just keep in mind that it needs to stay maintainable. Feel free to merge when you are done

Feichtmeier:
I mean, we could also move away from _tweaks completely. If we mark each yaru change in common.

Jupi007:
> I mean, we could also move away from _tweaks completely. If we mark each yaru change in common.

That could be simpler in some case but more complicated in other case.

Here we have the same problem as the inverted headerbar: if I want to re-overwrite the button border looks to default, I must restyle the whole button stack. So it's really better to just comment some lines in _common.

Feichtmeier:
There are several things in _tweaks that are really just nitpicks in my opion and rarely seen in apps, maybe we could have a dedicated discussion on this topic, I would prefer for libadwaita to stop any yaru modifications except _colors.scss

Jupi007:
> There are several things in _tweaks that are really just nitpicks in my opion and rarely seen in apps, maybe we could have a dedicated discussion on this topic, I would prefer for libadwaita to stop any yaru modifications except _colors.scss

Okay 👍

About this PR, unless you really want to move the patch to _tweaks (but that will need to add a mammoth part of code), I think I'll merge this as it.

Feichtmeier:
Yes, go for it

Jupi007:
Btw this PR also fix switches in selected rows.

Capture d’écran du 2021-05-18 19-28-39

Closes #2849

PR #2783, opened 2021-04-09
Sync Gtk4 with Libadwaita

Jupi007:
This is that I understand: Gtk theme is renamed to Default (was Adwaita) and will be simplified. So Libadwaita (which is a Gtk4 port of Libhandy) is now a superset of Gtk with more widgets. It is named Adwaita.
I compared Libadwaita scss with the Gtk one and it is exactly the same. Libadwaita just add more code for his own widgets.
So, Adwaita = Gtk default + Libadwaita (libhandy).

---

- Use Libadwaita repo as target for Gtk4 Actions
- Update Gtk4 upstream folder
- Import Libadwaita scss
- Reapply Yaru changes
- Import and update Gtk3 Libhandy Yaru changes
- Fix some problems

PR #2855, opened 2021-05-19
Better list row border left

Jupi007:
I used a gradient instead of an inset box-shadow, so there's no more rounded corner.

**Before:**

Capture d’écran du 2021-05-19 20-40-08

**After:**

Capture d’écran du 2021-05-19 20-35-26

The Gtk4 part will be added to the Libadwaita PR (#2783).


Feichtmeier:
Awesome 😎👍

Feichtmeier (approved):

PR #2869, opened 2021-05-27
Gtk3 theme: remove firefox border axing

Feichtmeier:
- it is no longer needed and this could cause issues once the fire next-UI-menus are in (forgot how that UI overhaul is named, but on windows firefox nightly the new menus are already in and they are big and round)

grafik


Jupi007 (approved):
Nice 👍

I tested this with nightly and with the default version and I didn't noticed any difference with Yaru master.
I don't think we need to wait the landing of the new Firefox's UI to merge this.

PR #2871, opened 2021-05-27
Make GDM widgets compliant with GS theme variant

### LINKED ISSUE #2870: Gnome shell login screen dark/light mix looks weird and could be unified
The mix of the super dark text input (gnomers call it "entry") plus the white buttons look kind of strange to me
grafik

Thinking of two solutions if you agree on the problem:
- do not override gnome shell login screen buttons to be white
- make text inputs white

@madsrh @clobrano @Jupi007 ?

Jupi007:
- Remove entry style overwrite on dark theme (was really too dark). Now it use the same background and border color as dash, search entry, popover, ...
- GDM buttons didn't respect theme variant. Now it use `$base_color` so this problem is solved.
- I also changed the fg color for buttons as they now use  `$base_color`.

Closes #2870

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_login-dialog.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_login-dialog.scss
@@ -11,27 +11,25 @@
   border: none;
   background-color: transparent;
 
-   // Yaru: we force the light colors because GDM doesn't change on shell theme switch
-  $_gdm_bg: $porcelain;
-  $_gdm_fg: $inkstone;
+  $_gdm_bg: $base_color; // Yaru change: use $base_color for dark/light theme switch
 
-  StEntry {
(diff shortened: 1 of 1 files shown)


Feichtmeier (approved):

Muqtxdir:
@Jupi007 do I need to enable the `gresource` option in meson for this ?

Feichtmeier:
you can install user themes and then switch to the light shell to test it

for the log in screen there is nothing you can do I think, except change the build to force a light theme.
the dark theme is already testable for both log-in and lock-screen

Jupi007:
Nope, just switch GS theme variant, or restart the shell with `alt+F2` and `r`.
Then lock the screen and enjoy :)

For the login screen, since I switched to GS 40 with shemgp ppa, I can't use yaru theme (I tried the method from `CONTRIBUTING.md` but that don't work).

PR #2881, opened 2021-06-03
Gtk3 Nautilus list view change

### LINKED ISSUE #2822: Nautilus listview: Improve the gray selection or revert to orange
I feel annoyed by the selection colour in list view in Nautilus. It is furthermore super-hard to distinguish the colour of selected items from the colour of the item that is hovered with the mouse.

**Expected Behavior**

I would expect to distinguish the selected items much more, and I would expect the theme to be coherent between list view and icon view in Nautilus.

**Actual Behavior**

In icon view, selected items have text with orange highlighting colour, and are clearly distinguishable from non-selected items. The hovered item (the one that resides under the mouse cursor) is very slightly lighter, hardly noticeable, but at least enough for me. 

In list view however, selected items have grey highlighting colour that is lighter than hovering colour, which is disturbing. Furthermore, this light colour is hardly distinguishable from white depending on the illumination outside of the screen (for instance if you are outside in the sun with your laptop). While I miss the orange highlighting colour, I would consider myself happy with the newer grey highlighting colour if it were darker (for better contrast). 

I would suggest to let the hovering colour as is, and make the selection colour even darker. I believe grey has a good potential, but a darker one.

Just open Nautilus and highlight a few items in list view. Compare with doing the same in icon view.

Ubuntu 21.04 with GNOME 3.38.4 and Nautilus 3.38.2

**Yaru version**

yaru-theme-gtk, version: 21.04.1

**Screenshots**

Below are two screenshots where I selected the folders "Langues", "Lecture" and "Modèles", and hovered "Documents" without selecting.

Capture d’écran de 2021-05-05 14-28-08
Capture d’écran de 2021-05-05 14-33-22

**Upstream check**

This bug does not occur with Adwaita. Only Yaru.

Feichtmeier:
- only color the selected row gray in backdrop

Peek 2021-06-03 17-29

Closes #2822

### DIFF
--- a/gtk/src/default/gtk-3.20/apps/_nautilus.scss
+++ b/gtk/src/default/gtk-3.20/apps/_nautilus.scss
@@ -91,20 +91,11 @@
             border-style: none;
         }
 
-        $_selection_bg: if($variant=='light', darken($bg_color, 5%), lighten($base_color, 5%));
-
         treeview {
             &:selected {
-                background-color: $_selection_bg;
-                color: $fg_color;
-
-                &:hover {
-                    background-color: if($variant=='light', darken($_selection_bg, 5%), lighten($_selection_bg, 5%));
-                }
-
                 &:backdrop {
                     color: $backdrop_fg_color;
-                    background-color: $_selection_bg;
+                    background-color: if($variant=='light', darken($bg_color, 5%), lighten($base_color, 5%));
                 }
             }
         }


Jupi007 (approved):
Looks good, it is important to clearly see that we have selected 👍

madsrh:
Good call @Feichtmeier 
This makes sense 👍

jlemonde:
I'm super happy with those changes :) Thank you for your work!

An end-user.

PR #2889, opened 2021-06-07
Add Gtk4 and Libadwaita install infos

Jupi007:
After days and days of trying and searching, I found all dependencies of Gtk4 and Libadwaita 😅
So I updated `CONTRIBUTING.md` with these new information.

@Feichtmeier That could be great if you can try this.

Gtk4 dev deps:
```console
sudo apt install sassc clang libglib2.0-dev libcairo2-dev libpango1.0-dev libgdk-pixbuf-2.0-dev libepoxy-dev libxkbcommon-dev libgraphene-1.0-dev cmake libwayland-dev libxrandr-dev libxi-dev
```

Libadwaita dev deps:
```console
sudo apt install sassc valac libglib2.0-dev
```

Both use the same **meson/ninja** command to compile:
```console
meson --prefix=/usr --sysconfdir=/etc --localstatedir=/var --libdir=lib/x86_64-linux-gnu _build .
cd _build
ninja
sudo ninja install
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,22 +41,26 @@ update-alternatives --install /usr/share/gnome-shell/theme/gdm3.css gdm3.css /us
 
 If you reboot, select the Yaru session in GDM and you should be all right.
 
-### Help to install Gtk4
+### Help to install Gtk4 and Libadwaita
 
-If you want to test the Gtk4 theme, you need to compile Gtk from source (awaiting an official ubuntu packaging).
+If you want to test the Gtk4 theme, you need to install both Gtk4 and Libadwaita.
+
+#### Install Gtk4
+
+**Note:** If you don't want to install Gtk4 from source, just install `libgtk-4-dev` package (Ubuntu > 21.04 only).
 
 Firstly clone the repository:
-```bash
-git clone git@gitlab.gnome.org:GNOME/gtk.git
+```console
+git clone https://gitlab.gnome.org/GNOME/gtk.git
 ```
 
 Then, you need to install dependencies (please report if more ones are needed):
-```bash
-sudo apt install libgraphene-1.0-dev
+```console
+sudo apt install sassc clang libglib2.0-dev libcairo2-dev libpango1.0-dev libgdk-pixbuf-2.0-dev libepoxy-dev libxkbcommon-dev libgraphene-1.0-dev cmake libwayland-dev libxrandr-dev libxi-dev
 ```
 
 And finally compile using meson:
-```bash
+```console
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Uhhh awesome - will check this out today and give you a review!

Feichtmeier:
Gtk 4 worked and also apps were installed to /usr/share/applications
grafik

libdwaita not but I could start it from the command line, but using the adwaita theme :thinking: 
grafik

CONTRIBUTING.md
```diff
@@ -41,22 +41,26 @@ update-alternatives --install /usr/share/gnome-shell/theme/gdm3.css gdm3.css /us
...
 Then, you need to install dependencies (please report if more ones are needed):
-```bash
-sudo apt install libgraphene-1.0-dev
+```console
+sudo apt install sassc clang libglib2.0-dev libcairo2-dev libpango1.0-dev libgdk-pixbuf-2.0-dev libepoxy-dev libxkbcommon-dev libgraphene-1.0-dev cmake libwayland-dev libxrandr-dev libxi-dev
 ```
 
 And finally compile using meson:
-```bash
+```console
```
Feichtmeier, line 63: eventually add a ```cd gtk``` here :D

CONTRIBUTING.md
```diff
@@ -65,6 +69,28 @@ sudo ninja install
...
+git clone https://gitlab.gnome.org/GNOME/libadwaita.git
+```
+
+Then, you need to install dependencies (please report if more ones are needed):
+```console
+sudo apt install sassc valac libglib2.0-dev
+```
+
+And finally compile using meson:
+```console
```
Feichtmeier, line 85: maybe cd libadwaita here :)

Jupi007:
> libdwaita not but I could start it from the command line, but using the adwaita theme thinking

It is normal, just like libhandy, you can only run `adwaita-1-demo` using the console (I'll add a note about this).
Also, I don't know why, but every app which use Libadwaita force the Adwaita theme, you must switch to Yaru with the Gtk inspector.

CONTRIBUTING.md
Jupi007: Good idea 👍

Feichtmeier:
@Jupi007 
ah yes this works :+1: 
I think, even though your gradient idea was good, this stripe does not work anymore in gtk4, we should remove it from tweaks for gtk4, don't you think?
grafik

Jupi007:
> I think, even though your gradient idea was good, this stripe does not work anymore in gtk4, we should remove it from tweaks for gtk4, don't you think?

It was looking very good when no margin nor border radius (nautilus).
Actually that looks a bit odd, so OK for removing.

PR #2890, opened 2021-06-07
Gtk4Theme: remove outdated tweaks

Feichtmeier:
Many improvements in the gtk4 theme we base yaru on make changes for the following widget stylings not needed anymore:

- menu border/shadow
- popover border/shadow
- menu item hover
- menubar item hover
- checks and radios inside menus
- keep the focus ring transition from upstream
- remove the gray+orange stripe selection for gtk4, because their new rounded gray row works fine for yaru, but with the orange rounded stripe it does not
- remove button selected comment block, because they removed this block themselves
- keep slider scrollbar width
- keep the viewswitcher border
- keep the bright window outline

All changes were achieved by copying the files from the upstream to src dir and only keeping the wanted changes


Feichtmeier:
@Jupi007 do you know where we overwrite the very nice new bright window border from gtk4? We really need this :D

Jupi007 (approved):
Nice you did it 👍

> @Jupi007 do you know where we overwrite the very nice new bright window border from gtk4? We really need this :D

I don't remember anymore why 😅
Can you revert that into this PR?

Feichtmeier:
Added more reverts (see the commit)
Result:
grafik
grafik

gtk/src/default/gtk-4.0/_drawing.scss
```diff
@@ -9,10 +9,10 @@
...
 // If $within is true, use focus-within instead of focus:focus-visible
 //
-@mixin focus-ring($target: null, $width: 2px, $offset: -$width, $outer: false, $focus-state: 'focus:focus-visible', $fc: $focus_border_color) { // Yaru: fix missing buttons transition
-  // transition-property: outline, outline-width, outline-offset, outline-color;
-  // transition-duration: 300ms;
-  // animation-timing-function: ease-in-out;
+@mixin focus-ring($target: null, $width: 2px, $offset: -$width, $outer: false, $focus-state: 'focus:focus-visible', $fc: $focus_border_color) {
+  transition-property: outline, outline-width, outline-offset, outline-color;
+  transition-duration: 300ms;
+  animation-timing-function: ease-in-out;
```
Jupi007, line 15: I remember commented these lines because that was disabling state transitions for buttons.
But it should be better to add a `button` selector on the transition tweak.

Feichtmeier: This is this jumping focus ring animation. 
I check if this is still the case. I guess button transitions are more important than the focus ring transition
Apropos transitions.... The transition for checks looks weird for the new menu checks, the little asset just shakes a little bit ^^ can we limit this change for all checks outside of menus?

Jupi007: As `$button_transition` target all properties, add the transition tweak for buttons will restore state transitions and keep the focus ones 😉

About menu's checks transitions, the `not()` selector only works for the direct parent, so we must disable transitions with something like this:

```scss
popover.menu {
  check,
  radio {
    transition: none;
  }
}
```

Feichtmeier: Peek 2021-06-07 23-28

Button transitions work fine, which one do you mean?

Feichtmeier: ah btw, upstream changed backdrop to be removed but somehow, somwhere they change the opacity in the backdrop instead - this does work with adwaita but not with yaru, only the headerbars child widgets opacity changes, I doubled checked if we got all changes and I could not find any other place :thinking: very strange

Jupi007:
Checks, switchs, sliders, etc. haven't got any transitions.
Do you know why they removed almost all animations in Libadwaita?

gtk/src/default/gtk-4.0/_drawing.scss
Jupi007: Just checked `adwaita-1-demo` with adwaita theme and only the headerbar opacity change on backdrop.
Inside the widget-factory, the backdrop opacity works everywhere.
I think something is broken with Libadwaita upstream.

Feichtmeier:
No idea, it must be something with the new backdrop trick.
They completely removed backdrop CSS and the backdrop effect is somewhere else applied - since it is still happening which you can see if you select default or adwaita theme 🤷

Jupi007:
Libadwaita has dropped :backdrop, but Gtk not.
Remember that the Gtk factory doesn't use the new Adwaita theme from Libadwaita which Yaru is based on.

Feichtmeier:
hmmmm so everything is fine from our side? :thinking:

Jupi007:
I think so. I could be an upstream bug (or not if it is the normal behavior ^^).

Feichtmeier:
hmmm...
GitHub kinda lagging right now, telegram or discord availible for you?

Jupi007:
I can't recover my Telegram access (I asked support but no response).
Discord works fine however.

PR #2891, opened 2021-06-07
Use correct placeholder for workspace previews

LinuxLlama:
wsthumbs_new

The upstream shell theme has a placeholder when dragging a window in-between two workspace previews. This is not the case on Yaru. This tiny change properly renders the workspace preview on Yaru shell.


Feichtmeier:
Uhhh, nice catch! Tyvm @LinuxLlama !

PR #2902, opened 2021-06-12
Gtk4 symlink to upstream

Jupi007:
- Move all tweaks to _tweaks:
  - Green levelbar ;
  - Inverted headerbar (moved to _inverted-headerbar) ;
  - windowbuttons size and orange close button.
- Symlink these files and folder to upstream ones:
  - **_colors-public**
  -  **_common**
  -  **_drawing**
  -  **_widgets**
  -  `/widgets`

Now all Yaru tweaks **must** be located inside **_tweaks** or **_inverted-headerbar**.


Feichtmeier:
I generally think this is a very good idea, since we have become fans of the upstream theme, and only change it to have our colors.
However, let @clobrano double check this, since I remember him saying symlinks can cause problems with meson or something :) I have forgotten what he said

Feichtmeier (approved):

clobrano:
> I remember him saying symlinks can cause problems with meson or something :) I have forgotten what he said

Just that currently meson copy the entire file when it finds a symlink, while **in the future** it will copy the symlink. So for now, all our symlinking magic works fine, but in the future we might need to rework it

(I'm actually writing this from fedora 34, so I think we can agree we've become upstream fans XD)

Jupi007:
> Just that currently meson copy the entire file when it finds a symlink, while **in the future** it will copy the symlink. So for now, all our symlinking magic works fine, but in the future we might need to rework it
> 
> (I'm actually writing this from fedora 34, so I think we can agree we've become upstream fans XD)

Okay, thanks @clobrano 
Let's merge this so.

PR #2907, opened 2021-06-13
Missing snapping icons for inkscape

### LINKED ISSUE #2895: snap-symbolic is causing problems...
In Inkscape, it's replaced at least three of the icons associated with snapping:

image

Does anyone know what the expected use is for this symbol...?

ubuntujaggers:
Closes #2895 

image

And while I'm doing missing icons, I added one for the selectors dialog:

image


Jupi007:
Very nice 👍

Jupi007:
Just a note: I see that there is a `snap` icon.
But our symbolic snap icon doesn't fit at all, maybe should we rename it to `snapcraft-symbolic` to avoid any problem?

Muqtxdir:
> maybe should we rename it to snapcraft-symbolic to avoid any problem?

yes, that would avoid the conflicts just like this one.

PR #2910, opened 2021-06-13
Drop GDM purple background

### LINKED ISSUE #2908: GDM background should follow overview
The upstream GDM theme uses the same background as the shell theme. This isn't the case on Yaru. There's a somewhat jarring transition from the striking purple to the subdued gray. I also think that the overview gray color would look nicer with the bubbles and text field.

Screenshot from 2021-06-13 10-05-23

Screenshot from 2021-06-13 10-05-41

Jupi007:
Revert GDM background to upstream (`$system_bg_color` - same as overview) to avoid jarring transition **purple** -> **grey**.

This is just for testing and get opinions, because I can't take screenshot.

Closes #2908

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_screen-shield.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_screen-shield.scss
@@ -66,7 +66,7 @@
 }
 
 #lockDialogGroup {
-  background-color: darken($purple, 10%); // Yaru: purple login screen
+  background-color: $system_bg_color;
 }
 
 #unlockDialogNotifications {


LinuxLlama:
Screenshots~
Screenshot from 2021-06-13 12-13-29
Screenshot from 2021-06-13 12-13-41

Jupi007:
Thanks @LinuxLlama 

IMO this look very much professional than the purple, especially with dark entries.
What do you think of @ubuntu/yaru-team ?

Feichtmeier:
I think we need to rather ask the canonical branding team for this one

Jupi007:
> I think we need to rather ask the canonical branding team for this one

I let you doing this, I have no idea who to ask :)

madsrh:
This would probably be elioqoshi, right @Feichtmeier?

I agree that the purple can be jarring, but a solid gray a bit too dull IMO. Can we do some fancy CSS effects? Perhaps just a gradient or something 🤷‍♂️

Feichtmeier:
I would keep the purple, personally
No, branding department is @waynecrosby

Jupi007:
The problem is that the transition between GDM and GS overview is very hard. There is no animation: it is a purple frame then a grey frame.

Jupi007:
We could also make the overview background purple, but I'm afraid that looks weird 🤔

Feichtmeier:
Doesn't the hard cut come from the fact that gnome40 session defaults to the overview? As far as I understood ubuntu desktop team dash to dock will happen again, so this hard cut will prbly not happen - maybe @3v1n0 can answer on this one

LinuxLlama:
It's possible to jump straight to the desktop on boot via extensions. Since Ubuntu uses a dock on the desktop, there's not much of a need for the overview to appear on login. However, it's still a thing to consider for those on different GNOME configurations

Jupi007:
> Doesn't the hard cut come from the fact that gnome40 session defaults to the overview?

This is indeed the source of the problem. If ubuntu-desktop comes again with dash-to-dock for GS40 it will be solved... but only for Ubuntu, any other Yaru user who want to have overview at startup will encounter this issue.

I totally understand the branding problem, this purple background looks very Ubuntu. But it is just like the plymouth theme which was purple too before. To get a flicker free boot loading, it was needed to make it black.

But let's wait Ubuntu branding guys answer :)

godlyranchdressing:
The black looks incredible, though the loss of purple would indeed make it less "Ubuntu", especially with the dropping of the dark headerbars but I suppose it's just something that we'll have to live with.

> I agree that the purple can be jarring, but a solid gray is a bit too dull IMO. Can we do some fancy CSS effects? Perhaps just a gradient or something man_shrugging

@madsrh we tried a gradient trick before, but there was some issues with how it was rendering and there was this very noticeable banding effect going on. The Shell's native CSS gradient support doesn't offer much control too, so solid colours or a texture (like old GNOME's noise) are pretty much our only options here.

Feichtmeier:
Even if the reason is perfect we need an ack from the ubuntu-desktop and/or their branding team for this one. This would remove a big branding part of the Ubuntu desktop.
Also the work on dash-to-dock for gnome40 is almost complete as far as I can tell https://github.com/micheleg/dash-to-dock/pull/1402#pullrequestreview-684623321

madsrh:
I remember the issue with the gradient @godlyranchdressing but perhaps it would be possible to create something that could work with solid colors?

The Ubuntu [wallpapers ](https://149366088.v2.pressablecdn.com/wp-content/uploads/2015/03/suru-desktop-wallpaper-ubuntu-vivid-750x422.jpg)have a pattern that would be great to reuse, but I've tried to recreate them without gradient and it looks like 💩 

I don't have a mockup, but check these ideas out for inspiration:
https://www.svgbackgrounds.com/#rainbow-vortex

I'm sure that Wayne could come up with something within this limitation 🤷‍♂️

Feichtmeier:
@vanvugt and/or @3v1n0 could you comment on this one? Do we need this change?

vanvugt:
> @vanvugt and/or @3v1n0 could you comment on this one? Do we need this change?

The answer depends on the Yaru Ubuntu theme for GNOME 40. And while I do a lot of GNOME 40 work I do not know what the overview colour will be for Ubuntu / Yaru. Whatever it is, the GDM background should ideally be the same.

elioqoshi:
Gray should probably be a safe bet but I'd let this eventually wait a bit longer once we have some more branding directions for the new Ubuntu branding?

In either case, shouldn't the background be light when using the light theme and use this kind of gray when using the dark theme?

Feichtmeier:
> Gray should probably be a safe bet but I'd let this eventually wait a bit longer once we have some more branding directions for the new Ubuntu branding?
> 
> In either case, shouldn't the background be light when using the light theme and use this kind of gray when using the dark theme?

Changing the gnome shell theme to yaru-light (which is installed by default) can not be done without installing the "user-themes" extension in gnome shell, which is as far as I know "not supported" except Ubuntu decides to support it :)

Muqtxdir:
Hi @elioqoshi 
> In either case, shouldn't the background be light when using the light theme and use this kind of gray when using the dark theme?

The **GDM** picks the theme-aspect from `/usr/share/gnome-shell/gnome-shell-theme.gresource` which consists themes for both ***Light*** and ***Dark*** variants of Yaru but GNOME does not provide any `gsetting` for switching GNOME-Shell themes unlike GTK which can smoothly switch between ***Light*** and ***Dark*** variants from `/usr/share/themes/Yaru/gtk-3.20/gtk.gresource`

To expand on what @Feichtmeier mentioned, [User-themes](https://extensions.gnome.org/extension/19/user-themes/) is one of the popular GNOME-Shell extensions that allows the option to switch GNOME-Shell theme via it's own `gsetting`

elioqoshi:
@Feichtmeier @Muqtxdir Thanks, but I didn't ask about the technical reasoning behind that rather the design rationale behind it. I mean we could agree that (technical limitation aside) login should be light in the light theme and vice-versa, right?

Muqtxdir:
>  I mean we could agree that (technical limitation aside) login should be light in the light theme and vice-versa, right?

Yes, I agree with you :+1:

didrocks:
Please consider multi user environment which have different themes (one is using dark, another one is using light), GDM is supposed to be neutral in that case.

Also, due to encryption and such, we don’t have necessarily access in GDM of the selected user option.

Jupi007:
Indeed, the GDM theme variant cannot be synced with the user selected GS theme. And as said @didrocks as multiple users can select different variant, the GDM theme should be neutral.

I would also like to point out that both light/dark GS theme variants use the same soft grey overview background, so also using it for GDM could fit both (and just a personal opinion, it looks better with dark entries).

elioqoshi:
That makes a lot of sense, thanks for the context. I agree with @Jupi007's view on this as well.

Feichtmeier:
Nice, if everyone agrees on the grey (especially the design) I'm fine as well. Let's wait until the branding colour has been decided and keep this pr opened until then and eventually merge it then.
Thanks everyone

Feichtmeier:
Since the branding discussion has been postponed for this cycle at canonical design, what shall we do with this one here now?
Keep it purple or switch to dark gray? @didrocks @kenvandine :smile_cat:

vanvugt:
The "jarring transition purple -> grey" doesn't happen anymore since we reverted to the old login animation when Ubuntu Dock is enabled. This means impish is now purple -> purple. The wallpaper then covers the whole screen and only after the overview is manually opened do you see grey for the first time. So there's no strict need for this anymore unless you have the Ubuntu Dock extension disabled in which case the jarring transition will occur again.

didrocks:
And as not having the ubuntu dock extension is not our default experience, let’s thus keep it purple.

Feichtmeier:
Alright closing it then as the consensus seems to keep the purple for various reasons - ack from jupi :)

Issue #2913, opened 2021-06-15
Proposal: remove the mixed theme from Yaru


Feichtmeier:
Gtk3 and Gtk4 both do **_not_** support having different background/text colors for headerbar than in the rest of the window. This means that to have bug free applications, the headerbar needs to have the same background and text "luminescence" as the rest of the window. The technical reason for this is, that the only "api" available for theming in gtk, public-color.scss, does not support headerbar colors.

Examples:
- https://github.com/ubuntu/yaru/issues/2358
![grafik](https://user-images.githubusercontent.com/15329494/93706763-04720180-fb2a-11ea-9c3e-7720d484f2ee.png)
![grafik](https://user-images.githubusercontent.com/15329494/93706777-181d6800-fb2a-11ea-9ecf-d17492384212.png)
![grafik](https://user-images.githubusercontent.com/15329494/93706779-210e3980-fb2a-11ea-8dc5-dc6335d30923.png)

- https://github.com/ubuntu/yaru/issues/2912
![grafik](https://user-images.githubusercontent.com/15329494/122020712-4ed89400-cdc5-11eb-8aac-4b3099071900.png)

The only solution to fix these problems is to "only" have yaru-light and yaru-dark themes - which is the defacto standard on other operating systems.

Personal note: I think theming should only be made with the tools available - if the platform does not support some design, it should not be forced.

Relevant discussion: https://github.com/ubuntu/yaru/discussions/2677 point 6

madsrh:
Here's another example from Firefox:

![image](https://user-images.githubusercontent.com/3986894/122463478-f0263c80-cfb5-11eb-8a9b-09071610a700.png)

Initu-Castilhos:
In case Yaru-light will be renamed to Yaru?

Feichtmeier:
Yes, we will probably stick to "yaru" and "yaru-dark" then ⚪ / ⚫

Benni3-1415:
I like the mixed theme very much.

#### One idea to keep it alive:
![idea](https://user-images.githubusercontent.com/73243667/122636892-eef72b80-d0eb-11eb-96d7-b61408c3a28f.png)

The lighter shade doesn't look as good as before, but the widgets could stay in the light theme!

elioqoshi:
[...]

PR #2922, opened 2021-06-19
Remove "yaru-light" and keep "yaru" as the fully light theme

Feichtmeier:
As discussed in https://github.com/ubuntu/yaru/issues/2913 and https://github.com/ubuntu/yaru/discussions/2677 the dark headerbar in a light theme is not supported in Gtk3 and Gtk4 and we often can not guarentee to not introduce bugs in gtk applications. This removes the theme "yaru-light" and keeps "yaru" from now on as the fully light gtk theme. We need to make sure that the gnome control center setting page is updated correctly.

CC @robert-ancell @elioqoshi

If this will be merged
Closes https://github.com/ubuntu/yaru/issues/2913


Muqtxdir:
Hi @Feichtmeier 👋

Thanks for doing this, giving a nice head-start for questions/discussions/stuff like this: 
- is there a reason why `light` option  is still present in `meson_options.txt`?
- would it makes sense to rename `Yaru-light` and `Yaru` in gnome-shell meson-files to revert back to `Yaru` and `Yaru-dark` but keeping the default to `Yaru-dark` ( _just suggesting a name change to keep it consistent with `gtk-themes`_) ?
- `gtksourceview` generated for `Yaru` still show up as `Yaru-light`
- under `/gtk/src/default/gtk-3.20/apps/_unity7.scss`, the `titlebar` still defaults to `$headerbar_bg_color`and `$headerbar_fg_color`. shouldn't this be changed to match `gtk-themes` ?

*unrelated to this PR:*
- probably need to remove `default` in metacity and rename `light` as `default`
- separate `unity-assets` from `gtk/src/default/unity` and keep it separate

Feichtmeier:
I don't think that we should rename the gnome-shell themes, unless GNOME finally figures out how to sync gtk and shell themes
grafik

The default theme is present as "default" (:de: "vorgabe" :) ) and as an alternative you can see "yaru-light" - this clearly indicates what you get if you want to manually switch, which is not possible without "modifying" ubuntu with the user-themes gnome shell extension. Let's have @clobrano have on opinon here, I personally do not have a preference :)

I kept gtksourceview because I don't what if @flexiondotorg intention was with the theme names. If martin can't answer currently, I don't feel like this is a blocker for this PR and we can fix it in a separate PR if needed.

I'll fix the unity theme, thanks! :+1:

clobrano:
If I remember correctly, the two entries "default" and "yaru-light" are a consequence of a change we made to make yaru available also in vanilla session, but I should double check and I'm AFK till tomorrow.

Feichtmeier:
I think we can leave the unity titlebars dark, as they are true titlebars and there can not be any widgets inside except the menu and the window controls. I don't see any problem there

Jupi007:
@Feichtmeier Because needed for the inverted headerbar, I had split the gtk4 tweaks file. As you remove ambiance here, this is no more needed. You can merge all these files into a single just like before.

elioqoshi:
I checked this with the brand team and we basically agree that this is the smartest choice to go with. We also hope to have a few more checkin meetings to ensure the style and the current Ubuntu rebranding are synced well :)

Feichtmeier:
> I checked this with the brand team and we basically agree that this is the smartest choice to go with. We also hope to have a few more checkin meetings to ensure the style and the current Ubuntu rebranding are synced well :)

Thanks for the acknowledgement!

I wait until @clobrano double checks the meson changes and merge after

clobrano:
Okay, I was a bit confused by the reference to the gnome-shell switch, but I saw now that the shell isn't involved here, so green light for me too

clobrano (approved):

Varun-garg:
This has broken my experience on arch. hybrid theme is removed, and in dark theme some apps are now fully light and some are now fully dark. firefox

PR #2943, opened 2021-07-02
Add desktop-logoff sound

ItzSwirlz:
When the sounds were added, the logoff one wasn't. @MadsRH says
this was because it'd be better to just log off immediately, and not
wait for a sound to play.

Session managers can handle this stuff, and play the sound
while logging off in the background. But if it doesn't work, the
sound can always be disabled.


madsrh:
Nice @ItzSwirlz 👍
Did you edit the files I send you? I think there's a bit of silence at the beginning that should be removed.

image

ItzSwirlz:
> Did you edit the files I send you? I think there's a bit of silence at the beginning that should be removed.

Only converted it to .oga via ffmpeg. I'll cut it down, + I'll link the original .oga and .wav file.
[logout_3.zip](https://github.com/ubuntu/yaru/files/6757458/logout_3.zip)

madsrh:
Hey @ItzSwirlz 
I think the logout sound is disabled in Ubuntu on Gnome, but perhaps it's different on Cinnamon Remix. Have you tested this?

ItzSwirlz:
Logout sounds are a feature in Cinnamon, and the packager can use gsettings overrides to adjust the sounds (see the ubuntucinnamon-environment package)

ItzSwirlz:
Well you were right. It's not
Too long but could be useful to have a fix: https://github.com/linuxmint/cinnamon-session/pull/146

Issue #2921, opened 2021-06-19
yaru cursors: kde/plasma support


jbouter:
**Expected Behavior**

For all cursors to properly display, in a KDE plasma behavior

**Actual Behavior**

Key components and symbolic links are missing that KDE plasma expects

**Steps to Reproduce the Problem**

1. Use KDE Plasma
2. Install the yaru cursor theme
3. Enable the cursor theme
4. Observe missing icons, falling back to the breeze cursors

**Yaru version**

```
20.04.11.1
```

**Missing Icons**

Thus far, I've been able to identify the following missing symbolic links:

missing file | should link to
------------|----------
pointing_hand | pointer
whats_this | help
half-busy | progress
ibeam | text
split_v | row-resize
split_h | col-resize
size-ver | default
size-hor | default
size-fdiag | default
forbidden | no-drop
closedhand | dnd-move
wait | watch

The following files are present in the breeze cursor directory, but not in Yaru, and I have not been able to identify which files they should point to yet:

```
center_ptr
color-picker
down-arrow
draft
grab
left-arrow
openhand
right-arrow
up-arrow
wayland-cursor
x-cursor
```

I hope it's possible for these symbolic links to be created, in order to be able to fully use the Yaru cursor theme within KDE Plasma.

Muqtxdir:
Hi @jbouter 👋
Thanks for sharing the files that needed to be linked. I'll do this

jbouter:
Hi @Muqtxdir 👋🏻 

Thanks! If there's any assisting I can do, please let me know. :-)

Muqtxdir:
Hi @jbouter 👋

I've created this PR: #2958 for all the **missing-symlinks** (_still need to do wayland-cursor and x-cursor_), please tryout this PR when free and share your experience

jbouter:
Hi @Muqtxdir ! Brilliant, I'll give it a go as soon as I can. I am however missing `grab` and `openhand` from the list. They seem to be related to each other.

How can I go about building your PR? I haven't really worked with cursors before, and only ever downloaded "compiled" cursors. Can I download the content of your PR and move it to the proper directory, or do I need to build cursors prior to doing so?

Muqtxdir:
> How can I go about building your PR? I haven't really worked with cursors before, and only ever downloaded "compiled" cursors. Can I download the content of your PR and move it to the proper directory, or do I need to build cursors prior to doing so?

[...]

PR #2944, opened 2021-07-04
Fix cursor inconsistencies

Benni3-1415:
source-cursors

### Fixes:
- Set all _additions_ line-join to miter
- Adjust some paths (max.⅓ px) to be pixel perfect
- Make the default cursor symmetrical
- Add missing gradients and align all globally
- Optimize hotspot positions
- Add `color-picker` and `wayland_cursor` for kde/plasma support
- change `source-cursor.svg` title to _Yaru Cursors_


Muqtxdir:
Hi @Benni3-1415,

Please render the changes you included in `source-cursors.svg` so that the differences can be compared

Benni3-1415:
I've included the current state in the description. ^

Is it ok if i don't build the cursors until the end after i've made the other two changes?

Should i remove all gradients or add the missing?

Jupi007:
Where are there missing gradients (in which cursor I mean) ?

Benni3-1415:
> Where are there missing gradients (in which cursor I mean) ?

- window edges
- arrows
- text cursors

Jupi007:
Okay thanks :)

IMO you should add missing ones.

Benni3-1415:
The hand of the dnd cursor is 2px higher than the hand of the hand cursor.

Should the gradient position depend on the shape (move 2px up) or on the grid (stay at the same position)?

Jupi007:
I'd like to say the second option (stay at the gradient at the same position), but I'm really not the cursors expert here.
@ubuntujaggers What do you think of?

ubuntujaggers:
I was going to say the opposite (move the gradient with the fist) but let's try whatever's easiest first!

Agree on missing gradients, it's a good-looking effect so I'm not in favour of removing the gradient 👍

Benni3-1415:
### All gradients aligned globally
source-cursors

Benni3-1415:
I've noticed that some hotspots are on pixels and some are in between:
hotspot

Is it possible to place the hotspot between pixels?
1. Yes, it is possible that the cursor is positioned between pixels on the screen.
2. Only on HDPI, at 96 dpi the cursor position is rounded.
3. No, the `render-cursors.py` script rounds the hotspot position for all screens.

Which theory is correct?

Benni3-1415:
The default and the zoom cursors are the only ones with rounded line ends.
Corrected:
left_ptr zoom-in
It would be more consistent, but the default cursor would also lose a unique feature.

I tend to flatten the end.

ubuntujaggers:
I don't actually know which scenario is correct, but the hotspots all seem to be working...?

Muqtxdir:
Hi @Benni3-1415,

In addition to fixing cursor-inconsistencies, would you also look into doing the following cursors to source-cursors.svg? (related issue: #2921)

- [ ]  color-picker
- [ ] wayland-cursor

Benni3-1415:
Hello @Muqtxdir,
here are some _color-picker_ concept designs:
| 1 | 2 | 2 empty |
|:---:|:---:|:---:|
| color-picker-1 <br> color-picker-1-large | color-picker-2 <br> color-picker-2-large | color-picker-3 <br> color-picker-3-large |
| Doesn't feel like Yaru. | :star: My favorite |  |

---

#### Some questions:

What is the Wayland cursor needed for in KDE?

Why is the X.org cursor not needed when the Wayland cursor is?

It's a logo, can we yarufy it?

Muqtxdir:
Thanks @Benni3-1415, awesome work! 

> Which one is the best?

I prefer the **2-option** over the others as it's less detailed and simpler

> What is the Wayland cursor needed for in KDE?

I too don't know where it's used, but the reason for suggesting addition of wayland-cursor was to make the cursor theme also compatible with KDE usage.

>   Why is the X.org cursor not needed when the Wayland cursor is?

The X.org cursor is already present see [X_cursor](https://github.com/ubuntu/yaru/blob/master/icons/Suru/cursors/X_cursor)

> It's a logo, can we yarufy it?

Since X.org cursor is also a yarufied, I think so neutral-yes from me. let's hear from others on this @ubuntu/yaru-team

Jupi007:
The color picker symbolic icon looks like this in Yaru:

Capture d’écran du 2021-07-28 18-18-22

Benni3-1415:
@Jupi007,
IMO, it doesn't fit to the other cursors.
There are no other rounded line ends in the cursor theme.

You are right, it's confusing to have a different shape and the symbolic looks a bit better …

Jupi007:
IMO your **option-2** looks good.
Just maybe make the head of pipette looks a bit more like the symbolic icon 🤔

Benni3-1415:
> Just maybe make the head of pipette looks a bit more like the symbolic icon :thinking:

### Like so?
color-picker-2-2
@Jupi007, you are right, it looks way better.

But I was just about to [remove all rounded line ends](https://github.com/ubuntu/yaru/pull/2944#issuecomment-882068879).

What should I do in both problems?

Jupi007:
> ### Like so?
> 
> color-picker-2-2
> @Jupi007, you are right, it looks way better.

This is perfect 👍

> But I was just about to [remove all rounded line ends](https://github.com/ubuntu/yaru/pull/2944#issuecomment-882068879).
> 
> What should I do in both problems?

Personally I prefer the current look of the cursors with rounded line end.
But it could be better to ask everyone CC @ubuntu/yaru-team (and especially @ubuntujaggers).

Benni3-1415:
@ubuntu/yaru-team,
what should I do now ([X.org cursor](https://github.com/ubuntu/yaru/pull/2944#issuecomment-888441287), [pipette](https://github.com/ubuntu/yaru/pull/2944#issuecomment-888493454))?

ubuntujaggers:
The pipette looks great to me and we should probably make the symbolic use the same design.

Not sure what the x-org/wayland cursors are - should we merge this as-is and do other PRs for any other additions?

Benni3-1415:
@ubuntujaggers,
### Symbolic `color-select` in the same design
color-select-high color-select

This cursor (width: 4px) uses a rounded line end. Are the default and the zoom cursors (width: 2px) the same case, or should they get a [flat end](https://github.com/ubuntu/yaru/pull/2944#issuecomment-882068879).

ubuntujaggers:
I would personally leave the existing rounded ends as they are.  The other cursors that have thin lines are stuff like cross hairs and right angles, which are meant to be sharp and precise.  In the case of the arrow and the magnifying glass, the lines represent objects that are cylinders in real life (the shaft and the handle) so I don't think they need to be consistent with the squared ends unless it looks better.

There's another line of the same thickness that has a rounded end and that's the pointing finger.  In all three cases - arrow shaft, magnifier handle, pointing finger - a rounded end makes sense to me.

Benni3-1415:
Please excuse the late response …
- Should I [change the symbolic `color-select`](https://github.com/ubuntu/yaru/pull/2944#issuecomment-896227128)?
- I would revert the cursor mimetype to match the default cursor again.
- One idea for the X.org and the Wayland cursor (Ubuntu bold):
  xw-cursor

Muqtxdir:
> Should I change the symbolic color-select?

> I would revert the cursor mimetype to match the default cursor again.

yes, but in separate PR 

> One idea for the X.org and the Wayland cursor (Ubuntu bold)

looks good and consistent with the cursor-set 👍

Muqtxdir:
@Benni3-1415 you need to add the new cursor file-names also into [`w32-make.sh`](https://github.com/ubuntu/yaru/blob/master/icons/src/cursors/w32-make.sh) and [`x11-make.sh`](https://github.com/ubuntu/yaru/blob/master/icons/src/cursors/x11-make.sh) as well

Benni3-1415:
> @Benni3-1415 you need to add the new cursor file-names also into [`w32-make.sh`](https://github.com/ubuntu/yaru/blob/master/icons/src/cursors/w32-make.sh) and [`x11-make.sh`](https://github.com/ubuntu/yaru/blob/master/icons/src/cursors/x11-make.sh) as well

Is the `X_cursor` needed in Wayland and vice versa?

Muqtxdir:
> is the `X_cursor` needed in Wayland and vice versa? 

@Benni3-1415 yes, both x_cursor and wayland_cursor should be present whether it's in a Wayland session or xorg session for fallback

Muqtxdir:
all right merging!

PR #2945, opened 2021-07-04
modify gnome-shell-theme.gresource

Muqtxdir:
Hi,

This should be okayed to merge only if the GNOME Shell is patched as discussed [here](https://github.com/ubuntu/yaru/discussions/2677#discussioncomment-963236)

patches that need to be applied on GNOME Shell for fully functional global dark-theme:
- [ ] [main.js](https://git.launchpad.net/ubuntu/+source/gnome-shell/tree/js/ui/main.js?h=applied/ubuntu/impish)
- [ ] [ubuntu.json](https://git.launchpad.net/ubuntu/+source/gnome-shell/tree/debian/ubuntu-session-mods/ubuntu.json?h=applied/ubuntu/impish)

### TO-DO:
- [x]  Make **Yaru** as default shell-theme
- [x]  Add separate dark and light directories
- [x]  Remove alias for gnome-shell.css in gnome-shell-theme.gresource.xml
- [x]  Update sessions to reflect the changes
- [x]  Enable `gnome-shell-gresource` option as default
<s> Symlink ~inverted-themes~ mix-themes option.</s>
- [x] Reflect changes for `debian/`


Muqtxdir:
with the last commit I enabled `gnome-shell-gresource` by default, @Jupi007 and @LinuxLlama could you tryout this branch and share whether GNOME Tweaks lists both **Yaru** and **Yaru-dark** under Shell theme.


*Note: Please remove any-trace of **Yaru** before trying out this branch to avoid any conflicts*

Muqtxdir:
> It does! 

That was quick @LinuxLlama ! and thanks  
> The default shell theme is also light now

Great, but do note this is not yet finalized and is one of the attempts to counter this [issue](https://github.com/ubuntu/yaru/discussions/2677#discussioncomment-963236)

gnome-shell/src/meson.build
```diff
@@ -96,6 +106,10 @@ foreach variant: variants
   if is_variant
     subdir(variant)
   endif
+
```
clobrano, line 109: I suspect this two `if` blocks (between 106 and 112) are not necessary anymore. `is_variant` is true if `variant != light` right? So:

- if `is_variant` is true, then `variant == 'light'` and the subdir command is equal to `subdir('light')`
- if `is_variant` is false, then `variant != 'light'`, but subdir command is still equal to `subdir('light)`
 
 I believe this was necessary when the variantes were three (default, light and dark).

gtk/src/inverted/install-inverted.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
clobrano, line 1: Could we add a description of what "inverted" refers to? It's bit cryptic in the code

clobrano (changes requested):

gtk/src/inverted/install-inverted.sh
Muqtxdir: Should I add this in description:

when `inverted` is true it will create symlinks like this:
- **Yaru-inverted** symlinked to **Yaru-dark** (*Yaru inverted is Dark-gtk and Light-shell*)
-  **Yaru-inverted-dark** symlinked to **Yaru** (*Yaru-inverted-dark is Light-gtk and Dark-shell*)

This two symlinks will be used if, the switch for ***keep gtk and shell in sync*** is disabled.

clobrano: That's fine for me. The main point was to describe the combination of shell and gtk used.

Muqtxdir:
> Tried to resolve merge conflict? Nice broken XML rofl

@vitalkanev I resolved it with the successive commit. see: [`2d3febe`](https://github.com/ubuntu/yaru/pull/2945/commits/2d3febe7bf75d5cc0f06b6ab776fccb1f0e0dd77)

Muqtxdir:
relevant LP-bug issue: https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/1943857

3v1n0:
I'm closing this, as the idea is already part of #3416.

PR #2948, opened 2021-07-11
Fix dock and panel look

Jupi007:
- Make panel background static into the overview ;
- Use a 2% darken `jet` as panel background ;
- Define dock background as transparent (will render half dark transparent, we need to understand why).

VirtualBox_Ubuntu 21 10_11_07_2021_17_15_20
VirtualBox_Ubuntu 21 10_11_07_2021_17_15_49


Feichtmeier (approved):

Muqtxdir:
Thanks @Jupi007 this is better!

Jupi007:
- Make panel background static into the overview ;
- Use a 2% darken `jet` as panel background ;
- Define dock background as transparent (will render half dark transparent, we need to understand why).

PR #2950, opened 2021-07-12
Revert to lighten dash background

Jupi007:
Caused by #2948

I didn't noticed that the dash (not the dock) background was based on top panel background var. Into #2948, I've darkened the panel bg. As many components (search entry, search results, app folder, ...) are based on dash bg, it result a too much darker look everywhere.

So this PR revert `$dash_background_color` to old `$panel_bg_color` value: `lighten($jet, 2%)`.


Feichtmeier:
AH I thought this was on purpose ^^ lgtm

Feichtmeier (approved):

PR #2955, opened 2021-07-20
Improve selected row look

### LINKED ISSUE #2954: Global row selection is problematic
The row selection for nautilus and apps with similar layout works.

However it does not work well globally because those rows can be everywhere
grafik
(portal)

grafik
(gnome software)

@Jupi007 do you have any idea how we can limit this to nautilus and gnome control center?

Jupi007:
Improve selected row look:

- Remove the little orange left stripe (except on Nautilus sidebar) ;
- Increase bg contrast.

Capture d’écran du 2021-07-20 19-42-49

Capture d’écran du 2021-07-20 19-43-18

Closes #2954

### DIFF
--- a/gtk/src/default/gtk-3.20/_common.scss
+++ b/gtk/src/default/gtk-3.20/_common.scss
@@ -1969,19 +1969,19 @@ treeview.view {
       @extend %selected_items;
     }
 
-    &:backdrop, & {
+    /*&:backdrop, & {  // Yaru change: due to grey row selection background we no more need to modify border-color
       border-left-color: mix($selected_fg_color, $selected_bg_color, 50%);
       border-top-color: transparentize($fg_color, 0.9); // doesn't work unfortunatelly
-    }
+    }*/
   }
 
   &:disabled {
     color: $insensitive_fg_color;
 
-    &:selected {
+    /*&:selected {
       color: mix($selected_fg_color, $selected_bg_color, 40%);
       &:backdrop { color: mix($backdrop_selected_fg_color, $selected_bg_color, 30%); }
-    }
+    }*/
 
     &:backdrop { color: $backdrop_insensitive_color; }
   }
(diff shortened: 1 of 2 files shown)


Feichtmeier:
Looks good!
How does it look in GCC and the portal app now?

Feichtmeier:
I like it, but let the others check 👍

clobrano:
I understand the problem for #2954, but I think all the selections must be the same, nautilus included

Jupi007:
@clobrano So maybe drop the orange stripe, and make all selections using this soft grey bg (even listviews)?

madsrh:
>  So maybe drop the orange stripe, and make all selections using this soft grey bg (even listviews)?

IF we do that, could we make both have rounded corners? Maybe with slightly more padding to the sides

Jupi007:
> IF we do that, could we make both have rounded corners? Maybe with slightly more padding to the sides

I'd like to say yes because it could look better. But less diff with upstream = less problems.
The rounded corners selected list row is a different widget, which have bigger row + margin + rounded corners.
The only place where we could be able to force border-radius is on nautilus sidebar because it has less scope, and that couldn't brake anything else.

Jupi007:
Rounded corners on nautilus selected row:

Capture d’écran du 2021-07-21 12-17-08

clobrano:
> @clobrano So maybe drop the orange stripe, and make all selections using this soft grey bg (even listviews)?

unfortunately yes :(

Jupi007:
No problem, I'll do it :)

Feichtmeier:
If we remove the stripe, I would keep it simple for now and play safe with the regular gray row, without a border-radius.

Jupi007:
Grey selected row everywhere:

Capture d’écran du 2021-07-26 12-20-37
Capture d’écran du 2021-07-26 12-22-39

Feichtmeier (approved):
Looking good ☺️

PR #2959, opened 2021-07-21
Proposed tweak to session-properties app icon

ubuntujaggers:
I've always thought this pictogram (inherited from Suru) was a bit odd. Tonight I was looking at the corresponding icon in Adwaita and Humanity (it's used for Startup Apps) and I realised that the oblongs are meant to represent app windows.  So, I'm proposing the following change.

**Current**

image

image

**Proposed**

image

image


Jupi007:
Really much better 👍

PR #2963, opened 2021-07-22
Suggestion for new characters app icon

ubuntujaggers:
Just an idea...  I always feel like the characters app icon is a bit out-of-place.  The smiley is very nice but looks a bit toy-ish, so I wondered if it would be more "Yaru" to go with a simpler design:

**Current**

image

image

**Proposed**

image

image

I'm not 100% sure I prefer the new one so honest opinions welcome!


ubuntujaggers:
I darkened the grey "O":

image

image

Muqtxdir:
Personally looks good even on small sizes like dock icon.

3v1n0:
Mh, this is now merged, but I feel one nice thing of the previous version was that it made clear that the app can be used to search emojis (as it also shows up searching `emoji` in the overview).

Maybe we can use another emoji-char to make this clearer again?

PR #2965, opened 2021-07-25
Cabinet drawer-based file manager icon

ubuntujaggers:
Hi everyone,

Here's a PR so you can try it:

image


Feichtmeier:
Nice idea. WDYT of changing the colors a bit away from the folder colors so there can't be a "this looks like a folder" argument :) ?
Not blue. But maybe a little bit more metallic gray and no gradients on the folder popping up in the back?

ubuntujaggers:
image

image

Speaking personally I think I prefer the folder colours :thinking:

Feichtmeier:
Another one of my strange ideas. What if you take the original idea of you with the gradient but don't cut out the right third of the folder baseplate? So the role thing looks more like a cabinet without using too many rows of drawers? 🥺 Sorry just brain 🧠 stormy

ubuntujaggers:
So the baseplate is like a normal square baseplate?

Feichtmeier:
Yes. :D

Jupi007:
@ubuntujaggers I really like the new icon from this PR 👍

Feichtmeier:
@ubuntujaggers Oh no I entered Inkscape again :see_no_evil: 
What I meant could be visually translated to "this" :) :
image

ubuntujaggers:
Yep, I got it 🙂 I've got a local version of that and I'll push it for people to try asap, got a busy week but back on it Monday 👍👍

madsrh:
After trying this PR it feels like the icon is kind of dark and - on this screen - falls too much in with the dock background.

> Another one of my strange ideas. What if you take the original idea of you with the gradient but don't cut out the right third of the folder baseplate? So the whole thing looks more like a cabinet without using too many rows of drawers?

What I liked about the initial design was that it's both a folder AND a file cabinet icon, so IMHO changing it too much belittles the idea. Your idea might still have that _folder icon_ association which is the most important thing IMO.

ubuntujaggers:
Current:

image

image

jgabriel98:
> What I liked about the initial design was that it's both a folder AND a file cabinet icon, so IMHO changing it too much belittles the idea. Your idea might still have that folder icon association which is the most important thing IMO.

Agree,
i think the initial design looks better (personal opinion).

Also it still feels like somethins is slightly missing something? maybe the gradient dark colour (the "start" color from the left side) should be less darker.

but still, great design! better then the default one, congrats!.

ubuntujaggers:
How about original design with brighter drawer:

image

image

madsrh:
> How about original design with brighter drawer

👍 I'm very happy with this one!

ubuntujaggers:
Will do this morning - just want to quickly try something first :+1:

ubuntujaggers:
Okay, what I wanted to try very quickly sucked, so I'm pushing as-is :)

ubuntujaggers:
@Feichtmeier @clobrano what do you think?

ubuntujaggers:
I've got to say, this screenshot of the About window looks a lot sweeter than the recent screenshot on OMG:

image

Feichtmeier:
Indeed looks Def. Much better! Thanks for trying

clobrano:
> @Feichtmeier @clobrano what do you think?

This is quite an improvement 🤟

ubuntujaggers:
Magic, I'll merge so we can try it for a while and see if we want to keep it 👍

3v1n0:
FYI, I've reverted this nice change in the impish branch as we're past UI freeze and this may affect the screenshots, so I'd prefer to land this early when 22.04 gates open.

PR #2966, opened 2021-07-25
Suggested new icon for aisleriot

ubuntujaggers:
Solitaire aka gnome-aisleriot is one of our oldest designs, and IMO it's not the smartest icon in the set.  Just an idea - does anyone think this is an improvement?  It's not the most crucial app but it's one of the defaults, so worth making it as nice as possible :)

**Current**

image

**Proposed**

image

image


ubuntujaggers:
Now has characters instead of dots on the two largest sizes:

image

image

madsrh:
@ubuntujaggers what do you think about making the separation between the cards a bit stronger?

cards

ubuntujaggers:
Good idea 👍

ubuntujaggers:
Current:

image

madsrh:
I think we can merge this one 👍

PR #2993, opened 2021-08-27
Fix code block formatting

computercam:
Fix code block formatting so markdown parser shows each section properly


Jupi007:
Thanks @computercam :)

PR #2998, opened 2021-09-02
Fix missing red button into some apps

Jupi007:
@Muqtxdir Can you confirm it fix the problem of red button, please?


Muqtxdir:
@Jupi007 I'll try this branch when I reach home 🏠

Muqtxdir:
It works!

Screenshot from 2021-09-02 19-12-30

PR #3004, opened 2021-09-06
Sync GS with upstream

Jupi007:
Sync our GS theme with upstream.

**Note:** they dropped entries border, so I reintroduce them.


Feichtmeier (approved):
Thanks! Yes indeed we want inputs with a border! :)

PR #3005, opened 2021-09-06
Sync GTK4 with upstream (remove old color vars)

Jupi007:
They did so much change that I can't tell if we want all of them.
Fortunately, thanks to the symlink system, I didn't have to do anything but delete unnecessary variables in our overwrite color file.

Capture d’écran du 2021-09-06 14-44-37

Capture d’écran du 2021-09-06 14-43-58


Feichtmeier (approved):
Thanks!
They played around with the border darkness, right? Something looks strange, but it is okay I guess? :)

Jupi007:
> Thanks!
> They played around with the border darkness, right? Something looks strange, but it is okay I guess? :)

IMO its okay for now, I think we'll wait things to being stable before doing our "tweaks".

PR #3019, opened 2021-09-20
Make cursor addition black

### LINKED ISSUE #3016: "copy" emblem in mouse cursor has poor contrast with orange selected item colour
Not a big deal, but the little plus icon on the drag-and-drop cursor is orange, and often what you're dragging-and-dropping is also orange (since it was selected), so this results in poor contrast:

yaru-cursor

Perhaps green would be better?

(rough GIMP scribble):

yaru-cursor

Or any other colour that is part of the yaru palette.

Jupi007:
Capture d’écran du 2021-09-20 20-25-33

Capture d’écran du 2021-09-20 20-30-49

Closes #3016


Muqtxdir:
Hi @Jupi007 this look good but I think it'd be hard to view on dark background, so how do you feel about adding a white outline to the + circle?

Jupi007:
Capture d’écran du 2021-09-20 21-05-23

Feichtmeier:
that looks cool :cook:

Issue #3023, opened 2021-09-22
Libhandy (and Libadwaita) icons are missing


Jupi007:
- [x] Add missing icons;
- [x] Add Github action.

![Capture d’écran du 2021-09-22 21-12-50](https://user-images.githubusercontent.com/36476595/134407997-092b58f2-d8d3-4ccd-bac1-f750b8520ae2.png)

See:
- https://gitlab.gnome.org/GNOME/libhandy/-/tree/master/src/icons/scalable
- https://gitlab.gnome.org/GNOME/libadwaita/-/tree/main/src/icons/scalable

PR #3024, opened 2021-09-23
missing symbolic-icons for libhandy and libadwaita

Muqtxdir:
fixes issue: #3023
@Jupi007 please have a look at this and add Github actions


Muqtxdir:
Thanks @Jupi007

Jupi007:
Thanks @Muqtxdir it looks much better with Yaru icons :)
But imo we should prefer use `pan-down` instead of `go-down`.

Muqtxdir:
I used go down beacuse in about page the symbol at hostname for expander is go-right

Jupi007:
Yes because the `go-*` arrow are used to navigate to another screen - on the info page it open a popup :)

IMO this could rather be compared to the color page:

Capture d’écran du 2021-09-22 21-12-50

Or with the system menu:

Capture d’écran de 2021-09-23 10-51-48

Where on both, the `pan` icon is used to explain an expandable box.

Muqtxdir:
sorry, this makes sense now to me. Please change then

Jupi007:
Done :)

Can you review, just in case?

Muqtxdir:
looks good

PR #2868, opened 2021-05-27
Simpler cursor shape / Correct cursor color

Benni3-1415:
# Simpler cursor shape
### Default cursor
- Remove the arrow line at the bottom of the default cursor
- Increase the size by `1px`
- Move the cursor and the _hotspot_ on both axes `1px` towards the corner

Simpler design for the default cursor
Simpler concept design for the default cursor (24px)")

#### Benefits
It's an easier to understand the cursor without the irritating details at the bottom.
This design looks more modern and is also used by Adobe:
The cursor of Adobe Illustrator

### Hand cursors
Increasing the radii of the fingers to remove the vertically line between them.

##### Pointer hand
The radii of the fingers are slightly larger because the index finger is smaller.
Adjusted the back of the hand, the index finger and the _hotspot_ upwards by `1px`, because the back of the hand is much bigger in the grap version.

Simpler design for the hand cursors
Simpler design for the hand cursors (24px)")

#### Benefits
The cursor has less irritating details.
This design looks more modern and is a little more consistent.

### Zoom cursors
Remove the white circle and color the symbol white:
zoom-in-design zoom-out-design

#### Benefits
It's more consistent with the theme, because
- no other cursor has a large white face
- the default line-width of `2px` is no longer interrupted by the dark outline (`1px`)
- the white `+` or `-` fits with the white triangle in the pencil cursor

#### Mimetype
Replace the cursor and remove the border:

mimetype

#### Benefits
- its consistent
- white on light gray is senseless

[go to Discussion >](https://github.com/ubuntu/yaru/discussions/2867#discussion-3383130)

# Correct cursor color

### Color the _watch_ cursors in the secoundary color

[more information >](https://github.com/ubuntu/yaru/issues/2779)

---

Closes #2779
Closes #2874

PR #3115, opened 2021-10-11
Revert cursor mimetype

Benni3-1415:
image-x-cursor

Revert the Cursor mimetype changed in #2868, since the default cursor was also been reverted.
Color the cursor in `#888888`, opacity: `50%`, because other [_system file_-icons are also colored gray](https://github.com/ubuntu/yaru/issues/2271#issuecomment-670846880).

---

Since this is some kind of _bug fix_, it should be included in the Impish branch too. Do I need to create another PR in the Impish branch?


Jupi007:
> yes, we could backport this but it would require someone from ubuntu-desktop team on launchpad to handle and review it (_which generally speaking takes lot of time, if it's not a critical bug_)

Okay, thanks :)

@Benni3-1415 as 21.10 isn't a LTS version, its not really important imo to backport this. As @Muqtxdir said, its not a critical bug and the procedure will be annoying.

So let's just merge this into `master`.
Thanks again for the fix @Benni3-1415 :)

PR #3129, opened 2021-10-19
Fix Gtk4 border-radius

Jupi007:
I removed the upstream symlink of `_common` to decrease border-radius from 6px to 4px.
The `_common` file isn't very long and it isn't very often updated so it isn't very annoying.

Capture d’écran du 2021-10-19 13-57-19


Jupi007:
Oh, I didn't think about that. Well let's keep upstream border-radius so.

PR #3137, opened 2021-10-21
update disabled-symbolics

Muqtxdir:
modified disabled symbolics to use 1px as disabled-line and 45 degree anti-clockwise direction to keep it consistent with suru-style.


Muqtxdir:
alrighty, merging then :+1:

PR #3188, opened 2021-11-03
Sort symbolic action listing

Jupi007:
I noticed we have a problem with the current listing command `find . -name "*-symbolic.svg"`, because it sometimes randomly change the listing sort, and so create a lot of spamming and useless PRs.

So I added a piped `sort` instruction to stabilize the output.
@clobrano do you think this will solve the problem?


clobrano (approved):
Good idea

Jupi007:
> Good idea

Nice, thanks @clobrano :)

PR #3207, opened 2021-11-07
Get some icons from mate branch

Jupi007:
As all these icons can also be useful for non-MATE users, I got some icons from the `mate` branch into master.
Even more, it will be easier to sync the `mate` branch because of less diff.


Muqtxdir:
Thanks @Jupi007 👍

PR #3225, opened 2021-11-13
Add icon theme variant support

Jupi007:
This PR add the possibility to create a fullcolor variant of the `Yaru` icon theme.
I added the `Mate` variant as a demonstrator.

This PR include the following changes:

- Move all fullcolor icons into a `default` subfolder ;
- Create a new `fullcolor/mate` folder with `Mate` icons overwrites ;
- Adapt `render-bitmaps.py` script to support variants: I added a new `--variant` flag which has `default` as default value ;
- Adapt `generate-symlinks.sh` to symlinks icons in all variants ;
- `generate-symlinks.sh` now check if the original icon exist and return an error if not (only in `default` variant) ;
- Adapt meson install script to support variants ;
- Add all `Mate` specifics icons and generate corresponding symlinks.

This will allow us to drop the `*/mate` branch with as benefit to avoid any update problems of this variants.
PRs which will add all missing `Mate` stuff (gtk, ...) will follow.


Feichtmeier (approved):
Very nice idea, thank you

icons/src/fullcolor/render-bitmaps.py
```diff
@@ -236,6 +235,12 @@ def characters(self, chars):
 
 parser = argparse.ArgumentParser(description="Render icons from SVG to PNG")
 
+parser.add_argument(
```
clobrano, line 238: I wonder if you would consider using choices instead of a simple string value (like this https://docs.python.org/3/library/argparse.html#choices).
I see the script use mkdir if the DEST directory does not exist, but we could add a bit more error control if permit to render bitmaps only of some supported variants, in this case "default" and "mate".
Pros are the error check in case of mispelling, Cons is that to add support for a new flavour, we should also modify this script, but it doesn't seem too bad from my point of view.

What do you think?

Jupi007: Thanks for the suggestion @clobrano :)

clobrano:
LGTM 👍

madsrh (commented):
Nice work :muscle:

Maybe add a guide about the variant flag to the [README.md ](https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/README.md)

Jupi007:
> Maybe add a guide about the variant flag to the [README.md ](https://github.com/ubuntu/yaru/blob/master/icons/src/fullcolor/README.md)

You're right @madsrh, thanks :)

Jupi007:
I also noticed a lot of broken symbolics symlinks. It is surly because I regenerated all of them. I'll fix that.

Muqtxdir:
sorry @Jupi007 didn't get the time today to tryout the branch, I'll check this one out tomorrow

Jupi007:
No problem :)

Muqtxdir (approved):

Jupi007:
Thanks @Muqtxdir :)

PR #3227, opened 2021-11-15
Rework .window-caption

Jupi007:
- smaller border-radius
- add border

Capture d’écran de 2021-11-15 14-46-02


Feichtmeier (approved):
Exactly like this :) Ty!

PR #3230, opened 2021-11-18
Fix applications-system icons

Jupi007:
- Fix duplicated icons ;
- fix wrong color sheme (`default` was using `mate`) ;
- fix symlinks.


Muqtxdir (approved):
thanks @Jupi007

PR #3231, opened 2021-11-19
Add GTK mate and mate-dark variants into master

Jupi007:
Add Gtk **mate** and **mate-dark** variants from `mate/jammy` branch into the `master` one.

**Note:** I also noticed the `GtkRange::slider-width` value in **default** variant was different from all other variants so I synced it.


Feichtmeier:
Nice - I like the symlink approach 👍 
Better let @clobrano double check though

Jupi007:
**Important note:** if you want to build this, be sure to delete the previous `build` folder, and also enable **mate** and **mate-dark** options in `meson_options.txt` !

Feichtmeier (approved):

clobrano:
> **Important note:** if you want to build this, be sure to delete the previous `build` folder, and also enable **mate** and **mate-dark** options in `meson_options.txt` !

So, giving `$ meson setup build --reconfigure` is not enough?

I only checked the meson.build files and it looks good to me

clobrano (approved):

Jupi007:
> So, giving `$ meson setup build --reconfigure` is not enough?

I tried to reconfigure the build on my computer but it fails, so I removed the `build` folder.

PR #3232, opened 2021-11-20
Force meson options to true in build CI

Jupi007:
As some meson options are `false`, the build CI doesn't test every part of the repository.
So I forced here all `false` options to `true`.


clobrano (approved):

Muqtxdir:
nice improvement 👍

PR #3234, opened 2021-11-20
Add gtksourceview mate and mate-dark variants

Jupi007:
Add `gtksourceview` **mate** and **mate-dark** variants.

### DIFF
--- a/gtksourceview/meson.build
+++ b/gtksourceview/meson.build
@@ -1,5 +1,5 @@
 flavours = []
-foreach flavour: ['default', 'dark']
+foreach flavour: ['default', 'dark', 'mate', 'mate-dark']
   if not get_option(flavour)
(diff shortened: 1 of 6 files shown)


clobrano:
Sorry for the delay @Jupi007.
While I think this change is working fine, I am concern over setting style changes inside a build script

Jupi007:
It's indeed a bit weird to put styling inside the build script.
I can fix that, but on the other hand it will duplicate files (I'm actually using symlinks).

Btw we have the same situation in the metacity PR #3235

clobrano:
I think in this case it's ok to have duplicate files

Jupi007:
Fixed, merging?

clobrano (approved):

Jupi007:
I'll also remove styling from meson files in the metacity PR.

PR #3235, opened 2021-11-20
Add metacity mate and mate-dark variants

Jupi007:
Add `metacity` **mate** and **mate-dark** variants.


Jupi007:
It is okay here too? @clobrano

clobrano:
> It is okay here too? @clobrano

Yes :+1:

PR #3237, opened 2021-11-22
Fix index.theme flavour overwriting

Jupi007:
The previous way of installing `index.theme` files was very problematic because the `configure_file` meson function is storing temp files in the same in the same temp folder. As they was using the same name, the last one was overwriting the previous one.
So I had the `index.theme` file of the **mate** flavour in both `/usr/share/icons/Yaru` and `/usr/share/icons/Yaru-mate` folders, with very weird side effects.

This PR solve the problem by prefix the file name, and then by rename it at the installation step.


clobrano (approved):
Great solution!

PR #3252, opened 2021-12-05
Use $success instead of $suggested

Jupi007:
Use upstream `$success` color vars instead of our `$suggested` ones.

Related to #3251


Feichtmeier (approved):
😁👌 tricky idea

Issue #3257, opened 2021-12-09
Missing Sim icon (Phosh)


giovannicaligaris:
I have compiled the latest master and the SIM icon is not rendering correctly. it shows a square instead. Please check the images for Yaru and Adwaita.

<img src="https://user-images.githubusercontent.com/26255347/145443189-8b5334f6-a6ba-4cc2-9046-f30e6e6a2767.png" width=35% height=35%> <img src="https://user-images.githubusercontent.com/26255347/145443244-15704834-5323-4ade-9fdd-4a5486b8830b.png" width=35% height=35%>

<img src="https://user-images.githubusercontent.com/26255347/145443751-b094cf94-37e0-4e58-92a0-c1acd2b969f2.png" width=35% height=35%> <img src="https://user-images.githubusercontent.com/26255347/145443813-df3a18fc-fac6-4e3c-9864-14d23cecde27.png" width=35% height=35%>

Muqtxdir:
Hi @giovannicaligaris 

Thanks for taking the time to report this issue. I'll look into this one.

Jupi007:
Hi @giovannicaligaris
How can we reproduce this view?

Muqtxdir:
> How can we reproduce this view?

from what i found, you can install phosh on ubuntu-21.10 with `sudo apt install phosh`  (_you should probably try phosh session in vm_)

PR #3258, opened 2021-12-09
phosh: auth_sim_icon 

Muqtxdir:
Hi @giovannicaligaris could you please build from this branch and see whether this resolves issue: #3257


Jupi007:
@Muqtxdir I don't think it will works because you only modified the icons inside `/src`, without build them.

Also, why did you modified the battery icon? :thinking:

Muqtxdir:
oh I'll push them, I did not modify battery weird that it shows up

giovannicaligaris:
Working.

Jupi007 (approved):
LGFM :+1: 

The battery icon doesn't matter, as it didn't changed anything.

giovannicaligaris:
Battery icon on lockscreen looks great, but is different from the toolbar.

Jupi007:
> Battery icon on lockscreen looks great, but is different from the toolbar.

Yes, but this is another problem.
It looks like we miss some battery icon.
@Muqtxdir can you investigate?

Muqtxdir:
>Yes, but this is another problem.
It looks like we miss some battery icon.
@Muqtxdir can you investigate?

I'll look into this but I think you can merge this. @giovannicaligaris could you open an issue for this one. I think there are many more icons that phosh uses that are not present in yaru.

PR #3264, opened 2021-12-13
Prepare yaru for incoming libadwaita changes

Feichtmeier:
- this removes aubergine from being used since libadwaita will only allow ubuntu to use one accent color
- libadwaita also has a gray close button, to not create inconsistency in the headerbar we use this in gtk3, too
- this also needs a change for the slider color to be some sort of white now, the warm_gray does not work inside an orange switch

TODO:

- [x] adapt gtk2
- [x] adapt icon theme (done in #3265)
- [ ] adapt unity, metacity, xfwm4 parts (_probably in another PR once this PR is finished_)

Closes #2925
Closes #2798

grafik
grafik
gtk4
grafik

PR #3265, opened 2021-12-13
Replace aubergine by orange in some icons

Jupi007:
Replace aubergine by orange in some icons.
I don't think more icons should use orange instead of aubergine, but tell me if you think so.

Capture d’écran du 2021-12-13 22-15-15
Capture d’écran du 2021-12-13 22-18-53
Capture d’écran du 2021-12-13 22-28-18
Capture d’écran du 2021-12-13 22-19-28

Related to #3264


ubuntujaggers:
Do we definitely have to take purple out of icons? It's a nice colour and an app icon doesn't have to be like a tiny screenshot of the app. We had a white folder on the Files icon for ages when we've never had white folders in the theme. Although that might be a bad example because we hated that icon...

Jupi007:
Imo, not all icons need to be switched to orange.
Folders (and so nautilus) and desktop icons are using colors from the default wallpaper which is still orange+purple, so we don't need to change them.

About the icons I modified, it is a question of consistency with changes of #3264 where slides and switches becomes both orange. Leave them purple could looks very strange imo.

ubuntujaggers:
I can see the logic, but I don't think the whole software settings icon needs to be orange even if the slider is... and I think the curved lines on the speaker are an abstract element representing the sound coming out? Can we just make the toggles and sliders orange in the icons?

FWIW I think the current icon theme is full of things that don't look like the corresponding UI element. The equals button on the Calculator icon, the line chart on the System Monitor icon, and the pie chart on the Disk Usage Analyser icon don't look anything like the ones in the app UIs. But tbh I think I'm just bummed that we're losing the beautiful purple accents from the UI :sob:

Jupi007:
I'm also very very sad to lost the purple accent 😔

The speaker icon is only used here:

Capture d’écran du 2021-12-14 10-42-00

And as the full UI uses now orange everywhere for accent things, it will be weird imo to keep purple here.

About the other icons if switches and slides have to be orange, we can keep the purple bg ofc:

Capture d’écran du 2021-12-14 10-39-25

Feichtmeier:
Maybe this is a chance to rework the second "A-Application" with a different icon?
The normal updater looks fine with the gray and the orange slider

ubuntujaggers:
Okay, how about we take the purple A icon out of this PR for now and I'll have a go at some designs, and we can orangify the other icons in this PR :+1:

Jupi007:
I removed `software-properties` from the PR.

PR #3271, opened 2021-12-20
Drop grey selected rows

Jupi007:
Capture d’écran du 2021-12-20 11-24-38


Feichtmeier (approved):

Feichtmeier:
Maybe we should revert this for nautilus only. It looks like the gtk4 port will take much longer. It is kind of easy what to expect in those rows and upstream does it too:

https://github.com/ubuntu/yaru/blob/13e18b5c27c5893520b5617a411a18d549373ca2/gtk/upstream/nautilus/Adwaita.css#L76-L95

PR #3278, opened 2021-12-23
Nautilus: keep gray rows until gtk4 port

### LINKED ISSUE #3277: Keep gray rows for nautilus
We should revert this for nautilus only. It looks like the gtk4 port will take much longer. It is kind of easy what to expect in those rows and upstream does it too:

https://github.com/ubuntu/yaru/blob/13e18b5c27c5893520b5617a411a18d549373ca2/gtk/upstream/nautilus/Adwaita.css#L76-L95

_Originally posted by @Feichtmeier in https://github.com/ubuntu/yaru/pull/3271#issuecomment-999931364_

It's the most opened gnome app on the desktop app we can risk an exception imo. Gnome settings gtk4 port is done I think, so this one will have Grey rows, too

Feichtmeier:
grafik

grafik

Closes #3277

### DIFF
--- a/gtk/src/default/gtk-3.0/apps/_nautilus.scss
+++ b/gtk/src/default/gtk-3.0/apps/_nautilus.scss
@@ -1,5 +1,48 @@
 .nautilus-window {
 
+    // Due to the ubuntu orange being quiet bright
+    // we change the nautilus row selection from a full orange
+    // row to a soft gray selection
+    // until the nautilus gtk4 port lands in Ubuntu
+    $_row_hover_bg: if($variant=='light', transparentize(black, 0.96), transparentize(white, 0.96));
+    $_row_selection_bg: if($variant=='light', transparentize(black, 0.91), transparentize(white, 0.91));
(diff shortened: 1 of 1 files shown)


Jupi007:
I say yes :+1: 

What about also revert treeview tweaks here?

Feichtmeier:
better not there was some usability issue to whether this small row is selected or hovered

Jupi007:
This can be fixed imo. If we use a very very light one for hover.
And honestly, it is easy to know if a row if hovered, as you have your cursor on it 😅

Feichtmeier:
I am fine with this but maybe you make the push then. Last time I tried we had someone opening an issue about the treeview hover/selection ^^

Jupi007:
> Edit: on the other hand would you have a "inconsistency" between the icon view selection and the treeview selection shrug

Well, let's make it grey too :)

```css
widget.view:selected {
  background: grey;
}
```

Jupi007:
I only works in inspector :/
Well, let's keep treeview and regular view selection orange so.

Jupi007 (approved):

Issue #3216, opened 2021-11-09
"Software & Updates" and "Software Update" icons are too similar.


luciotorre:
**Expected Behavior**

When trying to open the "Software Updater" app I type "update" in the search bar of gnome shell and icons come up. Ideally it would be easy to identify which icon i should click.

**Actual Behavior**

See attached screenshot. When one searches for "update" two icons appear which not only are very similar (both look like A's) but the clipping of the names makes it impossible to *know* without thinking which one one should click.

There are clues in the icons: the updater has a refresh like circle and the Software & Updates app has something that could be considered a toggle/slider. But this is very cryptic.

I don't know that the solution is a better icon. Maybe the solution also requires changing the names of the apps. Or unifying them. Or whatever. I just wanted to leave this here in case it is useful.

**Software that presents the issue**

"Software Updater" and "Software & Updates"

**Yaru version**

Package: yaru-theme-icon
Version: 21.10.2
Priority: optional
Section: misc
Source: yaru-theme
Origin: Ubuntu
Maintainer: Ubuntu Desktop Team <ubuntu-desktop@lists.ubuntu.com>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 117 MB
Homepage: https://community.ubuntu.com/c/desktop/theme-refresh
Task: ubuntu-desktop-minimal, ubuntu-desktop
Download-Size: 19.5 MB

**Screenshots**

![image](https://user-images.githubusercontent.com/469561/141003585-9ac38b72-5c77-4e84-998c-9daf147ce247.png)

madsrh:
Hi @luciotorre 
Thanks for taking the time to report this 🙏 
Sadly, we're well aware of the issue, but haven't come up with a solution. You can read up on earlier ideas in these threads:
https://github.com/ubuntu/yaru/issues/1051
https://github.com/ubuntu/yaru/issues/2292
https://github.com/ubuntu/yaru/issues/1940
https://github.com/ubuntu/yaru/issues/863
https://github.com/ubuntu/yaru/issues/862

Best solution IMHO would be to create just a single application for everything related to updates, but I doubt that will happen.

Muqtxdir:
This was addressed and re-drawn with PR: #3279

PR #3280, opened 2021-12-23
Adding reworked software-properties icon

### LINKED ISSUE #3279: Rework Software & Updates icon
grafik

With the switch to orange the strong purple on the baseplate does not fit anymore.
However as discussed in https://github.com/ubuntu/yaru/pull/3265#issuecomment-993967478
and due to various feedback from users how indistinguishable the two icons are a rework would be nice. Eventually step away from the A-metaphor for this icon completely. 

Note: we discussed this several times and thought that the "great rework" of gnome software and how ubuntu manages its updates would come "soon" so this app would basically be included elsewhere. Since this never happened and eventually never will we need this rework.

ubuntujaggers:
Have a go at this and see if we want further tweaks...

image

Closes #3279


Muqtxdir:
I really like the new icon @ubuntujaggers but it would be better in terms of differentiation between ***Software & Updates***  and ***Software Updater*** if **A** is dropped the from ***Software & Updates*** (_see #3216 for this_)

Feichtmeier:
> I really like the new icon @ubuntujaggers but it would be better in terms of differentiation between ***Software & Updates***  and ***Software Updater*** if **A** is dropped the from ***Software & Updates*** (_see #3216 for this_)
> 

I was thinking the same. But now the icon is so much different than the update notifier and the cog wheel metaphor is so strong that I think we can keep the A

ubuntujaggers:
Back after the festivities!  I think having the same pictogram on both is probably okay because Software & Updates is basically the settings for Software Updater?  I mean, we don't actually like the "A" much :slightly_smiling_face: but we've still never been able to come up with a better simple pictogram... and whatever we go for, I don't think it hurts to have some commonality between the two related apps?

PR #3308, opened 2022-01-14
Add phosh-splash styling

### LINKED ISSUE #3261: Dark mode splash (Phosh)
When using the Yaru-dark theme the Splash on Phosh should be black, instead it defaults to white.

The white image is using Yaru-dark and the black one Adwaita Dark.

Jupi007:
@giovannicaligaris can you try this branch and tell me if splash is correctly styled?

Closes #3261

### DIFF
--- a/gtk/src/default/gtk-3.0/_apps.scss
+++ b/gtk/src/default/gtk-3.0/_apps.scss
@@ -7,6 +7,7 @@
 @import 'apps/gnome-disks';
 @import 'apps/nemo';
 @import 'apps/not-csd';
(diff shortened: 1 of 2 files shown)


giovannicaligaris:
It is still showing white on both Yaru and Yaru-dark

Jupi007:
As I was afraid of it, this cannot be fixed.

We should create a new component like **gnome-shell**, but it wouldn't even help because **phosh** doesn't support custom theme.

I'm really sorry @giovannicaligaris, but this is not fixable :(
I'll close this PR and the related issue.

PR #3316, opened 2022-01-19
Revert close button to upstream look

Jupi007:
As we dropped the orange close button in Gtk, I think we should also revert our change of orange close for Gnome Shell, so it looks consistent.

Capture d’écran de 2022-01-19 13-39-06

What do you think of?


Feichtmeier (approved):

Feichtmeier:
maybe ad a little border around it? For the contrast

Jupi007:
Very nice idea! :)

Jupi007:
This looks pretty good:

Capture d’écran de 2022-01-19 18-17-03

PR #3317, opened 2022-01-19
Fix Yaru-dark gtk2 assets

Jupi007:
Fix Yaru-dark gtk2 assets which was using **mate** colors.

Very honestly; I don't know how I did that 😥


Feichtmeier (approved):
Woops 😸

Issue #3307, opened 2022-01-14
mate:  make the panels same across both default and dark


Muqtxdir:
the `default` theme has light-colors on both top and bottom-panel. Now that all panel icons are present under `unity-panel` and are used in those panels. Its better to have the panels in dark colors as the panel-icons are `white` in color.

#### screen-recording:

https://user-images.githubusercontent.com/54065734/149505028-2566c7bf-fb59-456b-a7ff-a89048931088.mp4

PR #3325, opened 2022-01-23
mate: add panel-icons

Muqtxdir:
- [x] add the panel icons used by MATE from `Ambiant-MATE`:
   - [x] 16x16
   - [x] 22x22
   - [x] 24x24
- [x] align audio-speaker icons with page-grid.
- [x] rename `unity-panel` to `panel`
- [x] update `meson.build` to install `panel` subdir by default as it is used by both unity and MATE.

_Note: For these panel-icons to work in **Yaru-mate** (light theme), the panels should be dark (see issue: #3307)_


icons/meson.build
```diff
@@ -3,23 +3,22 @@ fs = import('fs')
 foreach flavour: ['default', 'mate']
   if get_option(flavour)
     suffix = flavour == 'default' ? '' : '-@0@'.format(flavour)
+    icon_dir = 'Yaru' + suffix
```
Jupi007, line 6: @Muqtxdir why did you created this new var?
The `theme_name` one is exactly the same.

Muqtxdir: I added `icon_dir` so that if someone changes `meson.project_name()` from `Yaru` to something else (_for ex: `MyTheme`_), `theme_name` wouldn't be sufficient for that as it would look for `MyTheme` and `MyTheme-mate` icon folders instead of `Yaru` and `Yaru-mate`

Thanks for the review !

Jupi007: It's a very nice idea!
I never thought at this problem :)

But maybe doing this in a separated PR?

Muqtxdir: ok, I'll separate them in another PR and revert that change.

Jupi007: Thanks :)

Muqtxdir: done, I've reverted the changes made to `icons/meson.build`

Jupi007 (approved):
LGFM :+1:

PR #3330, opened 2022-01-26
Dash to dock updates for GNOME 41 (and dash-to-dock 71+)

### LINKED ISSUE #3111: The dock and the panel have different background colors and the dock is not transparent
grafik

We tried to band-aid fix this in the cycle but it didn't really work out. 
I think this looks really unpolished and needs a fix.

@3v1n0 anything you could do here? Something changed in the dock code from 21.04 to 21.10 so the styling is not correctly applied anymore.

### LINKED ISSUE #3327: LP#1942702 "Ubuntu Dock "Show Applications" button
Reported first on Launchpad at https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1942702

3v1n0:
Update dash-to-dock base CSS to respect latest upstream changes, fixing
the padding for edge icons.

Moved the logic to handle most of style such as color and transparency
of the dock here, as it's something we'd prefer not to change through
settings overrides.

Fixes #3111
Fixes #3327
LP: #1942702

---

There's some results, this requires also some dash-to-dock changes in order not to define colors via settings overrides anymore.

Screenshot from 2022-01-26 04-48-37

Screenshot from 2022-01-26 04-49-01

I'm also wondering if we could do something like this when the applications are showing as IMHO the show-apps button is easy to confuse with applications when it's checked:

```diff
diff --git a/gnome-shell/src/gnome-shell-sass/_dock.scss b/gnome-shell/src/gnome-shell-sass/_dock.scss
index 969d22d47..4975662b5 100644
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -383,6 +383,14 @@ $dock_color: $panel_bg_color;
                     }
                 }
             }
+
+            .show-apps {
+                &:active, &:checked {
+                    .overview-icon {
+                        background-color: $accent_bg_color;
+                    }
+                }
+            }
         }
 
         .notification-badge {
```

Which leads to:

Screenshot from 2022-01-26 04-52-04

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -1,15 +1,29 @@
-/* Default Ubuntu Dock styling, append !important to any changed rules */
+/* Default Ubuntu Dock styling, append !important to any changed rules
+   This must be in sync with
+   - https://github.com/micheleg/dash-to-dock/blob/ubuntu-dock/_stylesheet.scss
+ */
+
+$dash_bottom_margin: $base_margin * 4;
+$dash_spacing: round($base_padding / 4);
 
 // Stock
-$dash_spacing: $base_padding / 4;
-$dash_bottom_margin: $base_margin * 4;
-$dock_side_margin: $dash_bottom_margin / 4;
+$dash_edge_items_padding: $dash_padding - $dash_spacing;
+$dock_start_margin: $dash_bottom_margin;
+$dock_side_margin: $dock_start_margin / 4;
 $dock_fixed_inner_margin: $dock_side_margin;
 
 // Adapted to $dock_bottom_margin
 
 @function shrink($val) {
     @return round($val / 4);
+}
+
+@function shrink_light($val) {
+    @return round($val * 0.75);
+}
+
+@function is_horizontal($side) {
+    @return $side == top or $side == bottom;
 }
 
 @function opposite($val) {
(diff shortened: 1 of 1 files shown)


Jupi007:
Thanks @3v1n0

Personally, I think we should avoid to use the orange color for such large things as checked buttons.

3v1n0:
Yeah, the orange was just a quick test... But the point is that IMHO we should underline more when that button is checked and an app is focused.

Jupi007:
Yeah, maybe 🤔

We changed the `:checked` buttons look from darker to this lighter bg, because it looks better.
Maybe could we try to revert this for the `.show-apps` button ?

Jupi007:
@3v1n0 Is the spacing between the first item of the dock and the panel normal ?
It looks like it is twice the space used everywhere else on the dock:

Capture d’écran de 2022-01-30 11-18-23

DoctorRover:
> Moved the logic to handle most of style such as color and transparency
> of the dock here, as it's something we'd prefer not to change through
> settings overrides.
Will the option to disable the dock transparency still remain throug gsettings?

3v1n0:
> Will the option to disable the dock transparency still remain available through gsettings?

You will need to override the color (and transparency to do that). So yeah... It's still possible via gsettings.

3v1n0:
@Jupi007, well we already had some padding on top and it was justified by the fact that we didn't want the top/bottom icons to be too close to the edges. Mostly because the "Activities" button could otherwise be too close to it.

Jupi007 (approved):
Thanks @3v1n0 now I understand the reason :)

So it's a LGFM :+1:

DoctorRover:
> You will need to override the color (and transparency to do that). So yeah... It's still possible via gsettings.
Great. Thanks!

PR #3331, opened 2022-01-26
icons: modify meson and deb files

Muqtxdir:
- add icons_dir variable in icons/meson.build.
- update `yaru-theme-icon.install` to include panel-icons, as they are installed by default.
- remove panel-icons from `yaru-theme-unity.install` as they will already be present in `yaru-theme-icon.install`.


Jupi007:
I have a few notes :)

- Is the `unity` package still needed if the panel icons are installed by default?
- In `meson.install` I think you should also replace `theme_name` by `icon_dir` lines 30 and 32.
- Also, can you rename the `theme_dir` var to `install_dir`? This looks like a better name for me :)

Muqtxdir:
> - Is the `unity` package still needed if the panel icons are installed by default?

currently after this change `yaru-theme-unity` will only contain the assets, I think it's better to include those assets in `yaru-theme-gtk` since we do that for `metacity`, `xfwm4` and `cinnamon` as well. wdyt?



> * In `meson.install` I think you should also replace `theme_name` by `icon_dir` lines 30 and 32.
> * Also, can you rename the `theme_dir` var to `install_dir`? This looks like a better

ok, will make these changes :+1:

Jupi007:
> currently after this change `yaru-theme-unity` will only contain the assets, I think it's better to include those assets in `yaru-theme-gtk` since we do that for `metacity`, `xfwm4` and `cinnamon` as well. wdyt?

Wouldn't it be complicated to drop a package ?
Honestly, after thinking about this, I think we should keep the unity package.

> ok, will make these line-changes +1 but renaming `theme_dir` to `install_dir` would cause a bit of confusion since meson also has `install_dir`

You're right, maybe `theme_install_dir` so? Unless you have a better idea :)

Muqtxdir:
> Wouldn't it be complicated to drop a package ?
Honestly, after thinking about this, I think we should keep the unity package.

ok, makes sense. 

> You're right, maybe `theme_install_dir` so? Unless you have a better idea :)

this one's good. thank-you!

Jupi007 (approved):
LGFM :+1: 

Thanks @Muqtxdir :)

PR #3337, opened 2022-01-31
Add Cinnamon theme upstream fetch

Jupi007:
Add Cinnamon theme upstream fetch (https://github.com/linuxmint/mint-themes/tree/master/src/Mint-Y/cinnamon).


Jupi007:
This is working, merging so.

PR #3346, opened 2022-01-31
Sync Gnome Shell with upstream

Jupi007:
They added a new screenshot widget.
But we need GS42 to test this.

I added the corresponding assets:

Capture d’écran du 2022-01-31 11-28-43

See #3343 and #3344


Feichtmeier (approved):
looks very yaruish :+1:

PR #3350, opened 2022-02-01
Fix padlock-open-symbolic icon

Jupi007:
Fix `padlock-open-symbolic` icon, which was using Adwaita look.

**Before:**

Capture d’écran du 2022-02-01 21-20-03

**After:**

Capture d’écran du 2022-02-01 21-18-42


Muqtxdir:
Thanks @Jupi007 👍

PR #3361, opened 2022-02-03
Sync GCC symbolic icons with upstream 3359

Jupi007:
Sync symbolic icons with Gnome Control Center upstream changes.

See #3359 

_Note: I added a new icon:_

Capture d’écran du 2022-02-03 18-58-36


Muqtxdir (approved):
thanks for the symlinks and the new icon!

PR #3374, opened 2022-02-06
Sync icon theme with upstream

Jupi007:
Sync icon theme with `gnome-software` upstream changes.

See #3369


Muqtxdir (approved):
thanks for the symlinks!

PR #3387, opened 2022-02-11
Add review-symbolic icon 3383

Jupi007:
### New icon

**review-symbolic:**
Capture d’écran du 2022-02-11 12-50-08

Related to #3383


Muqtxdir (approved):
thanks @Jupi007!

PR #3403, opened 2022-02-24
GNOME Shell 42 theme sync

3v1n0:
Synchronize the yaru code to follow upstream changes.
Some changes have been dropped as not easy to apply to the new codebase
but they're very limited.

I will very likely want to merge this by tomorrow as per the FF, but we can do followups in future.

Only dark variant has been tested and seem to work as expected here.

There are also few new icons (added now to a new gresource file, to match upstream shell) that would need the _yarufication_.


Feichtmeier (approved):
Feature freeze doesn't mean UI freeze, right?
I looked over the changed files and the most important part will be to exchange the adwaita icons with yaru (as you said).
As the rest was already almost unchanged, let's merge it but then make really sure that the other changes go in before UI freeze this time :smile:

3v1n0:
> Feature freeze doesn't mean UI freeze, right?

No, sure... There's still time for UI changes, but given that yaru should go with newest shell I wanted it at least in an acceptable state.

Another thing that I think needs love is the handling of light theme, as I didn't test that at all so far.

PR #3408, opened 2022-02-24
Sync filemanager-app icon

Jupi007:
Capture d’écran du 2022-02-24 20-03-56

CC @flexiondotorg


Muqtxdir (approved):
Thanks @Jupi007, this was quick!

PR #3422, opened 2022-02-27
Add start-here-symbolic mate icon

Jupi007:
- add start-here-symbolic mate icon ;
- add `/scalable` to **index.theme** of mate icon theme.


Muqtxdir (approved):
Thanks @Jupi007

PR #3425, opened 2022-02-28
Fix keyboard in light and dark theme

Jupi007:
Fix keyboard in light and dark theme.

**Before:**

Capture d’écran du 2022-02-28 21-29-36
Capture d’écran du 2022-02-28 21-29-50

**After:**

Capture d’écran du 2022-02-28 21-28-44
Capture d’écran du 2022-02-28 21-29-03

## Video:

https://user-images.githubusercontent.com/36476595/156054455-4e85c13b-f121-452e-8867-161264a7fcc3.mp4

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_drawing.scss
+++ b/gnome-shell/src/gnome-shell-sass/_drawing.scss
@@ -102,19 +102,19 @@
   // hover key
   @else if $t==hover {
     color: $tc;
-    background-color: lighten($c, 7%);
+    background-color: if($variant=="light", lighten($c, 1%), lighten($c, 7%)); // Yaru change: Make keyboard work on both variants
   }
 
   // active key
   @else if $t==active {
     color: $tc;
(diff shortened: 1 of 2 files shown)


Feichtmeier (approved):
Oh man! Every damn release. We should upstream this!
thank you so much for going through this again :clap:

Jupi007:
Yeah, something I really don't understand is why GS got some kind of support for light theme, but it's completely broken 🤔

Muqtxdir:
> Yeah, something I really don't understand is why GS got some kind of support for light theme, but it's completely broken 

I think light theme is used in GNOME classic session

Feichtmeier:
Yes. Red hat enterprise uses this gnome classic session as default (among a bunch of gnome shell extensions :shrug: ). Which does not have the overview, so basically only the panel and popovers are "light theme" compatible

Jupi007:
Oh, I didn't know that.
Thanks for the answer guys :)

PR #3436, opened 2022-03-02
Add missing gnome-shell app icons (fullcolor and symbolic)

Jupi007:
Capture d’écran du 2022-03-02 09-35-35
Capture d’écran du 2022-03-02 09-35-19


Muqtxdir:
ok, makes sense since all icons are moved into `gnome-shell-icon-theme.gresource`

PR #3438, opened 2022-03-03
Fix close button hover bg

Jupi007:
Fix over bright hover close button.

**Before:**

Capture d’écran du 2022-03-03 09-25-12

**After:**

Capture d’écran du 2022-03-03 09-20-28

### Explanation

Before the sync with GS42, the default state was using 10% lighten of `$osd_bg_color` and hover 15% (so it's 5% lighter than normal state).
With the GS42 sync, there's a new var `$window_close_button_color` with the same value as normal state (10%). But the hover state is lighten this var to 15% again (so 25% which is really too bright).
So to fix that, I decreased the hover state lighten to 5%, so it looks like before.


Feichtmeier (approved):
Ty!

PR #3445, opened 2022-03-04
Sync Gnome Shell with upstream 3444

Jupi007:
Related to #3444


Feichtmeier (approved):
Can't test this currently but if you say it's good to go, proceed

Jupi007:
I tested it, it's a bug patch.
We can safety merge it.

Issue #3407, opened 2022-02-24
Update GNOME Shell 42 assets and tune theme


3v1n0:
- [x] Add missing yaru icons (per new GNOME Shell Icon theme) [ Handled in https://github.com/ubuntu/yaru/pull/3421 ]
- [ ] Review yaru-light theme (stuff such as the keyboard had special code that is now gone)
- [ ] Do another full-review :) `diff -u gnome-shell/upstream/theme/gnome-shell-sass/ gnome-shell/src/gnome-shell-sass/ -r`, and add missing `// Yaru` comments for the missing stuff or drop our changes if are not-intentional
- [ ] Button (notification / dialogs action) colors have to be checked (use https://github.com/phuhl/notify-send.py to test)
- [ ] Check missing `&:insensitive` in `.modal-dialog-button` at login-dialog`

_Originally posted by @Feichtmeier in https://github.com/ubuntu/yaru/pull/3403#pullrequestreview-892037695_

3v1n0:
Ok, one thing that is definitely to be fixed are notification actions:

![image](https://user-images.githubusercontent.com/345675/155546332-1afc0472-089d-4721-bd12-fb3f1d24f00a.png)

PR #3447, opened 2022-03-04
Fix focused unlock-dialog entry

Jupi007:
As the lock screen entry is transparent and as we are using a box-shadow for focused entries, it result a very weird effect of ugly background.

**Before:**

Capture d’écran du 2022-03-04 12-09-49

**After:**

Capture d’écran du 2022-03-04 11-58-48

Related to #3407


Feichtmeier:
Could you show focused/unfocused?

Jupi007:
I tried to click outside of the entry and use tab key but I can't unfocus the entry.

Feichtmeier (approved):

PR #3450, opened 2022-03-05
Sync icon theme with upstream

Jupi007:
- Add new `-rtl` icons ;
- Remove some `media-*-rtl` ones see this [upstream commit](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/commit/be4e6bf5ce3899f638ab38da84d7af227d3b99bf).

Related to #3448


Muqtxdir:
@Jupi007 since most applications during the 22.04 release are mostly from `GNOME-41`. They might still make use of the older icons. So I suggest to not remove the old-named symbolic icons.

Muqtxdir (approved):

PR #3468, opened 2022-03-08
Improve icons build scripts

clobrano:
Rework generate-symlink.sh to improve readability of the code and of the error reports

### DIFF
--- a/icons/src/symlinks/generate-symlinks.sh
+++ b/icons/src/symlinks/generate-symlinks.sh
@@ -83,84 +83,99 @@ VARIANTS=("default" "mate")
 
 # Fullcolor icons
 echo "Generating links for fullcolor icons..."
-# variants for loop
+symlinks_counter=0
 for VARIANT in "${VARIANTS[@]}"
 do
 	[[ $VARIANT = "default" ]] && THEME="Yaru" || THEME="Yaru-${VARIANT}"
-	# contexts for loop
 	for CONTEXT in "${CONTEXTS[@]}"
 	do
 		dlog " -- "${CONTEXT}
-		# Sizes Loop
 		for SIZE in "${SIZES[@]}"
 		do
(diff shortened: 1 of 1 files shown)


Jupi007:
@clobrano Can you import the change from this PR: #3452

clobrano:
> @clobrano Can you import the change from this PR: #3452

oh, I am sorry, I didn't see the review request. I will do the review and then I can rebase this PR on top of it

Jupi007:
No problem, thanks :)

clobrano:
Hey @Jupi007, after rebasing onto your master the rework was even easier :+1: 

Please have a good test because I might have missed some use case that this rework might affect

Jupi007:
Okay, I'll take a look at this.

Jupi007:
Sorry for the long wait.
I just tested the script.
It looks working as expected.
The console print return is much better :)

I have a only one notice. The example message is not appropriated:
```
example:
     To generate only the links defined in ./symbolic/apps.list
     $ generate-symlinks.sh --match apps
```
It should be something like:
```
example:
     To generate only power-profile-* symlinks
     $ generate-symlinks.sh --match power-profile
```
Because the `match` option doesn't looks at the symlink list file name but directly match strings inside them like this kind:
```
power-profile-balanced-symbolic.svg speedometer-symbolic.svg
```
The previous symlink will be generated as it contains `power-profile`.

Except that, great job :)

clobrano:
> I have a only one notice. The example message is not appropriated:

you're right, I will change it, thanks

> **Edit:** by running the script in `--all` mode, some symlinks that were not present are generated. Do you want to commit them here or doing this in another PR?

better doing it in a different PR, if it is Ok for you

3v1n0:
This looks good to me as well, let's get in as I'd have to tune the script a bit :)

PR #3416, opened 2022-02-25
appearance: add accent colors to yaru

Muqtxdir:
_Note: This just sets-up the necessary files under `gtk` for accent-colors, Colors used in this PR are temporary and acts as placeholder._

### TO-DO:

- [ ] gtk-2.0 (_symlinked at the moment_)
- [x] gtk-3.0 
- [x] gtk-4.0
- [x] icons for recolor by `accent_colors`:
     - [x]  filemanager-app (_provide another SVG as it needs a black/white gradient_)
     - [x] gnome-tweak-tool
     - [x] software-updater
     - [x] emblem-symbolic-link
     - [x] folders-* (_provide another SVG as it needs a black/white gradient_)
     - [x] user-desktop (_provide another SVG as it needs a black/white gradient_)
     
     _Note: also look into symlinks for fullcolor-icons_ 

#### screen-reocord:
https://user-images.githubusercontent.com/54065734/155732402-1b8355a4-33ec-4e33-a0e3-bfdd61e92a85.mp4

PR #3486, opened 2022-03-15
Improve gtk accent colour palette

lyubomir-popov:
- Introduce an improved colour palette: 8 colours that comply with accessibility contrast ratio requirements:
- Remove scss transformations on accent colours, as they are not needed (the proposed colours are carefully selected to work with both light and dark themes)
-  Small refactor to separate the light and dark theme. Instead of inline if statements for every colour, 2 blocks of code - one for the light theme, one for the dark theme. This will make it easier to audit them and work on them independantly.

In the image below, the numbers inside colour swatches show that the minimum required contrast ratios from the WCAG spec are met: 
- Text under 14px requires a minimum of 4.5:1
- Graphical elements and text above 14px/bold requires a minimum of 3:1

image

Below a list of screenshots of all proposed colours in both the light and the dark themes:

image
image
image
image
image
image
image
image
image
image
image
image
image
image

### DIFF
--- a/gtk/src/default/gtk-3.0/_colors.scss
+++ b/gtk/src/default/gtk-3.0/_colors.scss
@@ -2,242 +2,82 @@
 // it gets @if ed depending on $variant
 @import 'palette';
 
-$base_color: null;
-$text_color: null;
-$bg_color: null;
-$fg_color: null;
+$base_color: if($variant == 'light', #ffffff, lighten($jet, 6%));
+$text_color: if($variant == 'light', black, white);
+$bg_color: if($variant == 'light', #FAFAFA, lighten($jet, 8%));
+$fg_color: if($variant == 'light', $inkstone, $porcelain);
 
-$selected_fg_color: null;
-$selected_bg_color: null;
-$selected_borders_color: null;
-$borders_color_light: null;
-$borders_color_dark: null;
-$borders_color: null;
-$alt_borders_color: null;
-$borders_edge: null;
-$link_color: null;
-$link_visited_color: null;
-$top_hilight: null;
-$dark_fill: null;
-$headerbar_color: null;
-$menu_color: null;
-$popover_bg_color: null;
-$popover_hover_color: null;
+$selected_fg_color: $accent_fg_color;
+$selected_bg_color: $accent_bg_color;
 
-$scrollbar_bg_color: null;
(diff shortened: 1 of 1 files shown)


Jupi007:
Hi @lyubomir-popov,
all those color are nice, but I see two problem with the refactoring of `_colors`:
- you define the vars name three times, and so the file is three times bigger and it add a lot of duplication ;
- and we aim to be as close as possible to upstream file and this modification totally break this rule.

So except if you think this change is really needed, can you revert this file to it's original look please 🙂

lyubomir-popov:
@Jupi007 Yes, as @3v1n0 said we agreed to revert it, I'll push an update to my pr. 

As for my reasons to initially go this way - doing any extensive theming work with hundreds of inline ifs would be quite difficult, so I still think it's a good idea to pitch this refactor upstream, but not something that has to happen now by any means.

Regarding the duplication - it only increases the file size of the scss, not the generated css.

Jupi007:
@lyubomir-popov Okay, thanks for the explanation.

Honestly, Gtk3 is very stable since a while.
We do not plan to do more tweaks, and upstream didn't do any stylesheet update since a while.
We'd like to let Gtk3 die quietly and peacefully, that's why heavy modifications are now imo useless :)

lyubomir-popov:
> We'd like to let Gtk3 die quietly and peacefully, that's why heavy modifications are now imo useless :)

Didn't know that, thanks for the additional context. Makes sense to leave it be then.
That said, the above applies to whatever theming mechanism succeeds it.

lyubomir-popov:
@3v1n0 I've updated the colour names and restored the _colors.scss. I've removed some color transformations that I think are unnecessary. Please have a look and let me know if you need anything else.

(I've also added fancier colour names like viridian, olive, bark)

lyubomir-popov:
@Muqtxdir yes, they are final. The order should be as in the scss file - they are ordered by hue, from the default ubuntu orange on the left to the red on the right:

image

Muqtxdir:
thanks for the clarification @lyubomir-popov I'll remove any transformations on the orange selector and arrange all the colors as you've suggested based on the hue.

3v1n0:
Closing this as it has been merged in the branch that is part of #3416

PR #3492, opened 2022-03-18
metacity: Correct Metacity (Marco) theme contrast so they match GNOME/CSD windows

flexiondotorg:
This pull requests update the Metacity default and dark themes so that windows and window buttons match their equivalents present on GNOME/CSD windows.


Jupi007 (approved):
LGFM :+1: 
Thanks @flexiondotorg ❤️

PR #3496, opened 2022-03-18
gtk accent colours: subtler purple, magenta and red

lyubomir-popov:
Adjust some of the accent colours to be less saturated and more in line with the rest of the palette.

Current set:

image

The colours have been tested against white and black text, and meet the minimum contrast ratio required by [WCAG 1.4.11](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html) (This minimum is 4.5:1). The numbers inisde the colour swatches represent the actual contrast against white and black text respectively.

### DIFF
--- a/common/accent-colors.scss.in
+++ b/common/accent-colors.scss.in
@@ -15,11 +15,11 @@
     } @else if $accent_color == 'blue' {
         $color: #0073E5;
     } @else if $accent_color == 'purple' {
-        $color: #8856EB;
+        $color: #7764D8;
     } @else if $accent_color == 'magenta' {
-        $color: #BC33DB;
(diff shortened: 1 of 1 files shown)


3v1n0:
Merged (manually).

Muqtxdir:
hi @lyubomir-popov 

I just tried out the newer-changed colors, the purple and red look nicer but magenta-color seems to have shade closer to lighter aubergine and it does not look like magenta-shade now

### screenshot:
Screenshot from 2022-03-19 22-52-09

lyubomir-popov:
@Muqtxdir pr to fix the magenta here: https://github.com/ubuntu/yaru/pull/3526

PR #3520, opened 2022-03-20
Increase contrast of popover menus

### LINKED ISSUE #3490: LP#1965210 "Grey (gray) panel menus don't contrast with the grey login screen"
Reported first on Launchpad at https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/1965210

Jupi007:
Use a darker `$bg_color_dark` value and a lighten border.

**Before - `lighten($jet, 6%)`:**

Capture d’écran du 2022-03-20 11-31-36

**After - `lighten($jet, 2%)` + lighten border:**

Capture d’écran du 2022-03-20 11-46-23

This will affect all the popovers, but as we have the same issue with the overview which use the same bg as login screen, this is imo not a problem.

Closes #3490

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_colors.scss
@@ -3,7 +3,7 @@
 @import 'palette';
 
 $base_color: if($variant=='light', #ffffff, lighten($jet, 4%));
-$bg_color_dark: lighten($jet, 6%);
+$bg_color_dark: lighten($jet, 2%);
 $bg_color: if($variant=='light', #FAFAFA, $bg_color_dark);
 $fg_color_dark: $porcelain;
 $fg_color: if($variant=='light', $inkstone, $fg_color_dark);
@@ -17,7 +17,7 @@ $selected_borders_color: if($variant=='light', darken($selected_bg_color, 15%), 
 $borders_color_dark: lighten(desaturate(lighten($jet, 4%), 100%), 14%); // Yaru: used for dash and other dark elements on light theme
 $borders_color: if($variant=='light', darken($bg_color, 20%), $borders_color_dark);
 $alt_borders_color: if($variant=='light', darken($bg_color, 24%), darken($bg_color, 10%));
-$borders_edge: if($variant=='light', transparentize(white, 0.2), transparentize($fg_color, 0.93));
+$borders_edge: if($variant=='light', transparentize(white, 0.2), transparentize($fg_color, 0.85));
 
 $link_color: if($variant=='light', darken($selected_bg_color, 10%), lighten($selected_bg_color, 20%));
 $link_visited_color: if($variant=='light', darken($selected_bg_color, 20%), lighten($selected_bg_color, 10%));


Feichtmeier (approved):

vanvugt:
I don't think that solves the issue. Seems like the same level of contrast, just inverted.

Would a bolder shadow around it all work? I've spent the last couple of years fixing gnome-shell's shadow rendering so let's use it :slightly_smiling_face:

Feichtmeier:
I don't think we should alter the popover look any further. People spend 5 seconds on the screen and very rarely open up the calendar. If the contrast is really such a big issue, then we need to lighten up the gray background

vanvugt:
> I don't think we should alter the popover look any further. People spend 5 seconds on the screen and very rarely open up the calendar. If the contrast is really such a big issue, then we need to lighten up the gray background

It's more for the power menu than the calendar.

Jupi007:
It is better like this (reverted bg color + stronger shadow):

Capture d’écran du 2022-03-21 10-23-51

> It's more for the power menu than the calendar.

I took a screenshot of the calendar because it let me having more space around in the image than the system menu :)

vanvugt:
The stronger shadow seems like a reasonable compromise... Unless:

  * Canonical senior management or design dislike it (we'll find out later); or
  * Someone cites a vision/display problem, which may yet happen after release

Feichtmeier:
Alternatively we could just revert back to the old purple GDM :shrug:

Jupi007:
No, really please don't :)
Those dark entries blend very bad with the purple bg.

Jupi007:
Can I merge this, or are we waiting for any approval from the design team?

Feichtmeier:
I am not sure if this is really good visually. I feel that the shadow is too strong now. Could you try to just brighten up the gdm gray instead and revert to the current master popovers?

Jupi007:
Okay, I'll try this :+1:

Jupi007:
Capture d’écran du 2022-03-23 21-25-36

@Feichtmeier ?

Feichtmeier:
did you brighten up the gdm bg or the popover in this picture? 🤔

Jupi007:
I only lighten the popover bg 😅

Feichtmeier:
Oh... I meant, could you try to not change the popovers at all but instead increase the gdm BG lightness?

Jupi007:
Oh, yes I can try that.
In fact I didn't do it because this require to patch GS to sync the login animation background.
But let's try it anyway :)

Jupi007:
It looks like this if I only lighten `$system_bg_color` (I kept the lighten border-color):

Capture d’écran du 2022-03-23 22-44-05

But then the overview will looks like:

VirtualBox_Ubuntu 22 04_23_03_2022_22_43_05

Maybe could we found the right balance: lighten a bit (less) `$system_bg_color` and darken a bit `$bg_color`:

Capture d’écran du 2022-03-23 23-02-37

vanvugt:
/me votes to not change the system background colour. It looks more professional while darker. I think you had it right in https://github.com/ubuntu/yaru/pull/3520#issuecomment-1073673601

Also [we still need to copy that into our gnome-shell patch](https://launchpad.net/bugs/1965727).

Feichtmeier:
> It looks like this if I only lighten `$system_bg_color` (I kept the lighten border-color):
> 
> Capture d’écran du 2022-03-23 22-44-05
> 
> But then the overview will looks like:
> 
> VirtualBox_Ubuntu 22 04_23_03_2022_22_43_05
> 
> Maybe could we found the right balance: lighten a bit (less) `$system_bg_color` and darken a bit `$bg_color`:
> 
> Capture d’écran du 2022-03-23 23-02-37
> 
> Imo, this is the much better solution.

I would prefer this because those huge shadows are something we usually don't do. As far as I've understood the design team they're also very careful with shadows, for example on buttons.

vanvugt:
I think I'd rather see no fix than any lightening of the background.

Also I've just committed to #222222 in https://salsa.debian.org/gnome-team/gnome-shell/-/merge_requests/59

Jupi007:
Okay, so here are the three options we have, without stronger shadow and with a lighter border:

_Click on the image to clearly see them_

| Darker bg (same as entries and dash) | Current bg | Lighter bg |
| ------------- | -------------- | -------------- |
| Capture d’écran du 2022-03-24 09-55-16 Capture d’écran du 2022-03-24 10-12-43  | Capture d’écran du 2022-03-24 09-56-57  Capture d’écran du 2022-03-24 10-11-10  | Capture d’écran du 2022-03-24 09-58-30 Capture d’écran du 2022-03-24 10-08-18 |

My favourite is the darker one, because it is consistent with dash and entries backgound, and also, popovers are all darker than background in Gtk. The problem with the lighter one, is that the bg color is too close to the one of Gtk window (and it's the same same issue as this one).
But honestly, lighten the border clearly fix 75% of the problem.

Feichtmeier:
I don't really understand why we make a design decision around a technical deadline but with seeking a compromise, I'm okay with the darker popovers with the same black shade as entries (without changing the shadow)

Jupi007:
Okay, thanks for your opinion @Feichtmeier :)

So if you agree @vanvugt, I'll go using the first proposal of my table:
- Unchanged shadow and `$system_bg_color` ;
- lighten `$border_edge` ;
- and darker `$bg_color`.

vanvugt:
Out of curiosity, what changes if the login screen used the light theme? Presumably it would solve the menu contrast problem, but would anything else change?

Jupi007:
> Out of curiosity, what changes if the login screen used the light theme? Presumably it would solve the menu contrast problem, but would anything else change?

Having the light theme on the login screen doesn't look good honestly.
The background will still be grey, but all the buttons, entries and menus will become white.
So it will solve the problem of low contrast, but then we have a problem of too high contrast.

In fact, the login screen aim to be as neutral as possible.
Using the dark theme here will works for both people who like light or dark theme.
Using the light one here maybe aggressive for the eyes at night for example.

Jupi007:
Final version:

Capture d’écran du 2022-03-24 09-55-16
Capture d’écran du 2022-03-24 10-12-43

PR #3528, opened 2022-03-21
improvements to accented icons

Muqtxdir:
- fixed the gradient in filemanager-app and software-center.
- re-rendered all accented color files to make use of newer hue values from recent PR's


Jupi007 (approved):
LGFM :+1:
Thanks @Muqtxdir

PR #3532, opened 2022-03-22
metacity: Correct titlebar contrast for default and dark themes

flexiondotorg:
This pull request corrects `metacity-theme-1.xml.in` for the `default` and `dark` themes so Metacity matches GNOME/CSD windows. Essentially this is just copying the constants over from the `mate` and `mate-dark` themes.


Jupi007:
Thanks for the fix :+1:

PR #3535, opened 2022-03-23
icons: Remove Humanity icons languishing in legacy

flexiondotorg:
Some of the legacy icon folders host old Humanity icons that name space conflict (and therefore override) existing Yaru icons


Jupi007:
Thanks @flexiondotorg

PR #3538, opened 2022-03-23
Add panel-dark icons support

Jupi007:
This generate a new `-dark` variant of the icon theme which inherit the related light theme.
The dark variant only contains dark variation of panel icons.
This dark variant is only generated if mate-dark option is enabled.

Panel icons rendered by @flexiondotorg using sed


Jupi007:
https://user-images.githubusercontent.com/36476595/159696361-1bb2e42f-bafd-402d-ad0f-7cf49aae6405.mp4

flexiondotorg:
Test on Ubuntu MATE 22.04 daily-live and works a treat. Thanks for the collaboration on this! 

However, it would be good if the panel icons are built and shipped by default so that all the Yaru variations will work correctly in Ubuntu MATE. Ubuntu MATE ships `yaru-theme-gtk` and `yaru-theme-icon` in 22.04 onwards so that all the variations are also available.

Without the panel icons, it won't work correctly.

flexiondotorg (commented):
This will create a package conflict in the .deb. 

I suggest leaving the panel and animation icons in the `yaru-theme-unity` package I can seed that package in Ubuntu MATE.

flexiondotorg:
LGTM 👍

3v1n0:
Ouch, i wanted to jump in here as this would break a refactory I'm doing... but at this point I will actually change this quite a bit.

As from now icons must all be under Yaru/ subfolder, and symlinks should be all generated.

Jupi007:
Oh, sorry. @Muqtxdir told me you was working on symlinks, but I was thinking my change doesn't affect you.

Should I do a revert commit?

3v1n0:
No, will adapt to this, don't worry.

3v1n0:
NW, it was my fault, I saw this MR but forgot to comment about :D

3v1n0:
Ok this has been refactored too in https://github.com/ubuntu/yaru/commit/1d1673e7d86b375f3a254a2c8d22e8ef4312ef06

Note that meson changes could have been already done this way :)

PR #3540, opened 2022-03-23
mate: Drop unnecessary MATE variations of GtkSourceView and Metacity themes 

flexiondotorg:
Now that GtkSourceView styles and window buttons do not have accent colours, Ubuntu MATE 22.04 is using Yaru GtkSourceView styles and Metacity themes.


Jupi007:
Thanks :)

PR #3541, opened 2022-03-23
icons: Add uLauncher panel icons

flexiondotorg:
This pull requests adds panel/indicator icons for uLauncher that match the Yaru light/dark panel icons


Jupi007:
Thanks :)

PR #3542, opened 2022-03-23
Makes Unity7 panel follow bg color

Jupi007:
As we support now dark/light panel icons, this PR makes the Unity7 panel light in light theme and dark in dark theme.

**Before:**

Capture d’écran du 2022-03-23 17-02-17

**After:**

Capture d’écran du 2022-03-23 17-03-57


Muqtxdir (approved):
looks good, thanks !

PR #3543, opened 2022-03-23
Fix broken panel icons symlinks

Jupi007:
Reduce panel icons size by using symlinks.

@3v1n0 Will this interfering with your refactoring?


3v1n0:
Yeah, I noticed they went missing... I'm restoring that part.  I don't think (light) panel changed, right? 

/cc @flexiondotorg

3v1n0:
let's get this in, I'll move (and then drop) them soon again :)

PR #3545, opened 2022-03-23
Clean up panel vars

Jupi007:
Unify Unity7 and Mate panel vars into `_colors`.

Capture d’écran du 2022-03-23 19-21-14

CC @flexiondotorg


flexiondotorg:
That is so much better! Thank you :smiley:

Feichtmeier (approved):

PR #3554, opened 2022-03-24
Re-add panel icons

### LINKED ISSUE #3550: Dark icon sets are now missing which breaks MATE and Unity panels
@3v1n0 Earlier today [support for dark panel icons was added](https://github.com/ubuntu/yaru/pull/3538) to Yaru. Sadly, [this commit](https://github.com/ubuntu/yaru/commit/1d1673e7d86b375f3a254a2c8d22e8ef4312ef06) `-dark` icon sets for accent colors themes are no longer generated, breaking panel icons for MATE and Unity 😭

  * Every accent color variant must have a `-dark` icon set with an `index.theme` **that Inherits from Yark-dark**
  * `Yaru-mate-dark` has also regressed. **It should also inherit from Yark-dark.**

Please re-instate the `-dark` icons `index.theme` for each of the accent colours and MATE along with the appropriate inheritance.

cc: @Jupi007

Jupi007:
- Move panel icons to `/src/panel` ;
- add super simple build script (just copy and minimize icons) ;
- adapt symlink script ;
- generate icons and symlinks ;
- adapt `meson.build` to install `panel` and `panel-dark` variants ;
- also add two new options to `meson_options.txt` ;
- if `panel-dark` is enable, meson generate a new `-dark` variant for each flavour ;
  - each `-dark` variant are just an `index.theme` file, which inherits to `Yaru-panel-dark` and then the related regular variant (ex. for `Yaru-mate-dark`: `Inherits=Yaru-panel-dark,Yaru-mate`) ;
- `panel` and `panel-dark` options are enable when building debs and all panel icons are handled by `yaru-theme-unity` deb.

Closes #3550

CC @flexiondotorg


flexiondotorg:
I see the mate variants are enabled by default now. 

```
+option('mate', type: 'boolean', value: true, description:'build Yaru mate flavour')
+option('mate-dark', type: 'boolean', value: true, description:'build Yaru mate-dark flavour')
```

@Jupi007 Is this intended?

flexiondotorg:
Now `panel` and `panel-dark` are disabled.

@Jupi007 I think they should be enabled by default?

Jupi007:
It is "normal" because I prefer enable as less option as possible by default.
Gnome users doesn't need this, especially because it doubles the number of icon themes.

But don't worry, it is enable when building the deb, so it can be packaged into the unity deb :)

flexiondotorg:
> But don't worry, it is enable when building the deb, so it can be packaged into the unity deb :)

Excellent!

flexiondotorg:
Built and tested on Ubuntu MATE 22.04 daily-live. 

I can confirm that:

- The new icon inheritance is working in MATE Desktop 🥳
- The dark panel icons are restored in all the right places 💖
- And light themes continue to work as expected 🕺

Great work. Thank you for your hard work 😅 Gets a :+1: from me!

Jupi007:
Closed, as @3v1n0 fixed this in master.

PR #3563, opened 2022-03-25
icons: Add symlinks for fullcolor printer and printers icons

flexiondotorg:
The pull request adds symlinks for fullcolour `printer` and `printers` icons used by `system-config-printer`. 

`system-config-printer.desktop` is hidden from GNOME and KDE so this fix is for other desktop environments, such as MATE.


Jupi007 (approved):
Lgfm, thanks @flexiondotorg

PR #3565, opened 2022-03-26
add devices/speakers-* to all accented variants

Muqtxdir:


3v1n0:
Oh, I thought that would have implied using some magic to get the color done as it happens with symbolic icons that define special paths that can be colored on errors... But adding more svgs that needs to be manually changed all the times it's just too annoying to maintain.

An option we could use is to instead do the replacement ourself and use the SVGs as they are, by using an `accented-scalable` icons maybe. The problem is how to get those icons to be moved to the proper folder...

Muqtxdir:
yeah i used a script to create `devices` dir in all variants and replaced the _magic color_ with respective colors from `common` in for-loop, since these actually are specific to `gnome-control-center` I didn't add another option in the `colorize-dummy-svg.py` as this is too niche

Jupi007:
Can't we just create a source file inside '/src' for those icons?
Even if it is partial.

3v1n0:
Yeah, this would be my idea: see the rendering path in meson.

So basically the idea is to generate such icons in the build directory and eventually install them, can basically do all at configuration level and use `install_data` + rename to move them in the right spot.

flexiondotorg:
@Muqtxdir Any change coloured version for MATE can be included too? 🧉

3v1n0:
Let's include this for now, I will generate them on an upcoming refactor...

3v1n0:
@flexiondotorg mate accente added via 028fd6e5c6c75b74af2205d07a57b56cc946c712

PR #3569, opened 2022-03-28
icons: Add redshift and steam panel icons

flexiondotorg:
This pull requests adds correctly coloured panel icons for Steam and RedShift. Also adds a symlink for MATEs Preferred Applications control center.


3v1n0:
One day it would be nice to have such kinds of icons to be all generated for the dark variant, but as for now I think it's good to go! :)

Thanks

flexiondotorg:
> One day it would be nice to have such kinds of icons to be all generated for the dark variant, but as for now I think it's good to go! :)

Good idea. I've create an issue so we don't forget and include the snippet to do it :+1: 

#3570

PR #3595, opened 2022-04-06
Fix notifications shadow (revert to upstream look)

Jupi007:
Fix notifications shadow (revert to upstream look).

Capture d’écran du 2022-04-06 23-26-26

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_notifications.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_notifications.scss
@@ -31,7 +31,7 @@ $notification_banner_radius: $base_border_radius*1.5;
   
   // Yaru: make notifications more visible on different headerbars:
   &, &:hover, &:focus, &:active {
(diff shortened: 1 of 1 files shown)


Feichtmeier (approved):
very nice - eventually double check if light still looks good? 🤔

Jupi007:
> eventually double check if light still looks good? thinking

Indeed, I didn't think about it 😅
Thanks for the reminder!

This seems looks good on light theme:

Capture d’écran du 2022-04-06 23-36-14

Jupi007:
Wait, it is a bit strange with a light background:

Capture d’écran du 2022-04-06 23-41-00

It looks imo better to move up the shadow (so it is centered):

Capture d’écran du 2022-04-06 23-45-21

Wdyt?

Feichtmeier:
the border is great. I zoomed like mad, but whats the dif between those two pics?

Jupi007:
It's a very soft change :D

On the second pic, the shadow is 2px upper, so the bottom part of the shadow is less big.

PR #3596, opened 2022-04-07
Add ubiquity 22x22 size

Jupi007:
- Add ubiquity 22x22 size ;
- also fix the shadow at 24x24 size.


Muqtxdir (approved):
looks good at 22x22, thank!s

PR #3624, opened 2022-04-30
Sync with Libadwaita symbolics 3614

Jupi007:
Symlink new `adw-entry-apply-symbolic` with `object-select-symbolic`.

Related to #3614


Muqtxdir (approved):
thank!s @Jupi007

PR #3671, opened 2022-08-11
Sync icon theme with upstream - 3670

Jupi007:
**New icons:**

Capture d’écran du 2022-08-11 11-40-12
Capture d’écran du 2022-08-11 11-36-05

Related to #3670


Muqtxdir (approved):
Thanks again

Jupi007:
The test failed, but it is just the icon generation automation which is broken, @3v1n0 can you take a look at this, please?

PR #3672, opened 2022-08-11
Sync GS with upstream - 3665

Jupi007:
**New icon:**

Capture d’écran du 2022-08-11 11-46-57

Related to #3665

### DIFF
--- a/gnome-shell/src/data/icons/scalable/actions/dark-mode-symbolic.svg
+++ b/gnome-shell/src/data/icons/scalable/actions/dark-mode-symbolic.svg
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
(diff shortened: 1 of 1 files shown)


Muqtxdir:
@Jupi007 wdyt about this type of symbolic for `dark-mode`?

dark-symbolic

_it's similiar to the one used as quick-toggle in iOS, but I feel that this might look complicated?_

Jupi007:
Hmm, that would have been a good idea. But I'm afraid it looks messy at 16x16 in the panel 🤔

Muqtxdir:
> Hmm, that would have been a good idea. But I'm afraid it looks messy at 16x16 in the panel 🤔

yep, your initial symbolic would look better in panel 👍

Muqtxdir (approved):

Jupi007:
I just Yaru-ified the upstream one :D

PR #3674, opened 2022-08-11
Icon theme fixed shadow

Jupi007:
**Before:**

Capture d’écran du 2022-08-11 12-20-26

**After:**

Capture d’écran du 2022-08-11 13-19-23

Closes #3668

PR #3679, opened 2022-08-15
Add gnome-text-editor icons

Muqtxdir:
afaik, ubuntu 22.10 has replaced the pre-installed `gedit` application with `gnome-text-editor` and moreover it is a `GNOME-Core App`. So, this PR adds icons for it


Jupi007 (changes requested):
Thanks for working on this @Muqtxdir :)
The icon is nice!

I just have a few notes:

- Our pencil is always blue. Did you tried to use it, with another color for the currently blue top bar?
  _This is just an idea, and maybe using your current pencil will be better :)_
- I think the blue top bar shadow is a bit too height, I tried and I think using the same gradient but with 8px height looks less prominent.
- I see you used one shadow layer which is bugged in GS (see #3674), can you replace it with a full black one with 0.1 opacity.

And I think that's it.
I also ping @ubuntujaggers here, as he's working on rework small size icons.

Muqtxdir:
- [x] make the top bar shadow 8px from 16px
- [x] Replaced the _bulging shadow_ layer with _0.1_ opacity layer

_Note: The Icon is kept closer to the upstream Icon, thus the pencil has a yellowish-yaru shade similiar to the Upstream's pencil
TLDR; just yarufied upstream icon :]_

Jupi007 (changes requested):
Hmm, sorry @Muqtxdir but I think you dropped the shadow blur (20% in Inkscape).
The bugged shadow layer just needs to not be a gradient. It still needs to have blur.
All the rest is okay :)

Muqtxdir:
sorry for the late @Jupi007 i did not get notification for your earlier comment. just added 20% blur

Jupi007 (approved):
Perfect, thanks @Muqtxdir :)

Muqtxdir:
merging this, ci fails for symlink issue that is not targeted in this pr

Jupi007:
Sorry @Muqtxdir, I waited so long for the end of the CI that I forgotten to merge 😅

Muqtxdir:
no problem, thanks for review as always

PR #3689, opened 2022-08-22
gitattributes: Exclude .github from tarball releases

jbicha:
The Debian package excludes this folder but I think it makes sense to do it here instead (so that tarball releases don't have this folder).


Jupi007 (approved):
LGFM Thanks @jbicha :)

PR #3699, opened 2022-08-24
gtk: Optimize accent color definition taking background in account

3v1n0:
Compute accent-color value (the one used by links, so far) to have a
contrast value to be optimized according to W3C requirements.

See: https://twitter.com/alexm_gnome/status/1526531918166212608

### DIFF
--- a/gtk/src/default/gtk-3.0/_colors.scss
+++ b/gtk/src/default/gtk-3.0/_colors.scss
@@ -88,3 +88,8 @@ $coloured_focus_border_color: transparentize(white, 0.3);
 // Yaru: Ignore transformations when using dark accent colors
 $progress_bg_color: $accent_bg_color;
 $checkradio_bg_color: $accent_bg_color;
+
+// Optimize accent-color definition for default background colors
+@import 'sass-utils';
+$link_color: optimize-contrast($bg_color, $accent_bg_color);
+$link_visited_color: if($variant =='light', darken($link_color, 10%), lighten($link_color, 10%));
--- a/common/sass-utils.scss
+++ b/common/sass-utils.scss
@@ -92,13 +92,20 @@
     @return null;
 }
 
-@function pow($base, $exponent) {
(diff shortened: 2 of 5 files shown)


3v1n0:
To be clear, before:

image


After:

image

gtk/src/default/gtk-3.0/_colors.scss
```diff
@@ -88,3 +88,9 @@ $coloured_focus_border_color: transparentize(white, 0.3);
 // Yaru: Ignore transformations when using dark accent colors
 $progress_bg_color: $accent_bg_color;
 $checkradio_bg_color: $accent_bg_color;
+
+// Optimize accent-color definition for default background colors
+@import 'sass-utils';
+$yaru-accent-color: optimize-contrast($bg_color, $accent_bg_color);
+$link_color: if($variant =='light', $yaru-accent-color, $yaru-accent-color);
```
Jupi007, line 95: I think the conditional function is not needed here (same for gtk4).

3v1n0: indeed... -_-

gtk/src/default/gtk-3.0/_colors.scss
```diff
@@ -88,3 +88,8 @@ $coloured_focus_border_color: transparentize(white, 0.3);
 // Yaru: Ignore transformations when using dark accent colors
 $progress_bg_color: $accent_bg_color;
 $checkradio_bg_color: $accent_bg_color;
+
+// Optimize accent-color definition for default background colors
+@import 'sass-utils';
+$yaru-accent-color: optimize-contrast($bg_color, $accent_bg_color);
```
Jupi007, line 94: I think this var should be named $link-color because actually this one isn't used at all :D

3v1n0: defintely, bad copy and paste...

Jupi007 (approved):

PR #3700, opened 2022-08-26
GS43: minor tweaks to assets

Muqtxdir:
### dark-symbolic:

- reduced the size from 16 to 14 as it looks bigger when compared to other symbolics used in system-tray

 | current  | pr |
 | ------------- | ------------- |
 | Screenshot from 2022-08-27 03-08-37 | Screenshot from 2022-08-27 03-03-18 |

### checkboxes:
- updated the assets for checkboxes to keep it consistent with the ones offered by libadwaita
- _Note: Maybe the check-symbolic in gtk3, gtk4 also need to be changed for this?_


Jupi007:
I approve everything except the checkmark :D

IMO we should keep it, because it is part of the Yaru feeling.
I think we should prefer to change the asset in Libadwaita.
It would be a very minor change (just two assets).

CC @3v1n0

Muqtxdir:
> IMO we should keep it, because it is part of the Yaru feeling.
I think we should prefer to change the asset in Libadwaita.
It would be a very minor change (just two assets).

i feel the same way, _those precious yaru checkmarks_ If @3v1n0 could add them into the libadwaita patch and mentions it here, I'll immediately change it

Muqtxdir:
@Jupi007 I have reverted the check back to the yaru one.

Jupi007 (approved):
Fine, that LGFM :)

PR #3703, opened 2022-08-27
add styles for gnome text-editor

Muqtxdir:
### changes:

- make use of new `metadata` properties to correctly declare `light` and `dark` variants of `gtksourceview-5` styles.
- split `src` into `gtksourceview` (_style that is compatible with gtksourcview-2.0, gtksourcview-3.0 and gtksourcview-4_) and `gtksourceview-5` rather than only providing for `gtksourceview-5`

### other:
@jbicha could **gnome text-editor's** default style in `ubuntu` be changed to `Yaru` from `Adwaita` after this pr?


3v1n0:
I suppose we've to change also how this is managed by gcc / shell or is it not needed?

Muqtxdir:
not needed, since it follows system-preference as seen below in vide

https://user-images.githubusercontent.com/54065734/187039185-3451f554-f6af-4a57-b040-5199df6cd453.mp4

Muqtxdir:
@3v1n0 what I mean is, it is not needed to manage to switch for newer gtk4+libadwaita apps via gcc but for the older apps such as gedit it is needed.

Jupi007 (approved):
Tested LGFM :+1:

Muqtxdir:
Thanks!

PR #3705, opened 2022-08-27
Realign edge resize cursors

### LINKED ISSUE #3696: Edge resize cursor targets seem misaligned
When resizing windows or regions that use the edge resize cursors (`top_side`, `bottom_side`, `left_side`, `right_side` in **icons/src/cursors**), the mouse target—canonically positioned at the very tip of the default cursor—is positioned on the "outer" triangle rather than between the triangles, as the cursor's symmetry might suggest. This can make it difficult to position the cursor accurately, especially for small resize handles.

**OS/Desktop/Yaru version**

- Ubuntu 22.04.1 LTS (upgraded from Ubuntu 20.04)
- gnome-shell 42.2-0ubuntu0.2 (X11 Session)
- xorg 1:7.7+23ubuntu2
- yaru-theme-icon 22.04.4

**Screenshots**

The cursor target for Yaru's edge cursors is where the "outer" triangle is, but one might expect the target to be centered between the triangles. This makes it seem like the cursor gets "pulled in" towards the inside of the window/object being resized. In the screenshots below, the cursor target is centered as precisely as I could get on the window border for various programs (Gedit, GNOME Terminal, Visual Studio Code).

_Note: VS Code has its own custom window decoration that does not match exactly with the X11 window edge, which further exacerbates the "misaligned" appearance._

Gedit (Yaru left edge resize)
GNOME Terminal (Yaru top edge resize)
VS Code (right edge resize; Yaru)

Here is an example of the "expected" positioning demonstrated in Nautilus with the horizontal resize cursor, which has a similar symmetrical appearance:

Nautilus (Yaru horizontal resize)

The cursor's target position near the "outer edge" of the cursor image makes sense based on how the DMZ-Black edge resize cursors look. Here is an example with Evolution:

Evolution (DMZ-Black top edge resize)

Jupi007:
**Before:**

Capture d’écran du 2022-08-27 17-50-11

**After:**

image

Closes #3696


Muqtxdir:
@Jupi007 how are you testing it (_i remember their being a test-html, can't seem to remember where it is?_)

Jupi007:
In the cursor folder, there is a test one with a html file.
But you need to use a non-snap browser, I used Firefox nightly. Gnome web doesn't work because it wants to download the file instead of display it.

Muqtxdir:
got it, thanks for the browser-tip!

Muqtxdir (approved):
thanks for the fix, i assume this pr should be also considered for backport ?

Jupi007:
Indeed, it require a backport to 22.04 :)

PR #3758, opened 2022-10-12
Reducing edge highlight on full colour app icons in sizes 16-48px

ubuntujaggers:
Sorry this took so long!  I probably haven't caught all of them, but here are most (I hope) of the buttony-looking app icons.


Feichtmeier (approved):
Checked this out locally and this looks better in all sizes! Thank you @ubuntujaggers and sorry that it took so long to finally test this

PR #3775, opened 2022-11-23
Sync icon theme with upstream

Jupi007:
### New icons:

**pointing-symbolic**

Capture d’écran du 2022-11-23 14-06-45

**hearing-symbolic**

Capture d’écran du 2022-11-23 14-06-18

Related to #3774


Muqtxdir (approved):
this are from the new icons from accessibility panel redesign, right?

Jupi007:
Yes, it is from from the accessibility panel redesign.

PR #3784, opened 2022-11-30
Sync gnome-shell with upstream

Jupi007:
We can't test this because the next Shell version is still not available in Ubuntu.
So this is a "dirty" sync, and I will surly need to resync the theme from a fresh copy.


Feichtmeier:
then let's wait I would say

Jupi007:
Well, yes and no: merge this provide at least a working state for the next version, while do not mean that some stuff will be broken.

Jupi007:
Just installed a VM to test this branch, and I didn't noticed any problem.
In fact, even if the diff is very large, there is no design change, just a big refactoring/rework.

VirtualBox_ubuntu 22 10_04_12_2022_15_22_33
VirtualBox_ubuntu 22 10_04_12_2022_15_21_52
VirtualBox_ubuntu 22 10_04_12_2022_15_21_26
VirtualBox_ubuntu 22 10_04_12_2022_15_21_18
VirtualBox_ubuntu 22 10_04_12_2022_15_17_40
VirtualBox_ubuntu 22 10_04_12_2022_15_52_35

So I think it is safe to merge this 🙂

Feichtmeier:
ok :+1:

PR #3795, opened 2022-12-04
Rework list-remove icon

### LINKED ISSUE #3690: list-remove icon looks like a "no entry" (traffic) sign
The list-remove icon looks like a "no entry" (traffic) sign, which conveys a completely different meaning than what is intended (and thus it is confusing).

E.g.: https://github.com/ubuntu/yaru/blob/master/icons/Yaru/48x48/legacy/list-remove.png

Vs.: https://en.wikipedia.org/wiki/Prohibitory_traffic_sign#No_entry

This affects applications like 'gramps' or 'transmission-remote-gtk' for example. The other icon themes I tested don’t have this issue.

Using yaru-icon-theme 22.04.4 on Ubuntu 22.04.1, and whatever versions are current in Ubuntu for the affected applications.

I suppose that going forward it would be best for applications to switch to using list-remove-symbolic instead (at least in some cases?), but that change won't happen for any applications in the repositories right now, so a fix for the theme would be nice…

Jupi007:
**Before:**

image

**After:**

image

CC @janc13

Fixes #3690


ubuntujaggers:
Looks much better! 👍

Jupi007:
Thanks 😊

Muqtxdir (approved):
looks great!

PR #3812, opened 2023-01-05
Sync gtk4 with upstream #3803

Jupi007:
They just added new feature, so no compatibility problem.

Related to #3803


Feichtmeier (approved):
Lol! Finally the undershot again xD wb old friend

PR #3830, opened 2023-02-28
Timeshift icons overridden with backups icon

### LINKED ISSUE #3816: Timeshift icons overridden with backups icon
**Expected Behavior**

I would expect the Timeshift application to be represented by its [original icons](https://github.com/linuxmint/timeshift/tree/master/icons).

**Actual Behavior**

The same icon is used as for deja-dup. This is confusing for someone who uses both.

1. `sudo apt install timeshift`
2. click on "Show Applications" in the bottom-left corner
3. Confirm Timeshift icon is a safe (same as for deja-dup)

Freshly installed Ubuntu Jammy (22.04.01).

**Yaru version**

20.04.4

Jupi007:
@Muqtxdir Timeshift isn't a pre-installed app, and I agree with the linked issue comment that it can be confusing with Déjà-dup, WDYT?

Fixes #3816
Revert #3477

### DIFF
--- a/icons/Yaru/16x16/apps/timeshift.png
+++ b/icons/Yaru/16x16/apps/timeshift.png
@@ -1,1 +1,0 @@
-backups-app.png
--- a/icons/Yaru/16x16@2x/apps/timeshift.png
+++ b/icons/Yaru/16x16@2x/apps/timeshift.png
@@ -1,1 +1,0 @@
-backups-app.png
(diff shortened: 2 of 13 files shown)


Muqtxdir:
It makes sense to revert since timeshift is not even shipped on ubuntu-desktop. I don't remember why I symlinked it in the first place 😅

Jupi007:
> It makes sense to revert since timeshift is not even shipped on ubuntu-desktop. I don't remember why I symlinked it in the first place sweat_smile

Because somebody complained about a missing icon 😅

PR #3834, opened 2023-02-28
Sync Gtk4 with upstream

Jupi007:
Related to #3815


Feichtmeier (approved):
Thanks! I am glad I still understand SCSS :D

PR #3835, opened 2023-02-28
Suggested button green only in default variant

### LINKED ISSUE #3610: The default action button in dialogs is always green, regardless of the accent color
Currently, the default action button in dialogs is always green, regardless of the accent color. IMO, it would be great if this button was colored in the accent color.

Screenshot from 2022-04-18 09-23-27

Screenshot from 2022-04-18 09-23-43

Screenshot from 2022-04-18 09-23-55

Jupi007:
Capture d’écran du 2023-02-28 17-16-09
Capture d’écran du 2023-02-28 17-16-28

It is consistent with what we did in yaru.dart.

**Note:** `$suggested_bg_color` var is only used for suggested button

Closes #3610


Feichtmeier (approved):
Absolutely correct! Was it green everywhere? :see_no_evil:

Jupi007:
> Absolutely correct! Was it green everywhere? :see_no_evil: 

Yes 😄

PR #3855, opened 2023-03-17
Fix dash separator color in light theme

Jupi007:
The dash separator was white in the light theme.
This PR use the dark variant of the border colour.

| Before | After 
| --- | ---
| Capture d’écran du 2023-03-17 12-35-58 | Capture d’écran du 2023-03-17 12-34-51


Feichtmeier (approved):
! much better :D

PR #3861, opened 2023-03-20
Sync icon theme with upstream

Jupi007:
I added large variant of pan icons, then symlinked one to the new simple scan icon.

Capture d’écran du 2023-03-20 13-53-31

Reference: https://github.com/GNOME/simple-scan/blob/37caec48932ae1f91b166ddf94c44c6e9066e431/data/icons/scalable/actions/right-large-symbolic.svg

Related to #3858


3v1n0:
LGTM (for what it worth from the graphical side :))

Muqtxdir (approved):
sorry for the late

PR #3869, opened 2023-03-25
Gtk4: use lighter $menu_selected_color in dark theme

Jupi007:
| Before | After
| --- | ---
| Capture d’écran du 2023-03-25 21-07-02 | Capture d’écran du 2023-03-25 21-09-06

CC @3v1n0


Feichtmeier (approved):
PUhh,, abs. much better
How comes this is so different upstream between normal gtk4 and libadadwaita?

Jupi007:
>  How comes this is so different upstream between normal gtk4 and libadadwaita?

Honestly I have no idea 😅

3v1n0:
Oh, fair enough, let's get in!

PR #3877, opened 2023-03-28
Add missing folds on libreoffice-oasis-formula icon

Jupi007:
I just noticed that I had hidden the `folds` layer in Inkscape... 😥
See the github rich diff to see the difference.

@3v1n0 Is it still possible to get this in?


Muqtxdir (approved):

3v1n0:
@Jupi007 I think the change is minimal enough we can just do it in next upload.

PR #3884, opened 2023-04-03
Fix default decoration windowcontrols sizes

3v1n0:
Helps with rendering of #3876

### DIFF
--- a/gtk/src/default/gtk-4.0/_tweaks.scss
+++ b/gtk/src/default/gtk-4.0/_tweaks.scss
@@ -79,7 +79,7 @@ windowcontrols {
         $_base_button_color: transparentize($fg_color, 0.9);
         $_base_hover_color: transparentize($fg_color, 0.85);
         $_base_active_color: transparentize($fg_color, 0.75);
-        
+
         padding: 3px;
         margin: 0 4px;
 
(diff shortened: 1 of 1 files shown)


Jupi007:
@3v1n0 isn't this bug affecting Gtk3?

3v1n0:
I'm not aware of any application using these kind of decorations in gtk3,
as client-side should not use this, while server-side are all gtk4
nowadays.

Il lun 3 apr 2023, 22:04 Paul Kepinski ***@***.***> ha
scritto:

Jupi007 (approved):
Okay, I didn't understood that now SSD decorations are using Gtk4.
This is the reason why this problem dates from 23.04

This is fine so 👍

Muqtxdir:
I think software properties and updater also have this issue on xorg session, so it kinda affects gtk3 as well

Jupi007:
> I think software properties and updater also have this issue on xorg session, so it kinda affects gtk3 as well

This is why I was also thinking that it is affecting Gtk3 🤔

3v1n0:
> I think software properties and updater also have this issue on xorg session, so it kinda affects gtk3 as well

Yes, but those are all server-side decorations, and using gtk4 now. In fact, when running under wayland with CSD, they work fine.

3v1n0:
You can see the differences here (this is under wayland):

Server-side in front, client-side (gtk4) in the back:
image

The other way around:
image

The one in the front is csd, gtk4
image


I've changed the sizes to be consistent, however... In headerbar the window buttons are 24px, while in the csd/serverside normal decorations we use 22px buttons.

I think this is acceptable because the size of the container is also different, but I'm not sure we want to fix this too (in another MR though).

PR #3956, opened 2023-07-31
Fix panel icons size

Jupi007:
Fixes #3945

PR #3957, opened 2023-07-31
Fix Yaru panel icons size

Jupi007:
Backport of #3956 for jammy


Muqtxdir (approved):

Muqtxdir:
should this just be merged or require okay from desktop-team as well?

Jupi007:
We can merge it, but the upload will be done only if we do a SRU request.
That's why I asked @richcarni to create a Launchpad bug.

PR #3962, opened 2023-07-31
Sync Gtk4 with upstream

Jupi007:
Related to #3943


Feichtmeier (approved):
Interesting 🤔 
So the backdrop changes now only apply to the headerbar, right?

Jupi007:
> So the backdrop changes now only apply to the headerbar, right?

I think so: https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6175
The goal is to only change the headerbar and buttons on backdrop state.

PR #3969, opened 2023-08-23
gnome-shell: Add keyboard-brightness-off-symbolic implementation

3v1n0:
This is an initial version, made by a non-designer at least to fill the void.

Related to #3966


3v1n0:
Handled by #3970.

PR #3970, opened 2023-08-23
GS: fix keyboard brightness icons

Jupi007:
- centre keyboard brightness icons ;
- add missing `keyboard-brightness-off-symbolic` icon.

Capture d’écran du 2023-08-23 17-07-12

Replace #3969
Related to #3966


3v1n0:
Looks nice, I kinda of liked the padding between the "off" slash and the keystroke of upstream version, is something we should keep?

gnome-shell/src/data/gnome-shell-icons.gresource.xml
```diff
@@ -25,6 +25,7 @@
     <file>scalable/status/keyboard-brightness-high-symbolic.svg</file>
     <file>scalable/status/keyboard-brightness-low-symbolic.svg</file>
```
3v1n0, line 26: This can be removed from gresources as per #3966

Jupi007:
Well I added one. We also do this in Yaru icons, the removed space is up to the slash.
But on this icon, there was only a very small part removed.

Jupi007:
I updated the "off" icon to be consistent with our other "off" icon (like sound volume):

image

3v1n0:
I.e. something like this is too much?

image

Jupi007:
Well the erased part is always up to the slash:

image

So the maximum I can do is this:

image

3v1n0:
Ack, that's better.

3v1n0:
Thanks!

PR #3974, opened 2023-08-28
gnome-shell: Update to GNOME 45 and reduce some upstream delta

3v1n0:
A part from merging with upstream, and removing some redundant changes, there are fixes for the high-contrast theme (using dark by default again, as it's what upstream tests too), the main visual changes are the fact that now quick-list use upstream theming, but colors are fixed to be contrast friendly, so the result is nicer compared to the washed-out colors that we had which had some contrast issues too (and introduced too much delta in code).

Schermata del 2023-08-28 22-29-20
Schermata del 2023-08-28 21-51-02
Schermata del 2023-08-28 21-51-09

Includes #3879


Feichtmeier:
Not sure about this one. Can't we keep the tinted colours for not high contrast themes? Full orange on a big surface is always very problematic. Especially with white on it.

3v1n0:
The contrast ratio in the HighContrast case is 5.51 (cfr https://www.siegemedia.com/contrast-ratio#white-on-%23bb3e12) while in normal case is 4.5 (https://www.siegemedia.com/contrast-ratio#white-on-%23d34615) so in both cases at levels that [w3c](https://www.w3.org/TR/WCAG/#contrast-minimum) suggests.

We can even increase that value, but in general I'd prefer to keep us closer to the original design and reduce the theme delta, if we can preserve usability.

Feichtmeier:
While w3c is a very nice standard the Ubuntu orange is really an exception. Because it has such a high luminance, eventually one of the highest without being white.
It is very nice iconic when used as little gems or stripes, on a big surface it is really not good. We've been there many times. Taking the Delta in this case is a good exchange in this case. There are other areas where we could reduce Delta as for example in the panel height

Jupi007:
I also really don't like the full tinted surface... It is way too aggressive.

3v1n0:
Ok, I've added a further commit to use `blue-dark` as the default high contrast variant, so we get more consistent behavior.

Not sure about having really more contrast, but that's how it looks:

Schermata del 2023-08-28 23-19-23

3v1n0:
> I also really don't like the full tinted surface... It is way too aggressive.

Well, I can understand that, but that's what the upstream design wanted... Indeed with blue it's probably less aggressive, but also with our accents we ended up having washed out colors, which imho they didn't look very nice.

However, let's go with this for now in mantic so that we can unblock proposed, then feel free to reiterate to improve the initial version! :)

Jupi007:
> However, let's go with this for now in mantic so that we can unblock proposed, then feel free to reiterate to improve the initial version! :)

Ofc, we'll think about this :)

3v1n0:
Merged as @Jupi007 suggested in private, so that he can address the various concerns in other PRs.

3v1n0:
ah, for reference... Currently libadwaita apps in ubuntu already use the same logic for suggested-action buttons (as these toggle buttons are considered by the shell).

So, we could tune them all, but once the contrast level is set to an acceptable value, it seems it worked well enough for now.

Feichtmeier:
suggested action is not  the same thing from a design perspective or semantically

If those quick toggle buttons can be compared to anything then probably to the android quick toggles 
however material3, even though maybe a bit controversy or polarizing, manages to tint extreme colors down until they are convenient for the eyes, then uses either blackish or whitish foreground color on top
as long as we don't have something like this in gnome I would suggest to continue to tone down the accent colors for the system drop downs in gnome shell

PR #3989, opened 2023-09-06
Gnome Shell upstream sync

Jupi007:
This fixes the focus ring color:

image
image

It also adds a border to the lock screen input (I'm not fan):

Capture d’écran du 2023-09-06 10-01-28

Related to #3975


Feichtmeier (approved):
Ty! Also not a fan of the login entry border :)

Jupi007:
So I removed it:

image

Feichtmeier (approved):

PR #4009, opened 2023-10-09
Fix looking glass extensions cards

Jupi007:
**Before:**

Capture d’écran du 2023-10-09 21-22-40

**After:**

Capture d’écran du 2023-10-09 21-20-11


3v1n0:
Don't remember if we should also handle some states such as insensitive?

Jupi007:
There was no interactivity on those card. That's why I fixed it like that, because it was the simpler solution.

3v1n0:
ack, fair enough, thanks for the quick fix!

PR #4031, opened 2024-01-11
gtksourceview: Support gedit's forked library

jbicha:
Test Case
-------
- Install Ubuntu 24.04 LTS. Choose the dark style option.
- After the install, install gedit (as a .deb)
- Run gedit

What Happens
---------------------
gedit's text area has major readable issues. When run from the command-line, this warning appears:

`Style scheme 'Yaru-dark' cannot be found, falling back to 'tango' default style scheme.`

Other Info
---------------
The gedit maintainer has forked gtksourceview to a new libgedit-gtksourceview library, ignoring advice from several that it was not ideal. Ubuntu 24.04 LTS has gedit 46 where that change has been made.

### DIFF
--- a/gtksourceview/gtksourceview/meson.build
+++ b/gtksourceview/gtksourceview/meson.build
@@ -8,9 +8,11 @@ foreach flavour: flavours
       'FlavourThemeName': theme_name,
     })
 
-  foreach version: ['2.0', '3.0', '4']
-    gtksourceview_dir = 'gtksourceview-' + version
-    style_dir = join_paths(get_option('datadir'), gtksourceview_dir, 'styles')
+  foreach version: ['gtksourceview-2.0',
+                    'gtksourceview-3.0',
+                    'gtksourceview-4',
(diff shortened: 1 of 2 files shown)


gtksourceview/gtksourceview/meson.build
```diff
@@ -8,9 +8,11 @@ foreach flavour: flavours
...
     })
 
-  foreach version: ['2.0', '3.0', '4']
-    gtksourceview_dir = 'gtksourceview-' + version
-    style_dir = join_paths(get_option('datadir'), gtksourceview_dir, 'styles')
+  foreach version: ['gtksourceview-2.0',
+                    'gtksourceview-3.0',
+                    'gtksourceview-4',
+                    'libgedit-gtksourceview-300']
+    style_dir = join_paths(get_option('datadir'), version, 'styles')
```
3v1n0, line 15: nit:

```suggestion
    style_dir = get_option('datadir') / version / 'styles'
```

PR #4050, opened 2024-02-29
gnome-shell: Don't give login-dialog another background-color

vanvugt:
Its parent `lockDialogGroup` has exactly the same background-color set in the same file. Giving them both the same wasted render time on overdraw, and caused multi-layer blending artifacts (slight flickering of the grey background) when the login dialog fades in/out.

While the flicker can also be fixed using `set_offscreen_redirect`, there's no point adding that overhead when the extra layer doesn't need to be painted. Removing it halves the average render time of the login animation.

Closes: #4048
Copied-from: https://gitlab.gnome.org/GNOME/gnome-shell/-/commit/d09509b2cbce

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
@@ -14,7 +14,6 @@ $_gdm_dialog_width: 23em;
 
 /* Login Dialog */
 .login-dialog {
-  background-color: $_gdm_bg;
 }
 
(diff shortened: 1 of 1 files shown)


Feichtmeier:
Thanks @vanvugt 
The current yaru gnome shell theme targets gnome shell 45.XX
So this code would apply to this version, this is correct?
As we didnt start looking into the new shell for 24.04 here yet

vanvugt:
I think this change will work for all recent gnome-shell versions.

PR #4051, opened 2024-03-01
Update Yaru to new Dash-to-Dock code

sergio-costas:
Some changes were required in the scss to make Dash to Dock compatible with Gnome Shell 46, while also keeping compatibility with Gnome Shell 45. These are those changes. This is the PR in DashToDock: https://github.com/micheleg/dash-to-dock/pull/2146

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_dock.scss
+++ b/gnome-shell/src/gnome-shell-sass/_dock.scss
@@ -15,6 +15,8 @@ $dock_edge_items_padding: $dash_padding - $dock_spacing;
 $dock_fixed_inner_margin: $dock_side_margin;
 $dock_icons_distance: 4px;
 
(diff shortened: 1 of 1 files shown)


sergio-costas:
@Feichtmeier BTW: I found another difference between the current scss and the one in dash-to-dock:

Captura desde 2024-03-01 13-12-00

Captura desde 2024-03-01 13-12-09

I can do a MR if you consider that it is important...

Feichtmeier:
> @Feichtmeier BTW: I found another difference between the current scss and the one in dash-to-dock:
> 
> Captura desde 2024-03-01 13-12-00
> 
> Captura desde 2024-03-01 13-12-09
> 
> I can do a MR if you consider that it is important...

Sounds like it is an addition so I think yes please

PR #4052, opened 2024-03-01
Add extra changes from DashToDock

sergio-costas:
There was a difference between the scss file from DashToDock and the current one in Yaru. This PR adds it.


sergio-costas:
@Feichtmeier Here it is.

Feichtmeier (approved):
Thank you!

PR #4056, opened 2024-03-18
Fix apps view

sergio-costas:
The apps view shows an extra rectangle in the icons when they are hovered by the cursor.

This patch fixes it (extracted from patch for gnome 46 for Dash to Dock).


Feichtmeier (approved):
ty

PR #4072, opened 2024-04-17
Start here / app-grid yaru fixes

3v1n0:
Fix symlinks and avoid confusions adding two different kinds.

So the problem here was that we had start-here icon that was symlinked to the app-grid one, thus when optimizing the first one we were overwriting the app-grid one. This is now fixed and I've added different variant of them.

So that it's still possible to use the default grid icon in ubuntu, but the dock will continue using https://github.com/micheleg/dash-to-dock/pull/2197 the circle of friends by default, not to have to do an UIFe. But this can overridden by other distros and only affect installations with the dock running the ubuntu session.


3v1n0:
Description updated

Feichtmeier (approved):
As you've described the problem I approve this change, hoping that pure gnome fans won't be disappointed:)

3v1n0:
> hoping that pure gnome fans won't be disappointed:)

Well, they wouldn't probably use the theme anyways if they're purists :), but in case they hopefully do it, it won't be there in the normal gnome session or yaru one.

PR #4090, opened 2024-05-31
gnome-shell: Webauth qrcode light theme fixes

3v1n0:
Before:

immagine

After:

image

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
@@ -224,32 +224,32 @@ $_gdm_dialog_width: 25em;
 }
 
 .web-login-spinner {
-  background-color: rgba(0, 0, 0, 0.5);
-  border: 5px rgba(0, 0, 0, 0.0);
+  color: $osd_fg_color;
+  background-color: transparentize($osd_bg_color, 0.5);
+  border: 5px transparent;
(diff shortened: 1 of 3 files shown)


Muqtxdir:
hi @3v1n0 is this issue also happening with upstream's light theme?

gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
```diff
@@ -224,32 +223,32 @@ $_gdm_dialog_width: 25em;
 }
 
 .web-login-spinner {
-  background-color: rgba(0, 0, 0, 0.5);
-  border: 5px rgba(0, 0, 0, 0.0);
+  color: $osd_fg_color;
```
Feichtmeier, line 226: Looks good except this I'm not sure. Did this change to change it's value for light/dark ?
Back in the days it was always white

3v1n0: Yes, it's always white and it has to be white here because in both cases it's an osd-like spinner:

image

And when the dialog is white it has still to be white on gray.

3v1n0:
> hi @3v1n0 is this issue also happening with upstream's light theme?

It's happening in a PR we're working on, and I'll backport these changes there too.

Feichtmeier (approved):

PR #4157, opened 2024-11-09
Added redesigned org.gnome.Software icon

### LINKED ISSUE #4156: My Initial Take on Gnome Software Icon
### Discussed in https://github.com/ubuntu/yaru/discussions/4155

<div type='discussions-op-text'>

<sup>Originally posted by **ochi12** November  9, 2024</sup>
This concept design is derived from the app-center icon.
I was supposed to put this to my YaruX repo but I designed it to be a dependency for Yaru so there is no way I can override the default gnome-software icon without manually removing and replacing this icon in my Yaru install. 

One thing I can say is the old gnome-software logo generally look ugly(at least for me). I see it as a briefcase instead of a shopping bag.

I didn't started 16, 24, 32, 48 sizes and symbolic yet since I don't know yet if this will go anywhere

## NEW
image

## CURRENT
image

</div>

ochi12:
A redesigned gnome software icon that uses the shopping bag design of Ubuntu's app-center icon
image

Fixes #4156


ochi12:
I think I need to change the handle color to grey?

Muqtxdir:
> I think I need to change the handle color to grey?

I agree,but only on small sizes smaller than 64

ochi12:
> > I think I need to change the handle color to grey?
> 
> I agree,but only on small sizes smaller than 64

I actually overlook into it, I actually used pure white in the handles for icons smaller than 48 so I think it already looks goods to me...

ochi12:
> I agree,but only on small sizes smaller than 64
> 
I actually overlook into it, I actually used pure white in the handles for icons smaller than 48 since the paper bag color is somewhat grayish. It stands out but not too much. A gray handle would actually make the aesthetic look bad <atleast for me>. I heavily looked into `these image` so I think it already looks goods to me...

image

Muqtxdir (approved):

PR #4159, opened 2024-11-10
Created org.gnome.Snapshot Link Targetting the camera-app icon. 

ochi12:
Ubuntu replaced cheese in favor of gnome-snapshot since 24.04.
<br>
## FROM
image

## TO
image


Feichtmeier:
I think this is a good change but let the others have a look, too!

Muqtxdir (commented):
hi @ochi12

thanks for looking into this issue, I think it would be nice to have a separate icon for snapshot rather than symlinking to the cheese icon. I know few people coming from past ubuntu releases have both apps installed so it would be confusing

my suggestion would be to have the round lens icon for snapshot without camera body

ochi12:
> hi @ochi12
> 
> thanks for looking into this issue, I think it would be nice to have a separate icon for snapshot rather than symlinking to the cheese icon. I know few people coming from past ubuntu releases have both apps installed so it would be confusing
> 
> my suggestion would be to have the round lens icon for snapshot without camera body

Well I might work on it but my guess is camera-app.png is made as a link for any camera app that ships with ubuntu by default. 

It should be the other way arround, removing the cheese links instead

Muqtxdir:
please go with your suggested option1

ochi12:
@Muqtxdir @Feichtmeier new changes...

Tested it alongside other yaru icons to see if it fits the aesthetics.
image

Muqtxdir (approved):
thanks 🧡

ochi12:
@Feichtmeier I need some approval in symbolic icon changes

Jupi007 (approved):
LGTM, thanks a lot @ochi12 !

PR #4174, opened 2024-12-01
Proposal: Small Change to Zipper Styled Mimetypes

ochi12:
I tried to add a little bit more character to zipper styled mimetype icons.
My First Take:
<image src="https://github.com/user-attachments/assets/a659e215-7134-4e10-bf79-6f8a7403f5c6" width=200 />
My take would be to make only the icon for 256x256 and 256x256@2x . It looks good on max size but looses detail on smaller sizes between 64x64 >= sizes > 48x48. 
> The only solution to this is to make the added zipper style simpler until it looks good on 48x48

... So I decided to just add a dept within the zipper.

|  Original  |  Added Dept  |
|  ---  |  ---  |
|    |    |


Feichtmeier:
@madsrh can you review this since you were the mime type maestro?

madsrh:
Thank you @ochi12 
I'm don't think your first take would fit with the Yaru style, so I'm glad that you only added the depth 😊 I think this looks good 👍

ochi12:
@madsrh this is odd 
image
I think we both missed it. It rendered the wrong icon 
image

I will try to look onto this.

PR #4186, opened 2024-12-23
Improve the Look of Gnome Shell Dialogs

### LINKED ISSUE #4177: Gnome Shell Dialogs Have Too Large Padding
I recently upgraded to ubuntu 24.10 and these dialog boxes are the first hing that caught my eye.
Screenshot From 2024-12-07 13-43-33
In my opinion, the padding is too large

I quickly edited some scss properties, build and install and came up with this.
|  before  |  after  |
|  --- | ---  |
|  Screenshot From 2024-12-07 13-43-33  |Screenshot From 2024-12-07 16-45-01|

I used padding value and radius value same as with quick settings. The top padding has larger padding to emphasize the heading.

I think we should implement this widely even with modal dialogs, etc. under gtk.

ochi12:
Task:
- [x] rework prompt-dialog
- [x] rework session-dialogs
- [x]  rework run dialog

### Prompt Dialog
|  before  |  after  |
|  --- | ---  |
|  |  |

### Session Dialog
|  before  |  after  |
|  --- | ---  |
|  |  |

### Run Dialog
|  before  |  after  |
|  --- | ---  |
|  |  |

I might retain the original border radius later on. 

resolves #4177

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_dialogs.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_dialogs.scss
@@ -1,5 +1,13 @@
 /* Modal Dialogs */
 
+// Yaru Custom Definition
+$dialog_padding: $base_padding * 4 $base_padding * 3 $base_padding * 3 $base_padding * 3;
+
+// Yaru Custom headline
+.headline {
+  @extend %title_4;
+}
+
 // style for all dialogs
 .modal-dialog {
   background-color: $bg_color;
@@ -50,6 +58,8 @@
 /* End Session Dialog */
 .end-session-dialog {
   width: 24em;
+  border-radius: $yaru_menu_border_radius; // Yaru Change: Override default radius
+  padding: $base_padding * 3; // Yaru Change: Override default padding
 
   // special style for session warnings
   .end-session-dialog-battery-warning,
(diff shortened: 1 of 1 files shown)


3v1n0:
Nice improvements, thanks... Can you also comment all the changes that have been done so that will be easier to handle when we merge with upstream again?

ochi12:
> Nice improvements, thanks... Can you also comment all the changes that have been done so that will be easier to handle when we merge with upstream again?

I'm going to look into it soon. I'm kinda busy at school right now.

aleasto:
Looks like this needs a rebase

ochi12:
@3v1n0 hmmm i think this pr is still relevant as it makes dialogs specific for yaru theme look better in my opinion.  I don't know if I covered everything .

gnome-shell/src/gnome-shell-sass/widgets/_dialogs.scss
```diff
@@ -1,5 +1,12 @@
 /* Modal Dialogs */
 
+// Yaru Custom Definition
+$dialog_padding: $base_padding * 4 $base_padding * 3 $base_padding * 3 $base_padding * 3; 
+
+.headline {
```
3v1n0, line 6: Can you add me to yaru comment here too?

3v1n0: Done.

3v1n0:
Ok, let's try this...

PR #4201, opened 2025-03-13
build: various meson cleanups

3v1n0:
See commits for details.


cinnamon-shell/src/meson.build
```diff
@@ -81,7 +69,11 @@ endforeach
 # theme_gresource_xml = files('data/cinnamon-shell-theme.gresource.xml')[0]
 # icons_gresource_xml = files('data/cinnamon-shell-icons.gresource.xml')[0]
 
-foreach variant: variants
+foreach variant: yaru_flavours
+  if variant == 'default'
+    variant = DEFAULT_VARIANT
+  endif
+
```
aleasto, line 76: Is this getting `mate` variants now?

3v1n0: Yes, I did it on purpose if someone would be interested in a mate-styled shell too, but it happens only if mate is enabled.

Eventually it would just end up in a green accent.

This won't affect gnome-shell soon, though.

aleasto (approved):

PR #4203, opened 2025-03-13
ci: Drop the in-house action to build deb files, reusing desktop team one

3v1n0:
No need to maintain ours anymore while the re-usable action can also do linting and sources generation

### DIFF
--- a/.github/workflows/build-deb.yaml
+++ b/.github/workflows/build-deb.yaml
@@ -2,68 +2,39 @@ name: Build debian packages
 
 on:
   push:
-    branches: [master]
+    # branches: [master]
+    branches: "*"
   pull_request:
     branches: [master]
 
-env:
-  DEBIAN_FRONTEND: noninteractive
-
 jobs:
-  check:
+  build-deb-packages:
     name: Ubuntu Devel
     runs-on: ubuntu-latest
-
     env:
-      UBUNTU_VERSION: devel
-      BUILD_CONTAINER: yaru-build-container
-      RUN_CMD: docker exec -t -w /src/yaru yaru-build-container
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deb.yaml
```diff
@@ -6,64 +6,16 @@ on:
...
-        $RUN_CMD $AS_USER dpkg-buildpackage -b --jobs=$CONCURRENCY_LEVEL
-        $RUN_CMD sh -c 'mv -v ../*.deb .'
-        $RUN_CMD sh -c 'ls -lht *.deb'
-
-    - name: Archiving built debs
-      uses: actions/upload-artifact@v4
-      with:
-        name: debian packages
-        path: ./*.deb
+      - name: Checkout authd code
```
aleasto, line 15: `s/authd //`

aleasto (commented):
This does not verify that the generated debs are installable anymore.

Does `canonical/desktop-engineering/gh-actions/common/build-debian@main` expose the debs in a predictable place, so we can continue verify the installability here?

.github/workflows/build-deb.yaml
3v1n0: Thanks, damned copies...

3v1n0:
> This does not verify that the generated debs are installable anymore.
> 
> Does canonical/desktop-engineering/gh-actions/common/build-debian@main expose the debs in a predictable place, so we can continue verify the installability here?

Oh good point... Indeed that was one of the reasons why I exposed the artifacts in that way... Let me handle that!

aleasto (approved):
LGTM thanks

PR #4220, opened 2025-04-16
gtk-4/tweaks: Actually override the default windowcontrols definition

3v1n0:
We were not overriding the case in which an image was child of an hovered or active parent, so do it properly now by handling such case as base theme does, so that we can fully override it.

LP: #2107388

### DIFF
--- a/gtk/src/default/gtk-4.0/_tweaks.scss
+++ b/gtk/src/default/gtk-4.0/_tweaks.scss
@@ -76,28 +76,35 @@ spinbutton:not(.vertical) {
 // titlebutton
 windowcontrols {
     button {
-      min-width: 24px;
-    }
+        min-width: 24px;
 
-    button > image {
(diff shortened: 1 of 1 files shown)


gtk/src/default/gtk-4.0/_tweaks.scss
```diff
@@ -76,28 +76,35 @@ spinbutton:not(.vertical) {
 // titlebutton
 windowcontrols {
     button {
-      min-width: 24px;
-    }
+        min-width: 24px;
 
-    button > image {
         $_base_button_color: transparentize($fg_color, 0.9);
         $_base_hover_color: transparentize($fg_color, 0.85);
```
aleasto, line 82: This seems to be unused now

3v1n0: Yeah good catch, bad copy-paste indeed!

aleasto (approved):

PR #4221, opened 2025-04-18
Icon For Papers

ochi12:
image
Just some basic changes on the evince icon. Made the glasses black and added the have cover with fold

reference (original icon)
image

My machine doesn't have the nicest display so feel free to test the icon and let me know if its too de-saturated


Jupi007 (changes requested):
Great work!
The icon color seems okay for me.

I just have one small nitpick, the light right part is not aligned on the grid on the 48px one:

image

ochi12:
thanks for the notice

ubuntujaggers:
Great work! I have a few small notes:

- On the 16px icon, the highlight is overlapping the border instead of neighbouring it - BUT it looks like this mistake exists in the original Evince icon (and at least one other icon of this shape that I checked), so this isn't your error, it must be mine:

image

- The orange gradient overlay on the 32px is at 90% opacity, all other sizes have 100% opacity

- The highlight on the 48px icon is at 80% opacity, the 16-32px have 70%

- The edge highlight is a bit brighter than we normally use nowadays (at least when it appears over a very vivid colour)... maybe set it to 60% opacity for these icons?

- I'm fifty-fifty on the choice of red - I wonder if going to the next red down (as it appears on the Suru palette) would work slightly better? Not sure though

Current:

image

Next red down, with orange overlays set to 85% opacity:

image

ochi12:
> Great work! I have a few small notes:
> 
>     * On the 16px icon, the highlight is overlapping the border instead of neighbouring it - BUT it looks like this mistake exists in the original Evince icon (and at least one other icon of this shape that I checked), so this isn't your error, it must be mine:
> 
> 
> image
> 
>     * The orange gradient overlay on the 32px is at 90% opacity, all other sizes have 100% opacity
> 
>     * The highlight on the 48px icon is at 80% opacity, the 16-32px have 70%
> 
>     * The edge highlight is a bit brighter than we normally use nowadays (at least when it appears over a very vivid colour)... maybe set it to 60% opacity for these icons?
> 
>     * I'm fifty-fifty on the choice of red - I wonder if going to the next red down (as it appears on the Suru palette) would work slightly better? Not sure though
> 
> 
> Current:
> 
> image
> 
> Next red down, with orange overlays set to 85% opacity:
> 
> image
> 
> EDIT: looking at these on my phone, which has a better screen, I do prefer the more saturated version personally.

I think it looks too vibrant on smaller icons. Any ways I don't have a great display

ubuntujaggers:
Hi @ochi12 @Jupi007, would be great to get this icon into the theme? What do you think?

Jupi007:
@ubuntujaggers It's fine for me. If you don't have any more changes to request, I can merge it 👍

ubuntujaggers:
All good from my point of view! :+1:

Jupi007:
Thanks a lot for this very nice icon @ochi12 !

PR #4229, opened 2025-05-10
adjusted radii of login entry. Initial commit

ochi12:
before:
Screenshot From 2025-05-10 11-06-07
after:
Screenshot From 2025-05-10 09-30-56

followed 9 px radius of `entry` widget using `base_border_radius` + 6

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
+++ b/gnome-shell/src/gnome-shell-sass/widgets/_login-lock.scss
@@ -262,6 +262,7 @@ $_gdm_dialog_width: 25em;
 
 .login-dialog-prompt-entry {
   @extend %system_entry;
+  border-radius: $base_border_radius + 3; // Yaru Change: Match Upstream button radius 
 }
 
 .web-login-dialog-content-overlay {


Feichtmeier:
Is this 9px coming from upstream?

ochi12:
> Is this 9px coming from upstream?

I think so, I looked upon gtk inspector

Feichtmeier:
> > Is this 9px coming from upstream?
> 
> I think so, I looked upon gtk inspector

well IMO we should just try to copy what comes from upstream for consistency and less maintenance
also shouldnt this be centralized in all shell component textfields (entries) in the scss file?

ochi12:
> > > Is this 9px coming from upstream?
> > 
> > 
> > I think so, I looked upon gtk inspector
> 
> well IMO we should just try to copy what comes from upstream for consistency and less maintenance also shouldnt this be centralized in all shell component textfields (entries) in the scss file?

I'm still kinda navigating the whole gnome-shell directory and its contents so I'm in fact confused. I'm gonna look into this.

3v1n0:
What's the story here, how should we proceed?

3v1n0:
Ping?

ochi12:
I don't know if these change is needed or should just follow upstream changes.  It's the same thing

3v1n0:
I'm fine to revert to upstream radius, I'll check

Issue #4230, opened 2025-05-28
Icon Revamp For System Monitor


ochi12:
The Icon Revamp Proposal is based on an IBM Computer with the green-colored pulse signal is from the color of the real CRT monitor is emitting.

![Image](https://github.com/user-attachments/assets/2f9b23b2-16ca-4634-a429-51a12d4c047a)

upstream icon:

![Image](https://github.com/user-attachments/assets/2c218e12-ead6-4127-b4b1-9d68dfc0fd75)

before:
![Image](https://github.com/user-attachments/assets/a5c97832-43a0-4c76-adb2-a5d8983e780c)
after:
![Image](https://github.com/user-attachments/assets/d734b069-0fda-48cf-8cb1-1ec6781145ca)

<br>
<br>

Feichtmeier:
> before: ![Image](https://github.com/user-attachments/assets/d734b069-0fda-48cf-8cb1-1ec6781145ca)
> 
> flattened: ![Image](https://github.com/user-attachments/assets/c9fb9ada-2ecd-4015-9339-6ee2b8f5ff99)
> 
> I'm gonna try your suggestion on yarufying the literal upstream icon soon

I love the vibe and idea but I agree that partly because of the different oblongs inside each other one keeps starring at the corners and border radii and this somehow makes it feel busy. The cool thing about the current icon is that its super simple.

ochi12:
currently trying out @Muqtxdir  suggestionn

![Image](https://github.com/user-attachments/assets/4b56bcd4-6ebe-4af3-977f-9a9b1bbd6cd3)

ochi12:
> imho, the recent iteration looks really nice 🙂

you mean the one with yellow pulse?

PR #4232, opened 2025-05-30
Icon Revamp for System Monitor

ochi12:
# Overview
Main point: Following the upstream design closely 
**targets** issue [`#4230`](https://github.com/ubuntu/yaru/issues/4230)

| Upstream  | Yaru |
| ---- | ---- |
|  |  |

**considerations for the size 256 dial design:**
- alternative design
- increase diameter

# Applied
| On App Overview |
| ------ |
| Screenshot From 2025-05-30 18-53-16 |

<table>
   <tr>
      <th colspan=5>On Dock</th>
   </tr>
   <tr>
     <th>16</th>
     <th>24</th>
     <th>32</th>
     <th>48</th>
   </tr>
   <tr>
     <td>
        
    </td>
     <td>
        
    </td>
     <td>
         
     </td>
     <td>
         
    </td>
   </tr>
</table>

# Note for the pulse design
> that size 24, 32, and 48 have visual correction on the pulse design.
All follows 1 px thickness but with a factor. size 48 has a factor of 1.6 which means the new thickness of the pulse is 1.6 px a 30 percent incresse (60 % / 2). Size 32 and 24 on the other hand has factor 1.3 which means the new thickness of the pulse is 1.3 px a 15 percent increase (30% / 2).

> This type of visual correction is very common to google's material design to avoid aliasing in diagonal shapes or complex geometries (paper clip for example). I notice this aliasing on the mentioned sizes, so a slight intentional mis-pixel-fitting  as optical compensation is required.

#4230


ochi12:
I will just wait for a review and might also considder adding the paper fold effect onto it

ubuntujaggers:
This is a very nice icon and definitely a closer match for upstream. I wonder if the bezel on 16px and 32px is one pixel too thick, but it's hard to know if changing it would look worse without trying it.

ochi12:
> This is a very nice icon and definitely a closer match for upstream. I wonder if the bezel on 16px and 32px is one pixel too thick, but it's hard to know if changing it would look worse without trying it.

I agree, this is why I closely work with reviewers. Will  fix the sizes 16 and 32 soon

ochi12:
Made these additional changes:
- fold effect
- changed the base color and screen color
- decreased the bezel color of size 32 icon but kept the original size 16 bezel size as it could appear like a strip of pixel on actual preview
rect4158-8

3v1n0:
Let's go with this, it's already nice enough! :)

I hope it has not issue with the light theme, but both dock and overview should be fine anyways.

PR #4235, opened 2025-06-04
Yaru Icons For Caffeine

ochi12:
I recently made some changes to caffeine extension and made its icons allowed to be themed. 

prefs page:
image

caffeine icon:
Screenshot From 2025-06-04 11-16-49


ochi12:
image

made the infinity symbol more organnic instead of round. Yaru seems to move to wards organic shapes

PR #4244, opened 2025-06-29
Changing colour of Jockey icon to use colours from theme palette

ubuntujaggers:
I'm not 100% why, but when I did the Jockey icon I didn't use a green from our palette. I think at the time I thought it was more important to use a realistic green that a circuit board might have in real life, but I now think this icon looks better using greens from the Suru palette.

image


ochi12:
image
did you use the darkest green as the base pallete?

Just for me, its a little on the too vibrant side

maybe tone down a bit like this?
image

I can send here the file if you want so you can reference it

also maybe you can simplify it a little bit by removing the gray block under the black chip?

image
like this and also the solder needs to be gray IMO. but off course sometimes real life doesn't look good in this situation so all in all gold looks better.

All in all the icon looks very nice

Jupi007:
@ubuntujaggers To be very honest I prefer the current realist green over the Suru one :)

ubuntujaggers:
It's a gradient between the darkest green and the second darkest green. I will try something a little less bright, if we don't come up with something we all like better it's fine to leave it how it is 👍

3v1n0:
Also make sure that the green doesn't clash too much with the accent colors we may have :)

Or well, we could even have accented icon, at the end the boards colors are quite customized now :smile:

3v1n0:
Also for the small icon, maybe we can do something more visible?

ochi12:
> Also for the small icon, maybe we can do something more visible?

It would be

> Also for the small icon, maybe we can do something more visible?

maybe increase the size of the Chip

3v1n0:
Can this closed now? As #4290 landed?

ochi12:
Definitely. 

Edit: I think this is a different topic that pr is for firmware updater.

What I suggest now for @ubuntujaggers  is to use the same color as the firmware-updater icon

PR #4248, opened 2025-06-30
Ptyxis: Add Missing Full Color and  Symbolic Icons 

ochi12:
- [x] missing yaru symbolic icons added
- [x] missing yaru full color icon added 

Screenshot From 2025-06-30 08-21-05
Screenshot From 2025-06-30 08-20-36
Screenshot From 2025-06-30 08-20-14


ochi12:
@Jupi007

Jupi007 (approved):
LGMT
@ubuntujaggers Is it okay for you?

Muqtxdir (changes requested):
hey, I think we might wanna do the same approach as other upstream distributions:
- bundle two different icons one for ptyxis icon and another for generic icon based on the options picked up during build (see here: https://gitlab.gnome.org/chergert/ptyxis/-/blob/main/meson_options.txt?ref_type=heads#L2)

ochi12:
> hey, I think we might wanna do the same approach as other upstream distributions:
> 
>     * bundle two different icons one for ptyxis icon and another for generic icon based on the options picked up during build (see here: https://gitlab.gnome.org/chergert/ptyxis/-/blob/main/meson_options.txt?ref_type=heads#L2)

hmm it depends on what build the ubuntu package for ubuntu follow. Does it follow the generic one? If that's the case then I can push actual ptyxis icon I designed

ochi12:
Update - I think the ubuntu build for the ptyxis does come with the generic icon but the name `org.gnome.Ptyxis` is still in there. I think this MR is ready for merge. Also with the new loupe icon

3v1n0:
This is UI freeze material, but it looks good to me, so once @Muqtxdir is happy I think we can go

3v1n0 (approved):

3v1n0:
Can you please rebase the branch?

PR #4258, opened 2025-07-20
Added missing cog-wheel-symbolic icon for gnome settings

ochi12:
I noticed that the cog wheel icon in wifi settings uses upstream icon. A quick fix is added now it uses proper yaru icon

Screenshot From 2025-07-20 15-23-17

Screenshot From 2025-07-20 15-33-28


Feichtmeier:
THanks! Don't we have this symbolic already and could make a symlink?

ochi12:
> THanks! Don't we have this symbolic already and could make a symlink?

If I do that I will also make the other cog wheel a symlink. Currently there is 3 copies of the cog wheel  icon.

Issue #4262, opened 2025-07-26
Missing icons: CSV and TOML


manlaya:
**Icon Request**

CSV (Comma Separated Values) and TOML (Tom's Obvious, Minimal Language ; configuration files) files are widely used, still corresponding mimetype icons are missing from Yaru theme.

The corresponding mimetypes (from xdg-mime-info) are respectively: `text/csv` and `application/toml`.

**Steps to locate the icon**

TOML icon can be create using the official logo (see below).
CSV icon may require to create a icon from scratch.

**Software where the icon is missing**

- Name: Nautilus [or Thunar, etc.]
- Version: 47.5
- OS version: Ubuntu 24.04

**Screenshots**

TOML: https://toml.io/en/

CSV: not yet.

PR #4263, opened 2025-07-26
Icons - add: toml, csv mimetype icons ; fix: yaml mimetype icon

### LINKED ISSUE #4261: Yaml MimeType icon does not work
Hi,

Thanks for your nice theme!
When using `yaml` files, it seems that the corresponding icon from Yaru is not working.

**Expected Behavior**

When creating/using `*.yml`  or `*.yaml` files, the yaru icon `application-x-yaml.svg`
(or its png size variant) should be used.

**Actual Behavior**

Instead, the icon showed is blank generic text icon (`text-x-generic`). 

1. Create a yaml file

```bash
cat > toto.yaml << EOF
name: toto
requires:
  - test
  - test2

EOF
``` 

2. check icon file using Nautilus
3. check mime type:

```bash
xdg-mime query filetype toto.yaml
```

4. See problem: the return of last command should
be `application/yaml` but the icon showed is the generic
text icon.

5. The quickest fix is to rename the icon, removing the 'x':
`application-yaml.svg` (and title in svg file), then re-generate all the png.
Testing re-installing from sources with this modification solved it for me.

**Yaru version**

from git describe: `25.04.2-0ubuntu1`
Also tested from out of the box ubuntu 24.04: `2024.04.2-0ubuntu1`,
and in Fedora 41 (with Gnome): `24.10.4-1.20250228git24.10.4.0ubuntu1.fc41`.

**Screenshots**

Expected behaviour:

Image

Actual behaviour:

Image

manlaya:
Hi,

Working with csv files and some pyproject.toml (among other things), it would be very nice to identify these at glance.

Here are my propositions, as mentionned in #4262 :
- a `toml` mimetype icon, based on the official logo (https://toml.io/en/) with the mimetype `application-toml` ;
- a `csv` mimetype icon, I created to reflect the 'comma separated values' idea.

More, I included in this PR a proposition to fix #4261, renaming the `yaml` mimetype icon to match `xdg-mime query` output.

After a build and install on my computer, all 3 icons are well installed/recognized for their respective file types.

These are propositions, I would be happy to discuss about it.

Best,


3v1n0:
@ochi12 can you maybe review this?

Jupi007 (changes requested):
Looks great, thanks @manlaya 

I would prefer to keep `application-x-yaml` along to `application-yaml`.
Can you create a symlink, please?
You need to edit this file: https://github.com/ubuntu/yaru/blob/master/icons/src/symlinks/fullcolor/mimetypes.list
And then run the script to generate them.

manlaya:
thanks for the review @Jupi007!
I have just added the symlink creation in the `mimetypes.list` file and I generated the icons, to link `application-x-yaml`  to `application-yaml` as suggested.
Is it ok this way ?

PR #4267, opened 2025-08-02
Added missing bluetooth-acquiring-symbolic icon

ochi12:
Hmm, I think this isn't an obvious miss, because the icon only appears abruptly, and I need perfect timing just to capture a screenshot.

Target missing icon: bluetooth-acquiring-symbolic

Before:
Screenshot From 2025-08-02 12-06-37

After:
Screenshot From 2025-08-02 12-35-14

I decided to place the loading indicator lower because placing it in the center made the Bluetooth context unclear.


Jupi007 (approved):
LGTM thanks a lot @ochi12 !

PR #4284, opened 2025-08-20
Chore: Improve and Sync Wi-Fi Encrypted Indicator Icons

ochi12:
The network-wireless-encrypted-symbolic icon is not drawn correctly. According to the its css class it is supposed to be sized, $scalable_icon_size * 0.5 meaning the base icon is supposed to be 8x8 not 16x16 px . With the pixel fitted changed, the lock-small-symbolic is also transformed into a symlink syncing with network-wireless-encrypted-symbolic

I was supposed to just use the lock-small-symbolic icon but noticed that an outlined icon doesn't fit well in an 8x8 canvas
so I went and tried to pixel fit the original network-wireless-encrypted-symbolic.

**All in all**:
- Solved: lock icon for gnome-shell wifi indicator and gnome-control-center are now the same.
- the icon is already too small to be outlined so it was changed to be a filled icon.

## Preview:
**Before:**
Screenshot From 2025-08-20 15-53-05

**After:**
Screenshot From 2025-08-20 16-45-42

### DIFF
--- a/icons/Yaru/scalable/status/lock-small-symbolic.svg
+++ b/icons/Yaru/scalable/status/lock-small-symbolic.svg
@@ -1,5 +1,1 @@
-<svg version="1.1" viewBox="0 0 8 8" xmlns="http://www.w3.org/2000/svg">
- <path d="m1 3v5h6v-5zm1 1h4v3h-4z" color="#000000" fill="#808080" stroke-width=".5" style="font-feature-settings:normal;font-variant-alternates:normal;font-variant-caps:normal;font-variant-ligatures:none;font-variant-numeric:normal;font-variant-position:normal;isolation:auto;mix-blend-mode:normal;shape-padding:0;text-decoration-color:#000000;text-decoration-line:none;text-decoration-style:solid;text-indent:0;text-transform:none;white-space:normal"/>
- <rect x="3" y="5" width="2" height="1" fill="#808080" fill-rule="evenodd" stroke-dasharray="4, 8" stroke-linecap="square" stroke-linejoin="round" stroke-width="2"/>
- <path d="m6 3c0-1.3807-0.61929-2.5-2-2.5s-2 1.1193-2 2.5" fill="none" stroke="#808080" stroke-linecap="square" stroke-linejoin="round" stroke-width=".75"/>
-</svg>
+network-wireless-encrypted-symbolic.svg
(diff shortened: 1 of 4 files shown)


Jupi007:
Thanks @ochi12 for this fix!
Do you know where is the upstream source for this file?

ochi12:
> Thanks @ochi12 for this fix! Do you know where is the upstream source for this file?

I think it came from adwaita icon theme

Jupi007:
Okay I found it [here](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/Adwaita/symbolic/status/network-wireless-encrypted-symbolic.svg?ref_type=heads&plain=1).
So, correct me if I'm wrong, gnome-shell use `network-wireless-encrypted-symbolic` specifically as a badge for the network icon. `lock-small-symbolic` doesn't exist upstream right?
The problem is that upstream icon is 16x16. But as they use 2px stroke, it is looking better when downscaled.

@Muqtxdir Do you know if this icon is used by any other app? Because if so, a 8x8 icon will cause problems.
IMO it should be 16x16 (from a 8x8 scaled up).

ochi12:
> `network-wireless-encrypted-symbolic` specifically as a badge for the network icon. `lock-small-symbolic` doesn't exist upstream right? 


Yes it doesn't .It is specifically for gnome settings only or that's what i though because I tracked the icon name through gtk inspector

ochi12:
> Okay I found it [here](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/blob/master/Adwaita/symbolic/status/network-wireless-encrypted-symbolic.svg?ref_type=heads&plain=1). So, correct me if I'm wrong, gnome-shell use `network-wireless-encrypted-symbolic` specifically as a badge for the network icon. `lock-small-symbolic` doesn't exist upstream right? The problem is that upstream icon is 16x16. But as they use 2px stroke, it is looking better when downscaled.
> 
> @Muqtxdir Do you know if this icon is used by any other app? Because if so, a 8x8 icon will cause problems. IMO it should be 16x16 (from a 8x8 scaled up).

we can make final changes and make it a 16 by 16 icon with scaled stroke to 2 px that will not make a difference even when scaled up to 16 by 16 since it is in svg format anyways and 2px stroke width will just scale to 1px. But anyways, better make it 16 by 16 then.

I will push the changes soon.

3v1n0:
Cool, nice one!

This is likely not UI-Freeze material so we can wait a bit longer for the fix, but I'll follow this

Jupi007:
@ochi12 Yes you got my point :D
Every symbolic icon should be 16x16. But as this one specifically is used at 8x8, we should draw it at 8x8 and then scale it to 16x16 :+1:

PR #4290, opened 2025-08-26
New icon to be used for firmware-updater

ubuntujaggers:
Hi all,

Sorry for delay - as discussed on Discord! I think this is the icon that is used by Firmware Updater (although the devs will have to copy it, because it's a snap that won't be themed):

image


Jupi007 (approved):
Awesome, thanks @ubuntujaggers !

PR #4300, opened 2025-08-28
Sync icon theme with upstream

Jupi007:
image

Related to #4147


ochi12:
just a tiny nitpick. how about lower opacity for the disabled icon instead just like the bluetooth icon

Jupi007:
I already set opacity to 0.5.
Upstream use a strike-through so I prefer to follow their design :)

PR #4315, opened 2025-08-31
Added preferences-desktop-screensaver fullcolor icons for mate and cinnamon

### LINKED ISSUE #4193: Add missing icons
Add icons:

Image
preferences-desktop-screensaver

Image
go-home

ochi12:
## Preview
image

partially resolves #4193


Jupi007 (changes requested):
Great icon, thanks @ochi12 

I noticed that the screen base was under the `shadows` layer:

Image

Pictogram and borders/highlights are also all under the `background` layer.

ochi12:
> Great icon, thanks @ochi12
> 
> I noticed that the screen base was under the `shadows` layer:
> Image
> 
> Pictogram and borders/highlights are also all under the `background` layer.

Hmm i think that is the correct way because if you look at an actual monitor the shadow the screen is upfront and the stand is way back so if you place it on a floor and shine a light on top then the shadow will cast at the screen base.

I will fix some of the problems soon maybe 3 days from now . I just rushed those PR because I'm preparing for vacation and I might forget about it once I comeback

Jupi007:
> Hmm i think that is the correct way because if you look at an actual monitor the shadow the screen is upfront and the stand is way back so if you place it on a floor and shine a light on top then the shadow will cast at the screen base.

You're totally right, I hadn't thought of that 😅
So I can suggest to create a new layer below `shadows` for the screen base.

> I will fix some of the problems soon maybe 3 days from now . I just rushed those PR because I'm preparing for vacation and I might forget about it once I comeback

No problem, I understand :)

ochi12:
image

I  made this last night.

Added Screen base layer with sub layers for background, borders , and shadows

Jupi007 (approved):
Perfect, thanks a lot @ochi12 🙏

PR #4321, opened 2025-09-07
Adjust 48×48px calculator-app pictogram thickness to match proportions of other sizes

ochi12:
Just noticed that the pictogram at size 48 for calculator-app icon is too thin when compared to other icons.

adjusted from 1px to 2px

|Before|After|
| --- | ---|
|Screenshot From 2025-09-07 15-36-00|Screenshot From 2025-09-07 17-33-43|


3v1n0:
Good, not a problem with the UI freeze here, one more thing... For next version, this is likely an icon that it would be nice to have accented :)

ochi12:
> Good, not a problem with the UI freeze here, one more thing... For next version, this is likely an icon that it would be nice to have accented :)

let's see. there's a lot of candidate. example the extension icon and startup disk creator

3v1n0:
> > Good, not a problem with the UI freeze here, one more thing... For next version, this is likely an icon that it would be nice to have accented :)
> 
> let's see. there's a lot of candidate. example the extension icon and startup disk creator

Yeah, let's maybe open an issue to track them, so that we can have an unique list

PR #4324, opened 2025-09-09
ci: Avoid installing locales and man pages

adombeck:
Speed up CI jobs by avoiding installation of locales and man pages.

The major performance gain that can be expected is that it makes the "Processing triggers for man-db" step a no-op. That step can otherwise take more than a minute, for example in [this recent run](https://github.com/ubuntu/yaru/actions/runs/17451766904/job/49557610398#step:4:104) it took 1m 51s:

image

UDENG-7865


3v1n0:
Thanks!

PR #4329, opened 2025-10-03
gnome-shell: Optimize contrast for warning, success and destructive texts

3v1n0:
In practice the shell only use the warning color, but optimize them all in case some extension relies on the other ones (yep, ubuntu-dock!)

Schermata del 2025-10-03 05-39-04
Schermata del 2025-10-03 05-39-16
Schermata del 2025-10-03 03-23-54

Compare with original screenshots in #4320

Partially overrides #4320

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_yaru-colors.scss
+++ b/gnome-shell/src/gnome-shell-sass/_yaru-colors.scss
@@ -1,3 +1,5 @@
+@import 'sass-utils';
+
 // Yaru custom definitions
 $base_hover_color: transparentize(white, 0.8);
 $base_active_color: transparentize(white, 0.75);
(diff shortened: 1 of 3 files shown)


3v1n0:
/cc @juanruitina, @sergio-costas

This is just using the automatic computation of the contrast.

I'm uploading hoping to be in time for 25.10 😂 (I've another change but it's for another PR as I wanted to be minimal here), we can re-iterate in future.

juanruitina:
Thank you, definitely an improvement. I see the error in the last screenshot using ellipsis instead of wrapping: the kind of thing that we miss when the system is in English. Would be good to get that fixed.

3v1n0:
> I see the error in the last screenshot using ellipsis instead of wrapping: the kind of thing that we miss when the system is in English. Would be good to get that fixed.

Sure, that's fixed too: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/3898 (tracked [here ubuntu side](https://bugs.launchpad.net/ubuntu/+source/gnome-shell/+bug/2126740))

Issue #4325, opened 2025-09-14
Improvement to nautilus icon


ochi12:
I noticed in size 32 of nautilus icon that there is an inconsistency. I always notice that it is a little bit larger than other rectangular icon. 

`terminal app compared to nautilus`
<img width="53" height="104" alt="Image" src="https://github.com/user-attachments/assets/2dabfbe3-f8a7-4ec6-a58e-c3ac31f29571" />

I went and checked the template and indeed it is larger by 2 pixel. Also if it was taken from a rectangular template then it is already far from that template and if it is the square template then worst because not the rectangular is scalled from the 224x224 to 208x208. 

I made this initial adjustment to make it more grounded to the rectangular silhouette. The small pictogram is transformed from a pill shape to a more super ellipse shape.

before:

<img width="667" height="172" alt="Image" src="https://github.com/user-attachments/assets/ead186fa-1fb6-4d16-b4e8-ca1ef69d2250" />

after: 

<img width="666" height="148" alt="Image" src="https://github.com/user-attachments/assets/0ed23777-ab54-4073-bd9c-87413e99c9f3" />

Also noticed that there is not pointy corner on my initial adjustments, it is more rectangular.

### the template

<img width="958" height="257" alt="Image" src="https://github.com/user-attachments/assets/0fef6905-4c8b-4897-87b4-674c58937d73" />
<img width="952" height="221" alt="Image" src="https://github.com/user-attachments/assets/ca42f2a2-151c-4d72-bb14-13103b431652" />

it is also far from the square template if indeed it is based on that template:

<img width="328" height="643" alt="Image" src="https://github.com/user-attachments/assets/98d76b4a-ce63-4f02-8593-cca6681cd4d7" />

ochi12:
## Places Icon
I noticed that the current folder type places icons doesnt have proper yaru elements such us border for smaller icons, hightlights and unclear template reference.

The Suggested design has a more squased design closer to an actual folder and using the horizonal oblong template.

Current:
<img width="1601" height="677" alt="Image" src="https://github.com/user-attachments/assets/8bd92d11-39d2-4400-bcbb-385633bb7948" />

Suggested design:

<img width="1601" height="677" alt="Image" src="https://github.com/user-attachments/assets/b507cfec-45fe-423e-b766-5c706ed1dd72" />

ochi12:
@Feichtmeier  you might want to also take a look at these.

eyelash:
[...]

PR #4337, opened 2025-11-06
Improvement to nautilus icon and folder icons

ochi12:
#closes  https://github.com/ubuntu/yaru/issues/4325

### DIFF
--- a/icons/src/fullcolor/default/actions/folder-new.svg
+++ b/icons/src/fullcolor/default/actions/folder-new.svg
@@ -31,9 +31,9 @@
      borderopacity="0.25490196"
      inkscape:pageopacity="0"
      inkscape:pageshadow="2"
-     inkscape:zoom="2.2194728"
-     inkscape:cx="193.96498"
-     inkscape:cy="160.62373"
+     inkscape:zoom="1.8938129"
+     inkscape:cx="209.10197"
+     inkscape:cy="97.950545"
      inkscape:current-layer="g2131"
      showgrid="false"
      inkscape:grid-bbox="true"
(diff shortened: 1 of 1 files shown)


eyelash:
There are some subtle changes in the symbolics on the folders. Should these changes also be applied to the symbolic icons (that are visible in the Nautilus side bar for example) to keep them consistent?

folders
symbolic icons

ochi12:
> There are some subtle changes in the symbolics on the folders. Should these changes also be applied to the symbolic icons (that are visible in the Nautilus side bar for example) to keep them consistent?
> folders symbolic icons

I think we should leave those as is since they are optimized for that role

ochi12:
@Feichtmeier  does CI handle rendering the accented variants?

Feichtmeier:
> @Feichtmeier does CI handle rendering the accented variants?

sorry no idea. 
Some years ago we had a script to render the icons
Forwarding this question to @3v1n0

ubuntujaggers:
I've had a look at the svgs now - couple of comments, not blockers though (because this commit is an improvement on current):

- The white highlight element on Nautilus is brighter/more opaque now - we toned these borders down at one point to make the icons less "buttony" but I don't think the brighter highlight looks bad, I just mention it for completeness

- I think the unblurred shadows on the folders look a bit too bold (but overall folders are better than the current ones)

- FWIW I think the original edging that we had (which someone must have accidentally removed when they edited the file) was really nice... you can see what it used to look like in [this commit](https://github.com/ubuntu/yaru/commit/3682137811a14f97524f8209b796ec65)

image

ochi12:
its a lot of work to refactor specially the group of folder but here's what I can do since I can easily access this parts:

- reduce shadow opacity (keep it no blurred)
- reduce the dark border on smaller icons
- reduce the opacity highlights

all in all i think the highlights are not the problem, its just the dark stuffs (shadow and border) that makes it appear bolder. I will consider toning down the opacity of the highlights for the smaller icons tho.

Feichtmeier:
edit: moved to issue

PR #4338, opened 2025-11-09
feat! (major-refactor): adopt upstream with minimal tweaks in Gnome Shell

Muqtxdir:
### gnome-shell

- [x] update upstream files and symlink instead of maintaining separate src
- [x] adopt upstream's colors, styling and just minimally override typography, panel aspects to align with ubuntu look
- [x] update the build structure

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_tweaks.scss
+++ b/gnome-shell/src/gnome-shell-sass/_tweaks.scss
@@ -296,6 +296,63 @@ $yaru_popup_radius: $modal_radius * 1.05;
 .quick-toggle-menu {
   border-radius: $yaru_popup_radius;
 }
+
+.quick-toggle {
+  border-radius: $yaru_popup_radius;
+}
+
+/* More squircles...
+.datemenu-popover {
+  .icon-button,
+  .button {
+    border-radius: $yaru_popup_radius;
+  }
+}
+
+.quick-settings {
+  border-radius: $yaru_popup_radius;
+
+  .icon-button,
+  .button {
+    border-radius: $yaru_popup_radius;
+  }
+}
+
+.quick-toggle-menu {
+  border-radius: $yaru_popup_radius;
+
+  & .header {
+    & .icon {
+      border-radius: $yaru_popup_radius;
+    }
+  }
+}
+*/
+
+.quick-toggle-has-menu {
+  & .quick-toggle {
+    &:ltr {
+      border-radius: $yaru_popup_radius 0 0 $yaru_popup_radius;
+    }
+
+    &:rtl {
+      border-radius: 0 $yaru_popup_radius $yaru_popup_radius 0;
+    }
+
+    &:ltr:last-child { border-radius: $yaru_popup_radius; }
+    &:rtl:last-child { border-radius: $yaru_popup_radius; }
+  }
+
+  & .quick-toggle-menu-button {
+    &:ltr {
+      border-radius: 0 $yaru_popup_radius $yaru_popup_radius 0;
+    }
+    &:rtl {
+      border-radius: $yaru_popup_radius 0 0 $yaru_popup_radius;
+    }
+  }
+}
+
 /* OSD fixes */
 .osd-window {
   .level {


Muqtxdir:
this is big one, I tried breaking it down commit by commit - apologies for making the review harder!

Muqtxdir:
gnome-shell screenshots:

| before | after |
| ---------- | --------|
| Screenshot From 2025-11-11 00-07-02 | Screenshot From 2025-11-11 00-03-43 |
| Screenshot From 2025-11-11 00-07-15 | Screenshot From 2025-11-11 00-03-56 |
| Screenshot From 2025-11-11 00-06-44 |Screenshot From 2025-11-11 00-03-19 |
| Screenshot From 2025-11-11 00-12-10 | Screenshot From 2025-11-11 00-14-39 |

Initu-Castilhos:
It would also be great to have this implementation for Cinnamon.

Feichtmeier:
@Muqtxdir I am not sure if we should use this green for the success color though or the badges in the screenshot. looks kinda off brand
and thanks for your huge effort!

Muqtxdir:
> @Muqtxdir I am not sure if we should use this green for the success color though or the badges in the screenshot. looks kinda off brand and thanks for your huge effort!

Agree, Changed it to our palette green in `_dock.scss`

Muqtxdir:
Heads-Up: I will force-push commits in this branch to have only Gnome Shell changes so that it is easier to review. For rest all will push it in separate pr(s)

gnome-shell/upstream/theme/gnome-shell-sass/widgets/_login-lock.scss
```diff
@@ -51,25 +49,6 @@ $_gdm_dialog_width: 25em;
...
-  padding-bottom: $base_padding * 2;
-  padding-left: $base_padding * 2;
-  padding-right: $base_padding * 2;
-}
-
-.login-dialog-auth-menu-button-popup-menu-box {
-  padding: $base_padding * 3;
-}
-
-.login-dialog-auth-menu-button-item {
```
3v1n0: We need to keep these changes as they are used by patches that we provide in ubuntu

gnome-shell/upstream/theme/gnome-shell-sass/widgets/_login-lock.scss
3v1n0: Ignore the diff here, it's ubuntu-specific for now, will likely change for 50, but it's not yet there.

Feichtmeier (approved):

ochi12:
This is great! Focus on colors instead. I used vanila gnome shell for a while and the only thing i don't like about it are colors. vanila gnome shell + yaru-colors is basically the perfect gnome shell

ochi12:
@Muqtxdir  I'm to lazy rightnow on check out the changes you've made? Can you summarize it?

I assume you completely went upstream, then simply added color overrides to someting like overrides.scss. Or there are ubuntu specific patches that makes full upstream sync impossible?

ochi12:
@Muqtxdir 
https://youtube.com/shorts/3HU4Y2ACBQ4?si=cHdq7y3icnOSWgW7
Just saw this weird child + parent radius relation ship in the message tray. Is this upstream too?, or is it an old yaru value? The radius of that section must be theoretically `outer radius  - padding`.

ochi12:
Screenshot_2025-12-11-18-21-05-57_40deb401b9ffe8e1df2f1cc5ba480b12
Screenshot_2025-12-11-18-20-50-70_40deb401b9ffe8e1df2f1cc5ba480b12

Will open this issue upstream

juanruitina:
.@3v1n0 and I had a look at this. We think this may be going a bit too far bringing dowstream the Adwaita/GNOME aesthetics.

As we understand it, Yaru is different from Adwaita in that it is sharper, perhaps more elegant: thinner text, sharper corners. Using bolder headings and labels as suggested goes a bit against that, but we could see the case for it in terms of accessibility and content scannability. We couldn't find a similar usability rationale that justifies having Adwaita's even rounder corners.

We acknowledge that having less overrides facilitates maintenance, but maybe we can find a good balance. Also, we should probably define and document more clearly how Yaru is different from Adwaita as a theme, as we may have diverging opinions about it.

ochi12:
I couldn't agree on some parts tho. 
>> As we understand it, Yaru is different from Adwaita in that it is sharper

Yes it is sharper. I very much adore the contrast optimization of yaru. But the thing is yaru needs to adapt to some things. We can keep this sharp background, noticable border and the dark gray tone instead of adwaita's bluish gray dark mode. But I think is is a modern and better approach to just adapt to the the radii rules of upstream adwaita. The quick settings and message tray having rounded corners gives better isolation. If we couldn't agree with the upstream corners then we could atleast agree with making the message tray and quick settings corner in between the current yaru sizing and upstream.

Regarding the bold text in some parts, It is not just about accessibility but It just looks right in some part. Like hierarchy of text widgets like in the wifi toggle for example (bold text for the title, regular for the subtitle).

What I think of what we might bring too far is adapting the upstream colors except the ubuntu orange. I think we should keep developing our own color schemes.

aleasto:
> Regarding the bold text in some parts, It is not just about accessibility but It just looks right in some part. Like hierarchy of text widgets like in the wifi toggle for example (bold text for the title, regular for the subtitle).

I agree. Also in notifications, Adwaita uses bold for the title and regular for the description which imho looks better.

Feichtmeier:
IMHO, even though I rarely contribute to this repo currently, those micro changes on top of adwaita do not justify the maintenance routine in yaru to re-apply the changes every gnome release
So I would favour the symlink approach

Muqtxdir:
Hi @juanruitina @3v1n0 @ochi12 @Feichtmeier - Apologies for slow resposne, was caught in other things

> .@3v1n0 and I had a look at this. We think this may be going a bit too far bringing dowstream the Adwaita/GNOME aesthetics.

I think the initial draft when I opened PR was like this statement, I have since then made quite few iterations for better middle-ground:

- Re-added prominent yaru's palette colors and the contrast optimization aspects to the build via `_color-overrides.scss`
- Tweaked the font-weights to H1, H3, H5 weights from vanillaframework.io instead of the heavy bolds from upstream. (we were previously just using `font-weight`: `normal` everywhere)
- Holded on swapping the accent-color palette to upstream palette, that is something I agree with @3v1n0 to do post 26.04 release

> As we understand it, Yaru is different from Adwaita in that it is sharper, perhaps more elegant: thinner text, sharper corners. Using bolder headings and labels as suggested goes a bit against that, but we could see the case for it in terms of accessibility and content scannability. 

Hard-Agree on font aspect, Yes - we ought to use the proper font weights to better distinguish but we dont need to use the bold/heavy stylization of upstream.
 I think the attempt that I did in this PR looks good with Ubuntu Sans while also being distinguishable. (H1,H3,H5 - vanillaframework - normalized weights)

| Upstream | Yaru (Current) | Yaru (this PR) |
| ---------------| ---------------------|---------------------|
| Screenshot From 2026-03-11 17-45-10 | Screenshot From 2026-03-11 17-46-15 | Screenshot From 2026-03-11 17-45-55 |
| Screenshot From 2026-03-11 17-47-53 | Screenshot From 2026-03-11 17-48-09 | Screenshot From 2026-03-11 17-47-36 |
| Screenshot From 2026-03-11 17-50-32 | Screenshot From 2026-03-11 17-50-02 | Screenshot From 2026-03-11 17-50-18 |




> We couldn't find a similar usability rationale that justifies having Adwaita's even rounder corners.
- Using upstream's border-radius and shadows as current yaru border-radius was clipping/breaking the look in most of the gnome shell widgetry here and there
- plus this allowed to go with the route of minimal tweaks rather than manually changing every gnome-shell sass file

> We acknowledge that having less overrides facilitates maintenance, but maybe we can find a good balance. Also, we should probably define and document more clearly how Yaru is different from Adwaita as a theme, as we may have diverging opinions about it.

Thank you, my goal with this was to reduce the maintenance burden while still having the Yaru cosmetic look. I have updated the README.md to highlight the changes here - please let me know if you want it to much more detailed

3v1n0:
Indeed we have some differences, I still liked to have a smaller border radius in some places (in general in the pop-ups), but it's not fully my call.

I'm not particularly aware of places where this broke things, so would be hard to override it a bit (maybe we can have an intermediate value)?

aleasto:
@Muqtxdir maybe the notification title is a bit too thick compared to other bold elements?

juanruitina:
@Muqtxdir I appreciate the more moderate font weight for headings you are proposing, we make it clearer than it currently is but more subtly than GNOME. Thank you also for updating the readme file. Maybe say "sharper shapes" instead of "sharper corners" so advice is more generic, and covers icons too. This kind of documenting is precisely what I think we are missing, so it’s a good start.

About border radii, to summarise, the suggested approach of inheriting those styles:

- Has clear benefits in terms of maintenance. The new approach lets us delete almost 5k lines of code. Hard to argue against that.
- Avoids some clashes with GNOME widgets.
- Aesthetically, it is perceived as more modern by @ochi12, but @3v1n0 and I have some reservations over how it affects what's unique about Yaru.

Given the lack of clear guidelines to the contrary, the visual complexity in aligning border radii when nesting, and the technical arguments for it, I think the changes are reasonable for the time being.

That said, I think we should explore more programmatic border radius handling (perhaps upstream) so tweaking it for Yaru wouldn't be so cumbersome. We could even consider giving users the freedom to customize it.

3v1n0:
I've rebased this and added some fixes and tweaked a bit the fonts and radius, let's see how it goes.

PR #4340, opened 2025-11-11
feat: use upstream palette for icons

Muqtxdir:
### icons:
- drop current icon resources and use upstream accents


3v1n0:
FYI you can just leave the CI to regenerate the icons (or re-generate them all together)... Or we can do it at later point if CI fails, so we keep the MR simple

Muqtxdir:
ouch, imma force push and let CI handle it. I manually triggered the targets by ninja  render+symlink-icons-accented

Issue #4358, opened 2025-12-02
Include Default Icon Variant to the Build Part of Accented Variants.


ochi12:

3v1n0:
> Since we are adapting upstream palettes for icons except for our own orange color

This is not defined yet, we need some design decisions, but for sure we are not doing it for 26.04

So I'm fine supporting more colors, but not replacing ours yet

ochi12:
> > Since we are adapting upstream palettes for icons except for our own orange color
> 
> This is not defined yet, we need some design decisions, but for sure we are not doing it for 26.04
> 
> So I'm fine supporting more colors, but not replacing ours yet

If upstream colors is not planned yet atleast for 26.04 , then can we include our orange color as part of our accented variants and include it to be rendered? refactoring changes from an 2 source files could be a tedious process. One example is with what I'm currently working for folder icons.  

If not, then I will just include all source icon for yaru, yaru-variants, yaru-mate for now

ochi12:
@Muqtxdir the changes I'm currently working on nautilis and folder icons would benefit a change like this. Basically add the default variant in the accent variant build. What do you think?

PR #4359, opened 2025-12-03
Add improvements to nautilus icons and folder icons

ochi12:
closes https://github.com/ubuntu/yaru/issues/4325


ochi12:
@Jupi007  could you actually test this? This is very display color sensitive so I'm not sure testing this in my display. 

I have these parameters to adjust:
- Front section of the folder: Lighten overlay
- Back Section of the folder: Darken Overlay

I will not add the pictogram darken overlay as a parameter to review since it will be sync with the back section's darken overlay.

I also have to note that I still haven't made the default and mate variants  for now so that once the accented variant is good to go, then I can proceed adding the rest of the source files.

Jupi007:
@ochi12 no problem, I will test it this afternoon once back home 👍

Jupi007:
The overlay looks fine on my screen :+1: 

What do you think about add a subtle depth shadow to the pictogram?

image

ochi12:
> The overlay looks fine on my screen 👍
> 
> What do you think about add a subtle depth shadow to the pictogram?
> 
> image

Will it visually work with other pictograms tho?

Jupi007:
Yes, it works quite well with the other pictograms.
Maybe try shifting it by 2px on the Y axis and 1px on the X axis (on the 256x version) — I think it works better that way.
I haven't tried it at smaller sizes — maybe it won't work.

ochi12:
> Yes, it works quite well with the other pictograms. Maybe try shifting it by 2px on the Y axis and 1px on the X axis (on the 256x version) — I think it works better that way. I haven't tried it at smaller sizes — maybe it won't work.

the color shift from light to dark tone already gives a natural internal shadow illusion so I think smaller sizes are already fine without shadows (well in theory) 

for smaller size we could add  inset outline at 0.25 or 0.20 px thickness to create a soft shadow when rendered. But then again, smaller icons are harder to manage and harder to refactor. If my theory above is right, then we could refactor the pictograms for size 256*256

image

I think this is a better stamp effect. What do you think?

### Option 2: Stamp highlight instead of shadow (macos like)
A:
image

B:
image


But then again i have a strong opinion on not applying this to smaller icons since we have small wiggle room, the minimum drop shadow on those sizes is 1 px and it would create a `deep stamp effect`

ochi12:
@Jupi007  what do you think

Jupi007:
Sorry for the late reply, too much work these days...

I pretty much like the B. example with dark+light shadow.

ochi12:
@Jupi007 Currently I chose black @ 8 percent opacity  for the dark area and  and white @ 20 percent opacity for the light area for the Stamp effect. Do you have any knowledge on why mate and default colors are not included in the accented build. It would have been much easier instead of making 3 source files. I opened a separate issue [here](https://github.com/ubuntu/yaru/issues/4358). We could implement it through this PR just so we immediately benefit from it.

refactoring the source file for other variants is just a tedious (or so I thought, don't know the reason behind it)

Jupi007:
Sorry for the late reply.

> Do you have any knowledge on why mate and default colors are not included in the accented build.

Well it is quite simple : default and mate variants use a different style than simple accented ones.
The default variant use a gradient from purple to orange so it looks like the default ubuntu wallpaper (which use a gradient from purple to orange). About the mate variant, the color used is also different from mate accent color, to mimic mate wallpaper.

ochi12:
> Sorry for the late reply.
> 
> > Do you have any knowledge on why mate and default colors are not included in the accented build.
> 
> Well it is quite simple : default and mate variants use a different style than simple accented ones. The default variant use a gradient from purple to orange so it looks like the default ubuntu wallpaper (which use a gradient from purple to orange). About the mate variant, the color used is also different from mate accent color, to mimic mate wallpaper.

ok got it. What is the exact orange used for accents tho?

Jupi007:
@ochi12 Well it should be ubuntu's orange `#e95420`:

image

ochi12:
@Jupi007 I just pushed new files for default and mate. Also applied the stamp effect to sizes 256 . Can you check it out before I generate the png files

Muqtxdir:
hi @ochi12 
Apologies for delayed response, going through the thread. I will prefer to go with the stamp effect for the pictograms (option-b).

ochi12:
@Muqtxdir  @Jupi007  can you review the current progress before I push default and mate variants?

3v1n0:
Please, let us know if we can get this in, UIF is getting close, so a decision here would be nice to have since this is a quite noticeable change

ochi12:
could someone else render the pngs? I'm currently busy with other things.😅️

Jupi007:
> could someone else render the pngs? I'm currently busy with other things.😅️

@ochi12 I think the CI can handle it once merged.

ochi12:
> > could someone else render the pngs? I'm currently busy with other things.😅️
> 
> @ochi12 I think the CI can handle it once merged.

nice

Feichtmeier:
Fyi added the update branch button so you can update a pr from master

eyelash:
I just noticed that the new folders don't have the origami fold in the bottom right corner. Was that a deliberate choice to remove it or an oversight?

I also noticed that the shadow between the front and the back looks stronger in the filemanager-app than in the folders or at least the orange in the filemanager-app looks brighter.

ochi12:
> I just noticed that the new folders don't have the origami fold in the bottom right corner. Was that a deliberate choice to remove it or an oversight?
> 
> I also noticed that the shadow between the front and the back looks stronger in the filemanager-app than in the folders or at least the orange in the filemanager-app looks brighter.

I decided to remove it

3v1n0:
@ochi12 the CI did not succeed in generating the rendered icons as per:

```
[8](https://github.com/ubuntu/yaru/actions/runs/22334977274/job/64625381375#step:8:23189)
[16x16/status] linking blueman-disabled.png -> bluetooth-disabled.png
  ERROR: could not find symlink file "folder-open.png" in /home/runner/work/yaru/yaru/icons/Yaru/16x16/status
FAILED: [code=1] meson-internal__render+symlink-default 
env MESON_SOURCE_ROOT=/home/runner/work/yaru/yaru MESON_BUILD_ROOT=/home/runner/work/yaru/yaru/_build MESON_SUBDIR=icons 'MESONINTROSPECT=/usr/local/bin/meson introspect' /home/runner/work/yaru/yaru/icons/src/symlinks/generate-symlinks.sh --variant default
```

Can you please adjust the symlnks?

PR #4370, opened 2026-01-01
Improve cursors

### LINKED ISSUE #4367: Improving cursors
### Section A: The hand cursor
The current hand cursor design looks like optimized too much for small cursor size settings. When encreasing at larger sizes the design looks like it is too optimized for small size. The hand corner's bottom left corner looks sharp. The thumb design is too simple (which is why I mention that it looks like it was designed for small variant). Also I noticed the the grabing icons has a punch gesture instead of a grab gesture. Lastly the pointing gesture doesn't look good at size 24. The nuckle bumps are not so obvious at these size.

### Section B: shades for contrast optimization
Image

Referencing the macos cursors (image from above), I think I now understand the intent of having white fill colors to some of their cursors. Focusing on hand cursor which I think would have better semantic when in white fill color.

**Screenshots**
For the proposed hand cursors design. 
Before:
Image
After:
Image

ochi12:
v-one
yaru-cursors

v-two
yaru-cursors

closes #4367


ochi12:
@Jupi007 do you know someone who can verify if the current gnome-cursors mapped to other naming convention is correct. I just pushed new changes under commit hash `3b2be6a` .  Also modifying adwaiting upstream source file to also add color coded slices for platform specific cursors under this commit `4232c27`

image

Jupi007:
@ochi12 Sorry for the late reply.
Tbh I have no idea who to ask about this. I'm not that experienced with cursors.
@Muqtxdir maybe?

ochi12:
@Muqtxdir 
image
Upon looking the the free-desktop cursor spec, these cursors are meant for DnD (typically the default cursor + specific icon for the function).

But then I found out that these cursors are also for DnD.
image

I can't decide which convention to follow, `grabbing + function` or `default cursor + function`. Well freedesktop's cursor spec recommend using the default cursor. But I'm not the one to decide

3v1n0:
Do we want to land these by 26.04?

It would be nice to get a review so that we can decide which version we should pick

ochi12:
> Do we want to land these by 26.04?
> 
> It would be nice to get a review so that we can decide which version we should pick 

I think we can't. I'm still waiting for a proper list of (Gnome cursors + others) and some mapping list. Unless we drop third party support for now and start with gnome cursors only but it is unlikely to happen. Currently I can't reuse the mappings of the current one because the file names for each cursors are named different. The automation I added is rendered useless unless I have the full list of (gnome cursors + other) and the mapping list.

the linking list I last pushed is also incorrect.

PR #4400, opened 2026-03-12
Update gtk theme with upstream and add accented assets

3v1n0:
Add generator for gtk assets, and generate all the accents


aleasto:
@3v1n0 this zero-ed out some GTK3 files like `gtk/src/default/gtk-3.0/assets/check-symbolic.symbolic.png`

See https://bugs.launchpad.net/ubuntu/+source/yaru-theme/+bug/2147544

3v1n0:
mhmh

Issue #3838, opened 2023-03-05
Yaru accent color for calculator


vadimk1337:
V 42.2 The calculator version changes the color of the equal sign depending on the accent color

Why not change from the yellow calculator icon to different colors just like with the folder?
This would be consistent with the idea that not only the colors of the application itself change, but also the calculator icon
I think it's a good idea. 

I'm not saying that you need to do this for every application, I suggest doing it only for the calculator.

Muqtxdir:
seems like duplicate of #3756 

It's not good move to override the color opted by the app developers

VadimVadimVadim1337:
Isn't it the developers of yaru who made the icon and it's exclusive to yaru?  It's just that in the snap version there is a different icon.

VadimVadimVadim1337:
![image](https://user-images.githubusercontent.com/36221770/222946318-0fa4b1c1-ddd6-42cf-815b-87fbb66bff25.png)
This is not a duplicate, the developers fixed the accent color issue in the calculator a long time ago

![image](https://user-images.githubusercontent.com/36221770/222946361-e480dcb1-aa2a-4603-a7f9-12cbd2c8da72.png)
I'm talking about this icon, if it's not called an icon, I don't know what you call it in English?

VadimVadimVadim1337:
The icon is yellow and the equal sign depends on the accent color. Is my idea clear?

VadimVadimVadim1337:
![image](https://user-images.githubusercontent.com/36221770/222947359-c506e2d4-73ed-4b83-898f-19f325b8fa6d.png)
I would like it to be like this

Muqtxdir:
ah I understand what you meant now, sorry for confusing and tagging this as duplicate of the other issue. 

I think having an accent color based app icon is possible but I'd like to hear from others before moving forward with your suggestion.

cc: @ubuntu/yaru-team

Jupi007:
If the = button is using the accent color, I'm totally in favor of use it in the icon :+1:

madsrh:
What @Jupi007 said 😺 

Thanks @VadimVadimVadim1337

VadimVadimVadim1337:
Wait guys, am I wrong? I downloaded the snap version where the accent changes, but in the deb version it looks like it doesn't? I'm confused, can you check? I do not understand what is going on. If the accent for some reason does not change, then for now it is not necessary

VadimVadimVadim1337:
[...]

PR #4401, opened 2026-03-12
Calculator Icon: Use accent color

3v1n0:
See previews in the diff

Closes: https://github.com/ubuntu/yaru/issues/3838


3v1n0:
No objections so far (short time indeed), but let's get this in and in case we can tweak it better before the release.

PR #4406, opened 2026-03-14
Improve accented icons rendering

3v1n0:
- Avoid duplicating icons for rendering the default variant
- Renders icons for dark icon theme when needed
- Improve contrast for icons rendering


3v1n0:
Let's merge so hopefully CI will do the job of re-generating everything... Hoping it does not fail :joy:

PR #4410, opened 2026-03-17
gnome-shell/tweaks: Properly override all panel button properties

3v1n0:
We were selectively setting some panel button css properties, but we were missing many as per upstream usage of the panel_button mixin.

So, override the panel_button definition, by forcing the dark style for buttons all the time, and re-use it to ensure that the yaru style is used across all the panel elements

/cc @Muqtxdir, @aleasto

### DIFF
--- a/gnome-shell/src/gnome-shell-sass/_tweaks.scss
+++ b/gnome-shell/src/gnome-shell-sass/_tweaks.scss
@@ -72,6 +72,15 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


Muqtxdir:
@3v1n0 does the quick toggle button color change to gray shade when clicked or hovered for you?

3v1n0:
> @3v1n0 does the quick toggle button color change to gray shade when clicked or hovered for you?

It did change in your initial PR, then I fixed with 4b78b2bacd9dca45441385976dc7b89cc36f5ce4

PR #4420, opened 2026-03-27
Generate accessible colors with OKLCH instead of RGB

juanruitina:
RGB is not a perceptually uniform colour space, so relying on it to generate lighter/darker shades may result in inaccurate hues. Relying on [OKLCH](https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl) instead may help us, for instance, generate darker shades of Ubuntu orange that do not become too dull; in other words, new shades are preceived closer to the original accent colour.

This code was generated with Claude Opus. It borrows code from https://bottosson.github.io/posts/oklab/ for the functions converting RGB values to OKLCH and viceversa; this code was written by the creator of OKLCH itself. OKLCH is already supported by modern CSS. Shades are bit livelier and closer to the original accent as a result.

image

Related to https://github.com/ubuntu/yaru/issues/4325, https://github.com/ubuntu/yaru/pull/4359

### DIFF
--- a/common/test-sass-utils.scss
+++ b/common/test-sass-utils.scss
@@ -141,5 +141,5 @@
 
 @debug test('optimize-contrast black/blue', optimize-contrast(black, blue), #1769ff);
 @debug test('optimize-contrast white/lightgray', optimize-contrast(white, lightgray), #767676);
-
-@debug test('optimize-contrast green/green target 3.5', optimize-contrast(#00ff03, #00ff03, $target: 3.5), #008500);
+@debug test('optimize-contrast white/orange', optimize-contrast(white, #e95420), #d6430a);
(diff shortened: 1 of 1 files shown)


3v1n0:
This seems interesting, I would try to re-generate the icons with this and see were we get.

juanruitina:
> This seems interesting, I would try to re-generate the icons with this and see were we get.

I tried to locally, but I failed to generate all of them. I managed to add a comparison above for the document folder one, for demonstration purposes.

3v1n0:
Yeah, indeed it's a nicer shade!

ochi12:
Finally I hope these are the final changes. Thanks @3v1n0

PR #4431, opened 2026-04-03
build: updates to meson build system

Muqtxdir:
### description

- drops the X11/Xorg session entry and its `xsessions`
- updates to mode.json to be inline with ubuntu-desktop
- replaced all `join_paths()` calls with the `/` path operator across all components ([please see here](https://mesonbuild.com/Release-notes-for-0-49-0.html#joining-paths-with-))
- drop hardcoded paths and replaced with `get_option('datadir')`.
- drop post install-scripts and utilize gnome module calls as its replacement


3v1n0 (commented):
Thanks, LGTM, we can merge this once 26.10 cycle starts

PR #4464, opened 2026-06-01
design: upgrade app grid pagination dots to modern horizontal one's

devcodeww1:
Design Proposal
This patch modernizes the application overview layout by upgrading the small, generic circular pagination indicator dots into sleek, modern horizontal  indicators.

Key Modifications
- Changed the `.page-indicator-icon` dimension template from uniform bounding squares to a flexible layout (16px width by 6px height).
- Set an explicit structural `3px` border radius to maintain matching smooth, rounded contours across the design system.
- Adjusted active page elements (`:checked`, `:active`) to dynamically widen to `32px` using the official vibrant Ubuntu Orange hex code accent (`#E95420`).

Verified locally against running desktop environments to ensure crisp rendering, accurate alignment, and optimal UI accessibility.
BTW: this is my first time committing to Ubuntu and my goal is for this to make it into the next upgrade if not, i will try until i will make it. thank you for checking my work - devcodeww1
these


Mohataseem89:
I like the idea of making the pagination indicators more prominent. One question: is there a reason to use a hardcoded `#E95420` value instead of an existing Yaru theme variable? Using a theme variable might help maintain consistency with future accent color changes.

devcodeww1:
oops! I've updated the code to swap out the hardcoded #E95420 hex value for the $orange palette variable. This ensures everything stays completely consistent with future accent color changes. Thanks for the review!

gnome-shell/upstream/theme/gnome-shell-sass/widgets/_app-grid.scss
```diff
@@ -120,14 +120,19 @@ $app_folder_size: 720px;
...
-    background-color: $system_fg_color;
+    width: 16px;
+    height: 6px;
+    border-radius: 3px;
+    background-color: rgba(255, 255, 255, 0.3);
     transition-duration: 400ms;
+
+    &:checked, &:active {
+      width: 32px;
+      background-color: $orange;
```
3v1n0, line 134: We already have `-st-accent-color`

gnome-shell/upstream/theme/gnome-shell-sass/widgets/_app-grid.scss
```diff
@@ -120,14 +120,19 @@ $app_folder_size: 720px;
...
+    }
```
3v1n0, line 135: this code should be in `tweaks` if any

PR #4467, opened 2026-06-10
Drop humanity dependency

3v1n0:
The theme is deprecated and we don't want to rely on it


aleasto (approved):
Ubuntu Desktop has been shipping without humanity by default making the Inherits effectively void.
We haven't seen complaints of missing icons, so LGTM.

PR #3637, opened 2022-05-11
Add maximized_offset

imageck:

PR #4471, opened 2026-06-19
Update themerc

fran00mmcc:
Deleted Maximized offset=4 line. This makes selecting the "close" button more easy on maximized windows. Before, if you clicked just on the top right (or left) corner of the screen, "close" button isn't selected. Now, with the deletion of this line, yes.

Also, this changes unifies Yaru-default xfwm4 (white variant) with Yaru-dark xfwm4 (dark variant) , as this line doesn't exist on Yaru-dark xfwm4's themerc file.

### DIFF
--- a/xfwm4/src/default/xfwm4/themerc
+++ b/xfwm4/src/default/xfwm4/themerc
@@ -13,7 +13,6 @@ full_width_title=true
 title_vertical_offset_active=0
 title_vertical_offset_inactive=0
 button_offset=4
-maximized_offset=4
 button_spacing=0
 shadow_delta_height=2
 shadow_delta_width=0


Jupi007:
Hi @fran00mmcc and thanks for your PR.
Can you add before/after screenshots, please?

fran00mmcc:
> Hi @fran00mmcc and thanks for your PR. Can you add before/after screenshots, please?

Of course!

I did 3 screenshots using 3 xfwm4 themes on Devuan Excalibur, these themes comes from yaru package called yaru-themes-gtk, avaiable on my distro through the main repos, that are almost identical from Debian Trixie.

I pulled this request because I noticed that the close button behaved different, depending if it was selected "yaru", the white theme, or "yaru-dark", the dark theme, both of them having 3 variants: normal, hdpi, and xhdpi. 

When using the white variants, close button didn't appeared as selected, which was annoying to me because when I want to close a window, I just drag the cursor to the top right corner, without thinking if the cursor is inside the "close" window, or not. When using dark variants, this didn't happen. 

Reading the themerc files of the xfwm4 themes I discovered white and dark themes differed on one line: the "maximized_offset=4", which makes buttons go left by 4 pixels (I think).

I really don't know which is the expected behaviour, there are 3 options:

1. Expected behaviour is the current one, or
2. Expected behaviour is having both dark and white variants of themerc file of the xfwm4 yaru theme to have the "maximized_offset=4" line, or
3. Expected behaviour is having both dark and white variants of themerc file of the xfwm4 yaru theme to NOT have the "maximized_offset=4" line.

I expect the correct option, or more precisely, the expected behaviour, to be the third one, but as I almost know nothing about UI editing, and Linux in general (I just opened a Github account to talk about this), I would be glad that someone corrects me about this.

Here are a few examples of this, using yaru-hdpi. Cursor is located at top right corner of the screen. Window is maximized. All screenshots are 500x500px for comparing:

This is the original yaru-hdpi. Note how "close button" is NOT selected:
Captura de pantalla_2026-06-19_22-23-20

This is yaru-hdpi with the "maximized_offset=4" line deleted. Note how "close" button is selected:
Captura de pantalla_2026-06-19_22-24-17

This is the original yaru-dark-hdpi. Note how "close" button is selected:
Captura de pantalla_2026-06-19_22-24-53
#4471 

I'm sorry if I have said or done something wrong, I'm new on this :)

Jupi007:
Hi @fran00mmcc and sorry for the late reply.
Regarding to #3637, I think `maximized_offset=4` should be added to the dark theme instead of being removed from the light variant.

fran00mmcc:
@Jupi007 ok! but, what are the benefits to use `maximized_offset=4`? I thought that expected behaviour when moving the cursor to the top right (or left) corner of the screen, is to select the "close" button.

Jupi007:
Oh, okay, I understand now.
Well, #3637 was intended to fix the maximised look. Without 'maximized_offset=4', there is a 4px shift.
However, we didn't realise at the time that this was preventing the button from being clicked when the cursor was moved to the top right.

I am personally in favor of your suggested modification.
What do you think about this @Muqtxdir @Feichtmeier  ?

Feichtmeier:
> Oh, okay, I understand now. Well, #3637 was intended to fix the maximised look. Without 'maximized_offset=4', there is a 4px shift. However, we didn't realise at the time that this was preventing the button from being clicked when the cursor was moved to the top right.
> 
> I am personally in favor of your suggested modification. What do you think about this @Muqtxdir @Feichtmeier ?

I trust on your expertise :)

PR #4479, opened 2026-07-26
fix: add missing symbolics used in `libadwaita`

Muqtxdir:
### description
- added couple of new symlinks for the missing libadwaita symbolics
- moved `adw-*` prefixed symbolics target-dir to `libadwaita/` to better track which ones are used in `libadwaita`
- updated `generate-symlinks` and refreshed the symlinks post the changes


3v1n0:
Cool, thanks!

PR #4483, opened 2026-08-22
fix: outline the selected app in the Alt+Tab switcher

### LINKED ISSUE #4477: Cannot see the current app in the Alt+Tab switcher
**Expected Behavior**

Easily see the current/selected/highlighted application icon on the Alt+Tab switcher on any screen and lighting condition

**Actual Behavior**

The only difference between current/selected/highlighted icons is background color is only few percent of gray intensity,
which is only visible under ideal lighting condition. 

**System information**

Ubuntu 24.04 or 26.04
GNOME Shell 46.0 or 50.1

My problem is the same, or worse, as in https://github.com/ubuntu/yaru/issues/2320 "Low contrast on indicator of multiple windows",
which was closed in 2020, but the problem is still reproducible in Ubuntu 24 and 26 on any of my 2 laptops.
If I have to use them in a bright environment with a screen positioned at a slight angle (to reduce glare)
then there is no way to see which application is selected/highlighted in the Alt+Tab switcher window.
Here is an example (Ubuntu with Light or Dark theme, high contrast accessibility setting enabled or disabled - makes no difference).

Image

My problem is not just with yaru, which unfortunately uses close to 50 shades of gray to represent states of objects
and it is impossible for human eye to see so many shades,
but with gnome/ubuntu, which makes it super hard to modify that setting.
I wasted days begging ChatGpt and Claude to give me instructions on how to change that
and all I got was one hallucination after another, some of them prevented system from loading and requiring reinstall.

Finally, with a help of Claude, I came up with the following procedure, which works:
```
# On the first run make backup of the original resource
RESOURCE=/usr/share/gnome-shell/theme/Yaru/gnome-shell-theme.gresource
RESOURCE_ORIG=${RESOURCE}.bak # Original
if [ ! -e ${RESOURCE_ORIG} ]; then
  echo "sudo cp ${RESOURCE} ${RESOURCE_ORIG}"
  sudo cp ${RESOURCE} ${RESOURCE_ORIG}
fi

WORK_DIR=~/temp
mkdir -p ${WORK_DIR}
rm -rf ${WORK_DIR}/*
cd ${WORK_DIR}

# Extract original resource to ~/temp directory
while IFS= read -r gpath; do
    local_rel="${gpath#/org/gnome/shell/theme/}"   # strip prefix
    mkdir -p "${WORK_DIR}/$(dirname "$local_rel")"
    gresource extract "${RESOURCE_ORIG}" "$gpath" > "${WORK_DIR}/$local_rel"
done < <(gresource list "${RESOURCE_ORIG}")

# Patch several "background-color" items under various "switcher-list" elemwnts, not clear which of them are really used
BACKGROUND="rgba(255, 255, 255, 1.0)"
PATCH_CSS=("gnome-shell-dark.css" "gnome-shell.css")
for css in "${PATCH_CSS[@]}"; do
    f="${WORK_DIR}/Yaru/$css"
    sed -i '/\.switcher-list \.item-box:selected {/{n;s/background-color: [^;]*;/background-color: '"$BACKGROUND"';/}' "${f}"
    sed -i '/\.switcher-list \.item-box:highlighted/{n;n;s/background-color: [^;]*;/background-color: '"$BACKGROUND"';/}' "${f}"
    sed -i '/\.switcher-list \.item-box:highlighted/{n;s/background-color: [^;]*;/background-color: '"$BACKGROUND"';/}' "${f}"
done

{
    echo '<?xml version="1.0" encoding="UTF-8"?>'
    echo '<gresources><gresource prefix="/org/gnome/shell/theme">'
    find "${WORK_DIR}" -type f | sort | while read -r f; do
        echo "  <file>${f#${WORK_DIR}/}</file>"
    done
    echo '</gresource></gresources>'
} > "${WORK_DIR}/theme.gresource.xml"

# Compile
(cd "${WORK_DIR}" && glib-compile-resources theme.gresource.xml)

# Deploy
sudo cp "${WORK_DIR}/theme.gresource" "${RESOURCE}"

# Reboot (or re-login)
sudo reboot
```
This works:

Image

But it cannot be so complex!

Also, Claude suggested that there should be a way to modify the Yary theme locally instead of system level,
but so far I did not get working instructions.

Is there a better way? Any suggestions?

(By the way, Windows simply paints a thin white rectangle around the selected icon, which is always perectly visible).

Thank you

RividuPesara:
Fixes #4477 
     added a 1px white outline on the selected app in the Alt+Tab switcher


Jupi007:
Hi @RividuPesara, can you share before/after screenshots? Thanks :)

RividuPesara:
Hii @Jupi007 here are the before and after.


Before 
image




After
image

3v1n0:
I'm unsure we should rather follow the same design of the overview: Use a larger accent-color border on key-selected item.

Then in the high-contrast case, maybe the color may be tuned even further or we could just use white.

Jupi007:
I fully agree with you @3v1n0 
White makes sense in high contrast mode, but the regular design should be an accented ring.

RividuPesara:
Thanks for the feedback @3v1n0  and @Jupi007 , will update and push a new version

3v1n0:
Thanks, I've done some small adjustments, but I think we're good to go now!

